PR #7, opened 2015-01-07
search alternative names for the qmake executable

agaida:
some people really use Linux and other Systems :) No offense, but  it can't be compiled with Qt5 on Linux without the change


sfiel:
thanks a lot ... seems like it was a copy and paste mistake when combining the linux and windows cmake. I haven't tested compiling with Qt5 on Linux for a long time

agaida:
no - thanks for this great piece of software

PR #18, opened 2015-03-02
Correct right-click menu actions(Explorer and Metadata)

DoumanAsh:
Added uncheck in destructors for Metadata and File explorer widgets as both have own close buttons
Removed CloseEvent for File Explorer as it is the same as destructor...

Sorry for white spaces, my vim is angry about these :)
Let me know if you don't like that, i will re-upload commit without removal of white spaces

For my bug report:
http://nomacs.org/redmine/issues/417


diemmarkus:
thank you - removing unnecessary whitespaces makes sense - but for us it is harder to check the changes : )

PR #21, opened 2015-03-04
Fix build failure when all optional features are disabled

kensington:
When all optional features are disabled, build fails with undefined reference to FLT_MAX.

### DIFF
--- a/ImageLounge/src/er.h
+++ b/ImageLounge/src/er.h
@@ -29,6 +29,7 @@
 
 #pragma warning(push, 0)	// no warnings from includes - begin
 #include <cmath>
+#include <float.h>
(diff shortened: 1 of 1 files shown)


diemmarkus:
Thanks for the pull request -> why did you rename DkMath.h to er.h? could you revert the renaming please?

diemmarkus:
Since I am afraid of the file renaming, I will reject the request - but add the #include <float.h> line to fix the compile issue. thanks again!

kensington:
Sorry, the rename is some anomaly - definitely not intentional.

diemmarkus:
no problem - it's fixed now : )

PR #22, opened 2015-03-07
Fix cancel button in filter dialog

EXL:
QMetaObject::connectSlotsByName: No matching signal for on_cancelButtonSplash_pressed()


diemmarkus:
thank you!

PR #24, opened 2015-03-09
Corrext right click menu

DoumanAsh:
Two commits:
First with clean-up of white spaces
The second with correction itself

Not sure if i did it right, maybe i should make method inside of DkNoMacs for that instead of getting pointer to panelActions and setting updating it?
Also, each widget has CloseEvent(), would it be better to duplicate code there too?

### DIFF
--- a/ImageLounge/src/DkMetaDataWidgets.cpp
+++ b/ImageLounge/src/DkMetaDataWidgets.cpp
@@ -311,8 +311,7 @@ DkMetaDataDock::DkMetaDataDock(const QString& title, QWidget* parent /* = 0 */, 
 
 DkMetaDataDock::~DkMetaDataDock() {
     writeSettings();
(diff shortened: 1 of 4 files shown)


DoumanAsh:
Seems to me getPanelActions() should return vector itself, not reference to panelActions
So i add method to update panelActions

diemmarkus:
hi,
I have pulled your code but it does not compile with this issue:
```
1>C:\VSProjects\nomacs\ImageLounge\src\DkMetaDataWidgets.cpp(314): error C2352: 'nmc::DkNoMacs::setPanelActionChecked' : illegal call of non-static member function
1>          c:\vsprojects\nomacs\imagelounge\src\DkNoMacs.h(462) : see declaration of 'nmc::DkNoMacs::setPanelActionChecked'
```
I will now create a base class similar to DkWidgets.h to fix this issue.
thank you!

DoumanAsh:
Sorry for this inconvenience...

PR #31, opened 2015-10-25
OSX/clang compile fixes and warning suspended (CMakeLists.txt needs r…

pvanek:
OSX/clang compile fixes and warning suspended (CMakeLists.txt needs review for sure)

plus:
 - abs() -> std::abs()
 - >> in templates are replaced with > >
 - degree char is replaced with \u00B0 sequence
 - lots of && vs || precedence warning silented out


diemmarkus:
Hi Petr,
thanks for the fixes - I have now corrected the braces (see the commit message : ). We are now working on the developer branch for nomacs 3.0. It would be great if you find the time to check if this is compiling on OSX (I merged your changes there).

PR #34, opened 2015-10-28
fix compile on OpenBSD amd64/i386

sizeofvoid:
OpenBSD needs the uint64/int64 libtiff include define hack like MAC
window systems to build without errors.


diemmarkus:
thanks! - it would be awesome if you could also check the current developer branch - since this is the basis for nomacs 3 which will be released soon...

sizeofvoid:
see https://github.com/nomacs/nomacs/pull/36

PR #40, opened 2015-12-13
Add Fujifilm RAF file recognition

brainshave:
Hi,

Nomacs was loading the embedded 1920x1280 previews from RAF files with no problems but was ignoring them when showing the list of files in thumbnail view.

This solves the problem for me on Linux. Please let me know if the `*.raf` extension needs to be registered in any other part of the source code.

Cheers
Szymon


diemmarkus:
thanks!

PR #45, opened 2016-01-13
nomacs.desktop: German translation

pmattern:
Should be self-explanatory. `Name[de]` omitted as "Nomacs" gets used that way, too.


sfiel:
great, thanks

PR #50, opened 2016-02-10
Fix ifs in cmake file

tjom:
I stumbled across these minor issues in the cmake file (and fixed the indentation of that endif).


sfiel:
tnx

PR #73, opened 2016-08-18
Update README.md

xcsrz:
a couple README corrections to get the osx build to complete.  It still crashes for me and cannot be launched as a default image viewer, but at least I was able to build it.


diemmarkus:
thanks!

PR #76, opened 2016-09-17
Add support for High DPI screens

### LINKED ISSUE #75: High DPI support
nomacs currently seems to have no high DPI support. Setting 

QT_AUTO_SCREEN_SCALE_FACTOR=1

causes the UI to look correct, but images are pixelated. Without that environment variables, images look correct but the toolbar icons are too small.

Using

QApplication::setAttribute(Qt::AA_EnableHighDpiScaling, False)
QApplication::setAttribute(Qt::AA_UseHighDpiPixmaps)

And setting the toolbar icon size using pts rather than pixels should fix these issues (the typical approach is to make the toolbar 1.5 * lineheight where you get line haeing fromQFontMetrics of the application font.)

https://doc.qt.io/qt-5/highdpi.html

kovidgoyal:
Instead of changing the icon size to pts, we keep the old pixel setting
and dynamically rescale based on screen dpi. This fixes #75

### DIFF
--- a/ImageLounge/src/DkCore/DkSettings.cpp
+++ b/ImageLounge/src/DkCore/DkSettings.cpp
@@ -32,6 +32,7 @@
 
 #pragma warning(push, 0)	// no warnings from includes - begin
 #include <QImageReader>
+#include <QScreen>
 #include <QDesktopServices>
 #include <QTranslator>
 #include <QFileInfo>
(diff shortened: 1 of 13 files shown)


diemmarkus:
awesome - thanks!

kovidgoyal:
You are welcome, and thank you for nomacs :)

Note that I am only a basic user of nomacs, so there might well be areas of the interface that I missed.

PR #78, opened 2016-09-24
Changes to filter dialog

nholmgaard:
This pull request makes the filter dialog work with search terms with leading and trailing spaces without resorting to regex. Thus, an annotated file name such as:
20160924-2304 Picture of cat.jpg
can now be filtered out from a lot of 
2016xxxx-xxxx.jpg
by simply searching for a single space, " ".
The filter dialog will still fall back to regex if the search string does not match anything.

The pull request also makes a minor change to the default button, which was Cancel when invoking the search with Ctrl-F, and replaces a search term done two different places with a call to the same function.


diemmarkus:
thanks a lot! - I will only fix the wrong default button (find&filter can be called from two different menus which sets different default buttons - your pull request seems to remove this function)

PR #80, opened 2016-09-29
Use rsvg-convert for better SVG rendering

kovidgoyal:
At runtime, the presence of rsvg-convert is detected, and, if available
it is used to render SVG images. This is needed because QSvgRenderer
only supports SVG Tiny. An example of an SVG file it does
not render correctly: https://commons.wikimedia.org/wiki/File:Emoji_u1f337.svg

Note that QSvgRenderer is still used to render the application icons.

Downsides:

1) SVG rendering is slower. This is mitigated by using a disk cache to
store the rendered svg files, keyed by the file path, render size and
file mtime.

2) No support for SVG animation. This could be remedied in the future
by using another backend to render SVG. Options are librsvg and
QWebEngine.

EDIT: Also contains some commits to prevent nomacs from spamming stdout (I run nomacs from the terminal)

### DIFF
--- a/ImageLounge/src/DkCore/DkSettings.cpp
+++ b/ImageLounge/src/DkCore/DkSettings.cpp
@@ -969,7 +969,6 @@ void Settings::init() {
 	if (nmc::Settings::param().app().useLogFile)
 		std::cout << "log is saved to: " << nmc::DkUtils::getLogFilePath().toStdString() << std::endl;
 
-	qInfo() << "Hi there";
 	qInfoClean() << "my name is " << QApplication::organizationName() << " | " << QApplication::applicationName() 
 		<< " v " << QApplication::applicationVersion() << (nmc::Settings::param().isPortable() ? " portable" : " installed");
 
--- a/ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
@@ -854,7 +854,7 @@ void DkFilePreference::createLayout() {
 
 	// cache size
 	int maxCache = qMax(qRound(DkMemory::getTotalMemory()*0.5), 1024);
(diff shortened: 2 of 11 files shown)


kovidgoyal:
In this PR I use an env var to suppress printing of qInfo() messages. I did this so as not to disrupt anybody else's use case. However, IMO, the default behavior should be to not print qInfo() messages as they are only useful if you are debugging a problem/developing nomacs. As such rather than having an env var to suppress messages, we should have an env var to enable them. If you agree, I can make that change.

diemmarkus:
Hi @kovidgoyal,
thanks for your effort. Though we are aware, that the QSvgRenderer is not perfect, I am sorry that we cannot accept this pull request since it introduces more drawbacks than advantages:
- dependency to external software
- compatibility?
- what if user zooms?
- writing to the hard disk for loading
- performance (as you mentioned)

Please do not change debug commands like "Hi there", the nomacs output works just fine for us - so you should not change that.
If you need to output some debug statements you should use `qDebug()` rather than `qInfo()` - since `qInfo()` is written to the log file qDebug() not.

QWebEngine won't be added to nomacs, since it is a huge dependency.

kovidgoyal:
- There is no dependency. If rsvg-convert is not available, the code fallsback to QSvgRenderer
- What compatibility?
- It works just as before
- So what? Why is writing to the hard disk, in a cache location a bad thing?
- Thanks to the aforementioned cache, performace, apart from the initial load, is actually better.

The point is, I do no want spam output in my console. I dont use qInfo(), the existing codebase does, incorrectly. My commit adds *an option* to silence the useless qInfo() output that already exists. And I'm sorry but spamming stdout with useless messages like "Hi there" is just wrong. 

Anyway, it's your project, so feel free to reject the PR. I will just continue to use my fork.

PR #87, opened 2017-02-23
Compile fix

cornytrace:
These changes ensure that the CMake build directory can be the same as the source directory, and still being able to do "git add ." without the wrong files being included.
quazip.vcxproj gets auto generated, so no need to keep it in the git repo.

My VS2015 complained that assert was not defined in DkProcess.cpp, so I added #include <cassert>

### DIFF
--- a/ImageLounge/src/DkCore/DkProcess.cpp
+++ b/ImageLounge/src/DkCore/DkProcess.cpp
@@ -41,6 +41,8 @@
 #include <QWidget>
 #pragma warning(pop)		// no warnings from includes - end
 
+#include <cassert>
+
 namespace nmc {
 
 /// <summary>


sfiel:
Hi,
Thanks, but what should be the advantage of allowing an in-source build? Also cmake faq recommends out-source builds [1] because you can build multiple variants with different settings. I always have at least 2 (32 and 64 bit). Even if you have added a long list of files to ignore, you cannot cover all cases for the future (for example in your commit linux in-source builds are not covered ... here you would also have to cover gcc and clang builds a.s.o) and I'm pretty sure that a project file will be added to the source directory by accident because someone is not paying attention. Currently he would see a long list of file additions and is careful.



Regards,
Stefan



[1] https://cmake.org/Wiki/CMake_FAQ#Out-of-source_build_trees

cornytrace:
You're right, that is actually a better solution.
I'll leave just the compile fix then.

diemmarkus:
thanks!

PR #90, opened 2017-03-02
Enhancements for easier handling of large number of tabbed images

cornytrace:
My primary use case for this program involves having many images open at the same time in different tabs.
In order to make this easier I have added several options:
Open File can now open multiple images at the same time, each in their own tab. (This is optional, as selecting one file works just as before)
New menu buttons "Open Tabs" and "Save Tabs", which respectively load and save a text file containing the file paths of every open tab, one per line. This allows for easy opening of many different sequences of images, not necessarily in the same folder and/or order.
Elision of tabs: The tabs now get smaller when there are a large number of them open. When the minimum tab size has been reached, the Tabbar scrolls as before.
Fixed bug in loadFileToTab, see commit.
I added a function to switch to a tab by idx, and changed previoustab() and nexttab() to use it. This in order to add the functionality that when opening multiple tabs, after opening all of them it switches to the first one opened, as that is usually the image you want to see first.
I have made these changes in order for the program to fit my use case. I do not know if the changes I have made fit within your vision for the program. Please comment on what you would like changed for this to be accepted.

### DIFF
--- a/ImageLounge/src/DkCore/DkActionManager.cpp
+++ b/ImageLounge/src/DkCore/DkActionManager.cpp
@@ -944,7 +944,7 @@ void DkActionManager::createActions(QWidget* parent) {
 	mFileActions[menu_file_open_dir]->setShortcut(QKeySequence(shortcut_open_dir));
 	mFileActions[menu_file_open_dir]->setStatusTip(QObject::tr("Open a directory and load its first image"));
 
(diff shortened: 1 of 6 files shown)


diemmarkus:
Hi,
nice feature (I especially like the scrolling in the tabbar - didn't know that)...
thanks!

PR #91, opened 2017-03-03
Quick Save Tabs fix

cornytrace:
File which was saved to by Save Tabs was not emptied before writing.
Whoops!


diemmarkus:
thx : )

PR #93, opened 2017-03-05
More tab enhancements

cornytrace:
This pull requests adds:
Fixes for the open/save tabs file type filter
An setting was added to not allow opening files which are already open. (Check for Duplicates on Open)
change removeTab to close the last tab, but replace it with an empty tab.
Add "First Tab", "Last Tab", "Go to Tab", "Close all Tab" options to menu.
Add the current and total number of tabs to the window title, if there is more than one tab.
Changed tab count to start at 1, internally still starts at 0.
Move new, close, close all tab to file menu.

One question, is my removeTab refactor correct in all cases? Why the for loop if with updateTabIdx() you can force the mTabInfos idx and getTabIdx to match, thus making the for loop obsolete?


diemmarkus:
thanks again for the extension - I will leave a few comments before merging...

ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -1272,23 +1272,47 @@ void DkNoMacs::openFile() {
...
+			}
+		}
+
+		if (!dup) {
+			qDebug() << "os filename: " << fileName;
+			getTabWidget()->loadFileToTab(fileName);
+		}
+	}
+	if (duplicates.count() > 0) { // Show messagebox if at least one duplicate was found
+		QMessageBox dupbox(this);
```
diemmarkus: I think a messagebox is too much here (note: we hardly ever use messageboxes), please use either:
- a DkMessageBox (for the user can permanently switch it off)
- `getTabWidget()->getViewPort()->getController()->setInfo(tr("some message"));`

ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -1272,23 +1272,47 @@ void DkNoMacs::openFile() {
...
+		QString duptext = "The following duplicates were not added:\n";
```
diemmarkus: this is rather important: messages that are brought to the user need a `tr()` - for nomacs supports translations : )

ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -2131,13 +2155,18 @@ void DkNoMacs::setWindowTitle(QSharedPointer<DkImageContainerT> imgC) {
 
 void DkNoMacs::setWindowTitle(const QString& filePath, const QSize& size, bool edited, const QString& attr) {
 
+	QString title;
+
 	// TODO: rename!
+	if (getTabWidget()->getTabs().count() > 1) {
```
diemmarkus: I do understand that this is needed for `Go To` tab but I don't like to have much information in the title bar (there were lot's of requests to put other information there too). Moreover, there are a few update glitches (e.g. wen opening a new tab) - so please remove it.

ImageLounge/src/DkCore/DkActionManager.cpp
```diff
@@ -1279,22 +1301,23 @@ void DkActionManager::createActions(QWidget* parent) {
...
-	mViewActions[menu_view_close_tab]->setShortcut(QKeySequence(shortcut_close_tab));
-	mViewActions[menu_view_close_tab]->setStatusTip(QObject::tr("Close current tab"));
+	mViewActions[menu_view_first_tab] = new QAction(QObject::tr("F&irst Tab"), parent);
+	mViewActions[menu_view_first_tab]->setStatusTip(QObject::tr("Switch to first tab"));
 
 	mViewActions[menu_view_previous_tab] = new QAction(QObject::tr("&Previous Tab"), parent);
 	mViewActions[menu_view_previous_tab]->setShortcut(QKeySequence(shortcut_previous_tab));
 	mViewActions[menu_view_previous_tab]->setStatusTip(QObject::tr("Switch to previous tab"));
 
+	mViewActions[menu_view_goto_tab] = new QAction(QObject::tr("&Go to Tab"), parent);
```
diemmarkus: `Go to`, `First Tab`, and `Last Tab` are from my point of view too much - maybe we could add an option _extended tabs_  that adds these items to the menu (including open tabs & save tabs)?

diemmarkus:
since you are working hard on improving nomacs' tab behavior - I summarized two feature requests (#94, #95) that we received from a lot of users (in case you find the time and like the requests : )

cornytrace:
Those are certainly features that interest me as well, I will try to implement them.

ImageLounge/src/DkGui/DkNoMacs.cpp
cornytrace: Do you have any suggestions on where I can place this information then?

cornytrace:
Your comments have been addressed, and #94 is done. But i'm afraid #95 may be a bit out of my league, and I use nomacs portable anyways.

That leaves one thing: showing the current and total tab amount. I could put them in the window title behind the extendedTabs barrier, but i'd understand if you didn't want that. I could put it on the Status bar, but it is disabled by default. We could put it in a setInfo(), but then it would only and mostly unnecessarily show up when changing tabs.
For now i'll put it as the default value of the Go to tab button, but it would be great if it could be put somewhere more obvious.

diemmarkus:
thanks a lot - I will have a look at #95 as soon as I find the time. I gave it a thought and maybe the title, like you had it, is just fine - since it is not default behavior anymore (the user has to select the extended tabs). After that I guess we are good to go (merge) the pull request...

cornytrace:
Last commit has been amended. As far as I can see, it's good to go!

diemmarkus:
thanks!

PR #100, opened 2017-04-03
Don't install to absolute paths

ahmedtd:
Static files should probably be installed relative to the installation root, rather than via absolute paths.

Ran into this while packaging `nomacs` for the `nixpkgs` repository.


diemmarkus:
thanks!

Issue #109, opened 2017-05-05
make libnomacs-core a private lib


agaida:
aka install it to usr/lib/\*/nomacs/libnomacsCore.so.\* instead of usr/lib/\*/libnomacsCore.so.\* - if this would be possible upstream a lot of busy package maintainers would be thankful - esp. myself :P 

Really, libnomacscore3 is a pita to pack, esp the symbols - if libnomacscore would be a private lib used only by nomacs there would be no need to pack it in a extra package (and also don't mess around with symbols)

sfiel:
hi,
if we change the install directory in cmake to lib/nomacs/ then it also affects the installation into the /usr/local . After a "make install" i'm unable to execute nomacs (I also executed ldconfig), because the system is unable to find libnomacs. seems like,ldconfig does not look for libs in /usr/local/lib/nomacs. do you know how to solve this issue? thanks

agaida:
hihihi, diesen Part wollte ich ganz gepflegt auf Euch abwälzen. I will take care of.

diemmarkus:
I guess that's it - right?

agaida:
right, does the trick

PR #120, opened 2017-05-17
Fix runtimepath on linux

derlunz:
This PR should fix issue #109


agaida:
Cool, impressive - works like charm. :)

derlunz:
Thanks! Most of the work was already there.

BTW, if you stumble over packaging issues, just contact me. I am running Nomacs under Linux myself, so I am in a good position to give feedback on that platform.

agaida:
than we are at least two who use nomacs with linux - and i  maintain the debian packages

diemmarkus:
I guess that's good to merge than : )

agaida:
ok, your latest changes created a little bit havoc in the packaging - thats cleaner now. with the patch applied we can deliver nomacs as one binary with private libs included. That will ease the packaging pain a lot.

PR #121, opened 2017-05-18
Include a git version tag, if built from a git repo.

derlunz:
- makes `nomacs --version` report:

   [INFO] Hi there
   Image Lounge 3.7 (git: 3.6.1-48-g3494189-dirty)

- About dialog shows the same string.

### DIFF
--- a/ImageLounge/CMakeLists.txt
+++ b/ImageLounge/CMakeLists.txt
@@ -16,6 +16,7 @@ endif(COMMAND cmake_policy)
 project(nomacs)
 
 # needed for soname
+
 set(NOMACS_VERSION_MAJOR 3)
 set(NOMACS_VERSION_MINOR 7)
 set(NOMACS_VERSION_PATCH 0)
@@ -27,7 +28,20 @@ else()
 	SET(NMC_ARCHITECTURE "x86")
 endif()
 
-add_definitions(-DNOMACS_VERSION="${NOMACS_VERSION}")
+if(NOMACS_BUILD_TAG)
+    #we've been passed a version tag, use that
+    set(NOMACS_VCSVERSION "(build: ${NOMACS_BUILD_TAG})")
+    set(NOMACS_FULL_VERSION "${NOMACS_FULL_VERSION} ${NOMACS_VCSVERSION}")
+else()
+    include(cmake/GitVersionString.cmake)
+    get_git_version_string()
+    if(NOMACS_VCSVERSION)
+        #we have git, add the current description
+        set(NOMACS_FULL_VERSION "${NOMACS_FULL_VERSION} ${NOMACS_VCSVERSION}")
+    endif()
+endif()
+
+add_definitions(-DNOMACS_VERSION="${NOMACS_FULL_VERSION}")
 
 set(BINARY_NAME ${CMAKE_PROJECT_NAME})
 set(NOMACS_BUILD_DIRECTORY ${CMAKE_BINARY_DIR})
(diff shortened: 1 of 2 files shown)


agaida:
Nice, but it would be nice to have a way to put in a version from a kind of version file  - rationale: yes, i do git builds and the git version would be nice to have - but i don't build  directly from git.

The workflow is:
* get current sources
* mention it in changelog
* build a source package
* upload it to the builder

The reason for doing so - i don't build things local, only in clean build environments - most distributions do this to be sure that the receipe will work and no dependency is missed in the receipe. But it would be a big help to have the git version string for these builds too.

derlunz:
The git version tag is optional. It will only be appended, if building from a git repo. So the PR should meet the requirements right now. This is just a minimal implementation.


You want a version file? No problem. I'd could implement a scheme akin to the one found in [opencv](https://github.com/opencv/opencv/blob/master/cmake/OpenCVVersion.cmake):

- have a version file e.g. `${CMAKE_SOURCE_DIR}/src/DkCore/version.h`, that defines  the current version:
```
...
#define NOMACS_VERSION_MAJOR    3
#define NOMACS_VERSION_MINOR    1
#define NOMACS_VERSION_REVISION 0
#define NOMACS_VERSION_STATUS   "-dev"
...
```
- have `cmake`parse that file, and set set corresponding variables within the build system, possibly append a git tag.
- have `cmake` generate a final version file `${CMAKE_BUILD_DIR}/version.h` :
```
...
#define NOMACS_VERSION_MAJOR    3
#define NOMACS_VERSION_MINOR    1
#define NOMACS_VERSION_REVISION 0
#define NOMACS_VERSION_STATUS   "-dev"
#define NOMACS_VERSION_VCSTAG   "<A_NICE_GIT_TAG_OR_EMPTY>"
#define NOMACS_FULL_VERSION "3.1.0-dev <A_NICE_GIT_TAG>"
...
```
or
```
...
#define NOMACS_VERSION_MAJOR    3
#define NOMACS_VERSION_MINOR    1
#define NOMACS_VERSION_REVISION 0
#define NOMACS_VERSION_STATUS   "-dev"
#define NOMACS_VERSION_VCSTAG   ""
#define NOMACS_FULL_VERSION "3.1.0-dev"
...
```
- include the generated `${CMAKE_BUILD_DIR}/version.h` within application code.

That way we get:
 - version is set in dedicated header file
- version information is used by the build system
- the git tag is only appended when building from a git repo.

Concerning the build work flow:

>    get current sources

```
     git pull origin/master
```
>    mention it in changelog
```    
$EDIT ImageLounge/src/changelog.txt
git commit -am 'updated change log for release X.Y'
```
>    build a source package
``` 
build and check for goodness
```

>    upload it to the builder
```
git tag 'X.Y.Z' 
git push --tags origin/master
```
How do you cause the builder to start? Doesn't it use git to retrieve the sources?

derlunz:
I just saw what you meant. It seems, you use the travis builder. In that case you want to pass a tag description via environment variable to cmake. Something like:

```yml
#filename: .travis.yml
...
before-script:
  - NOMACS_CI_TAG=`git describe --tags --always $TRAVIS_COMMIT`
  - CMAKE_ARGS+=" -DNOMACS_BUILD_TAG=$NOMACS_CI_TAG"
...
```
NOTE: This is just conceptual. I haven't ever touched a travis file.

agaida:
nop, you misunderstand - i upload a source build to my repo - and thats all. The same way i would upload a source build to debian or ubuntu or a ppa on launchpad. And no - i would never use git to get the sources in the builder.

derlunz:
So that means no git tag in a binary for debs. Considering the patches applied during packaging that doesn't make sense anyway. Other question: 

Does it make sense, resp. is there a way to brand the binaries as coming from some distro?
Why can you work with a "version file" and not with an environmental variable? Why not patching the build script?

agaida:
Thats easy - why should one patch sources if there is a way to simply read a file or may be an option? There might be several ways to provide some informations. Environmental variables may be one way, something like options are another way. I should have a look at, should be an one or two-liner to achieve that. Sorry for the late answer, was busy that weekend, building my first Arm-System.

derlunz:
Take your time, the source tree will be still be there. Thanks for looking into it.

diemmarkus:
I'll close this request - since there are no recent updates...

Issue #115, opened 2017-05-12
nomacs silently refuses to open files with no extension


protist:
To replicate:
1. Have two (image) files in a directory: `a.png` and `b.png`
2. Rename `a.png` to `a`
3. Attempt to open `a` in nomacs

Expected results:
* Image `a` is opened in nomacs

Observed results:
* Image `b.png` is opened in nomacs

Gwenview, image viewer, okular, etc. open up `a` fine. `file a` shows it as `PNG image data`.

Oddly enough, if there is only `a` in the directory, nomacs partially recognises it as an image. It displays fine, but there is an index overlay on the top left. Also, up an d down arrows scale the image, but the mouse scroll wheel does not. Pressing left or right pops up `/path/to/directory does not contain any image`.

nomacs 3.6.1
Arch Linux (up-to-date)

eadmaster:
also, if the extension is wrong, emacs may suggest to rename the file like Irfanview does...

vext01:
Same problem here.

My gopher client wrote out a file named `/tmp/tmpctjb9n76` which nomacs will not open. `feh` is ok with it though.

It's a jpeg.

protist:
Thank you for the fix. I can confirm that the latest version works as expected.

Ultimatedude10:
sadly this does not work on version 3.17.2263

PR #123, opened 2017-05-19
More robust file loading.

derlunz:
This PR fixes issue #115 by:

- adding file content detection via MIME database in `DkUtils::isValid()` as last resort.
- notifying the user in GUI when loading of a file failed.

further
- it stops opening new Tabs, when dropping a file onto 'untabbed' nomacs.
- it partially addresses inconsistencies when dropping directories.

### DIFF
--- a/ImageLounge/src/DkCore/DkUtils.cpp
+++ b/ImageLounge/src/DkCore/DkUtils.cpp
@@ -59,6 +59,7 @@
 #include <QMainWindow>
 #include <QMouseEvent>
 #include <qmath.h>
+#include <QMimeDatabase>
 
 #include <QSystemSemaphore>
 #pragma warning(pop)		// no warnings from includes - end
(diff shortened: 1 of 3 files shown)


diemmarkus:
thanks!
I prefer this behavior (only add tabs if it's already in 'tabbed' mode). However, there seems to be an issue with states: when I now drop an image, the background logo blinks for a short period of time - this indicates, that we do a lot of extra work. can you have a look at that - since it is a new behavior?

derlunz:
Jep. I'll look into this.

derlunz:
While throwing files at nomacs I remembered, running into infinite loops in [`DkViewport::loadFile`](https://github.com/nomacs/nomacs/blob/34941894e9c11e1ba8f74e2b1d05fd09e86f4fab/ImageLounge/src/DkGui/DkViewPort.cpp#L1598), when passing valid urls. Another effect with the same root is occasionally longer load times.

Another issue i see, is that  dropping zips is broken.

diemmarkus:
I see - I can fix the looping issue - I don't think that's caused by your changes...

PR #124, opened 2017-05-19
RFC: Add colormapped mode to pseudocolor toolbar.

derlunz:
This PR adds a new mode to the PseudoColor toolbar, that displays an imag channels according to some well-defined colormap:

pseudocolor-demo

- Colormaps are pulled from [matplotlib](http://matplotlib.org/examples/color/colormaps_reference.html).
- Include a selection of 10 colormaps, including `viridis`, `magma`, `inferno` and `gist_rainbow`.
- The maps are sampled at 256 points, and stored as xml. This is a limitation of the current script, and can be changed to support segmented colormaps, that are considerably smaller.

At this point this feature is a bit rough with respect to UI, but fully working.


diemmarkus:
works - thanks! two questions (before merging)
- why do we need to distinguish between color (created by the user) and colormap?
- can you compress the colormaps?
  - 180 KB resources for colormaps is a bit an overkill
  - more important: the user cannot really edit these colormaps because there are 255 supporting points

there is an easy way of compressing:
````
c = colormap('jet');
sum(diff(diff(c)),2) ~= 0; % 1 if we need the supporting point
````
 does that make sense?

diemmarkus:
closed due to age

diemmarkus:
closed because it's old...

PR #128, opened 2017-06-11
Use the correct variable name

cybre:
`OpenCV_FOUND` is set in the `.cmake` files but `OPENCV_FOUND` is used in the CMakeLists.txt which means it will always say that it will not be compiled with OpenCV support because `OPENCV_FOUND` is never set.


diemmarkus:
thank you!

Issue #133, opened 2017-07-09
Build .dmg files for homebrew cask for Mac?


jcrben:
Hi - I went ahead and compiled this from source following the macOS / OSX instructions https://github.com/nomacs/nomacs#build-nomacs-osx. The only hitch is that that I had to tweak my PATH to pick up qmake with `export PATH=/usr/local/opt/qt/bin:$PATH`. Did I miss something, or should I make a PR to update the instructions?

If you built a .dmg file as part of the release, it would be pretty easy for me to write up a cask (following https://github.com/caskroom/homebrew-cask/blob/master/doc/development/adding_a_cask.md) for it and submit it to homebrew.

homebrew is more popular in the Mac world these days than macports.

NOTE: it would also be worth looking into getting this into nixpkgs - I don't feel like homebrew is very durable. For example we're not pinning the versions of the dependences in the command `brew install qt5 exiv2 homebrew/science/opencv libraw quazip`.

Noticed this ticket https://nomacs.org/redmine/issues/300 which pointed to me https://github.com/nomacs/nomacs/pull/37 - seems like Travis is just about set up to compile the dmg...?

diemmarkus:
Hi,

travis is set up and building fine, yes. I never tried to deploy the travis executables for we cannot test them. Here seems to be a discussion on how to deploy travis dmgs: https://stackoverflow.com/questions/29587581/how-to-build-macosx-executables-on-travis-ci
If you know how to do this, we would appreciate a PR (updating the instructions would be fine too : )

cheers,
markus

scrubbbbs:
Mac OS build was updated in 97bbcd5f3cf9083068be039b1b37f6203f2b9e08. We use homebrew to compile so it should be possible to contribute a cask.

PR #180, opened 2017-12-27
Update macOS build instructions

tcmb:
Fixes two things I encountered while trying to build latest master on macOS 10.13.2:

1. `opencv` is now in homebrew/core, specifying `homebrew/science/opencv` path doesn't work any longer
2. The `qmake` binary is not on the `PATH` after installation, as was also mentioned in [issue 133](https://github.com/nomacs/nomacs/issues/133). Linking it into `/usr/local/bin` seems standard (rather than adding the qmake dir itself to the PATH).

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,7 @@ Install [Homebrew](http://brew.sh/) for easier installation of depenedencies.
 Install required dependencies:
 
 ```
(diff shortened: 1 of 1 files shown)


diemmarkus:
thanks!

PR #187, opened 2018-01-29
cmake: Stop calling QT5_WRAP_CPP() when building quazip and libqpsd.

rakuco:
`cmake/Utils.cmake` already sets `CMAKE_AUTOMOC` to on.

Calling `QT5_WRAP_CPP()` used to be just redundant, as `QUAZIP_MOC_SRC` and
`LIBQPSD_MOC_SRC` were never actually added as source dependencies in
`MacBuildTarget.cmake` and `UnixBuildTarget.cmake`. In other words, CMake's own
automoc infrastructure was actually being used and the moc invocations from
`QT5_WRAP_CPP()` were not being made at all.

Starting with Qt 5.9.4, calling `QT5_WRAP_CPP()` disables the `AUTOMOC` property
on the macro's input files, which means neither CMake's automoc
infrastructure not `QT5_WRAP_CPP()`'s code were being used and we ended up
with several 'undefined reference to vtable' errors when linking.


diemmarkus:
thx!

PR #200, opened 2018-03-13
add NSPrincipalClass in Info.plist

ThomasSevestre:
This resolves blurry interface on Retina screens


diemmarkus:
thanks again : )

PR #202, opened 2018-03-15
Update Bosnian, Croatian and Serbian translations

dglava:
I have updated the translation files for the above mentioned languages.

I'm sorry for not using the Crowdin web-interface. I find using Qt Linguist just easier and more convenient compared to a web-interface when working on bigger translations.


diemmarkus:
thanks!

PR #212, opened 2018-04-13
Fix cmake with Qt 5.11

a17r:
cmake fails due to a change in Qt 5.11 beta3:
```
CMake Error at cmake/UnixBuildTarget.cmake:61 (qt5_use_modules):
  Unknown CMake command "qt5_use_modules".
Call Stack (most recent call first):
  CMakeLists.txt:178 (include)
```

(only tested with Linux)


diemmarkus:
win cmake is fine too - thanks!

PR #215, opened 2018-04-23
Make compilation of libqpsd optional

Miriup:
Had some problems with compiling libqpsd. Added a cmake compile flag that allows selectively turning its usage off.


Miriup:
Weird. Yes, will look into it.

diemmarkus:
closed due to age...

PR #224, opened 2018-06-04
Fix macOS compile

riverol:
Updated MacBuildTarget.cmake to fix compile with latest cmake and qt5 on macOS.


diemmarkus:
thanks!

PR #259, opened 2018-10-01
install manual pages in the correct folder on OpenBSD

sizeofvoid:
Simple fix to install manual pages in the correct folder on OpenBSD


diemmarkus:
thanks!

Issue #318, opened 2019-04-08
Dark theme on no borders mode shows a thin 1-pixel box


hyphenrf:
Setting dark theme then choosing to show no panels/borders produces this image box.
But you have to be in normal (not maximized) view.

OS is Solus 4.0 with Budgie 10.5 DE and GTK+ theme Plata-Noir-Compact
Issue persists with different Plata variants
Haven't tested on other themes
This only happens with Nomacs' in-house dark theme selected -- regardless of the GTK theme.

Example:
![image](https://user-images.githubusercontent.com/26802762/55731183-0de0da00-5a1a-11e9-9a02-9e3ecab4f994.png)

Nomacs version is Image Lounge 3.12

diemmarkus:
Thanks for reporting. Telling from the screenshot provided, it is hard for me to distinguish between nomacs and the OS UI. If I understand correctly, the black 1px border results from the nomacs `Dark theme` which is unwanted?! Could you please take a full screenshot (including the nomacs window frame)?

thanks!

hyphenrf:
absolutely. Here it is:
![image](https://user-images.githubusercontent.com/26802762/55870142-3850a580-5b88-11e9-88fa-fffb29246b29.png)
Seen here: the blue painting is the nomacs picture and the border around it is the undesired ui element.
The rest is regular UI.

The 1px box is what takes the place of a 'window border' in frameless mode I assume.
It defeats the purpose of being 'frameless'

diemmarkus:
now I understand, thanks! Unfortunately, I cannot reproduce this on my system:
![img-2019-04-10 12 23 36](https://user-images.githubusercontent.com/4182443/55872319-faa24b80-5b8c-11e9-827a-71270f1f7460.png)

but there might be a fix for the issue:
- open [nomacs install dir]/themes/Dark-Theme.css
- locate these lines:
```css
nmc--DkCentralWidget {
	border: 1px solid #000;
}
```
- change `#000` to `#f00`
- restart nomacs and see if the border turns red
- if so, change the line to `border: none;`

please let me know if this fixes the issue (I will change it in the upcoming release)

hyphenrf:
I was not notified for this reply. Will try this fix right now and get back to you asap.

hyphenrf:
Oh I also forgot to mention: on frameless mode if the "window" is maximized (like in your case?) The picture will be frameless and the 1px border will not show. 
[...]

PR #332, opened 2019-05-13
Removed 1px border from frameless windows

hyphenrf:
When not maximized, a frameless window will show a thin 1px border.
This solves the issue #318


diemmarkus:
thanks!

PR #349, opened 2019-07-07
Fix parenthesis in logical expression

tcberner:
DkPluginManager: fixes wrong nullptr check of DkPluginInterface::interface_batch


diemmarkus:
thanks!

PR #375, opened 2019-09-12
The French translation for 'Make' is 'Fabricant'.

damorinCan:
'Construire' is a verb but it should be a noun. This can be compared with the German version 'Hersteller' which translate to 'Fabricant' as per Google translate (and Collins).


diemmarkus:
thx!

PR #409, opened 2019-12-21
Fix build with opencv 4.2

antonio-rojas:
Port away from deprecated C API which is no longer available by default in 4.2


diemmarkus:
thx!

PR #454, opened 2020-03-26
Add FreeBSD support

vkarlsen:
Add FreeBSD to the list of operating systems that need stdlib.h
instead of malloc.h.


diemmarkus:
Thanks!

PR #455, opened 2020-03-31
Add inno setup for windows installer

fitu996:
The original ".msi" installer lacks some features such as install nomacs for all user, file association during install, etc.

I create an installer config, which uses inno setup. It supports install for current/all user, add file association based on users' choice, reset user settings, chooses language automatically and of course, uninstall.

Try this installer here:
[https://mega.nz/#!uB1A2IyY!4wbuRh6cs9BCGHDfBrVj6dVf8tAJrTrRlabikAVyeN8](https://mega.nz/#!uB1A2IyY!4wbuRh6cs9BCGHDfBrVj6dVf8tAJrTrRlabikAVyeN8)


fitu996:
To build the installer:
1. download and install inno setup
2. install all unofficial language support for inno setup
3. download nomacs portable version and extract to installer-inno
4. open `installer-inno/nomacs setup.iss` and compile

There may have built fails on line 118 due to the lack of `bin/settings.ini` in the latest stable nomacs portable version (which exists on previous versions). Delete this line or create an empty `bin/settings.ini` solve this problem.

diemmarkus:
thanks for the contribution! 
from what I see now, I specifically like:
- setting the langue w.r.t to the installer language
- installing nomacs for multiple users (this is something people ask for often)

Since this is the forth installer bundle after NSIS, Qt, WiX, I have a few questions not to make the mistakes we did before:
- is there a reason why you chose *.exe instead of *.msi?
- did I get it correct that 'local user only' does not require admin permission?
- is there a reason why you added all our licenses (from what I understand, this is not necessary)?
- are file associations greedy? (is the user's default image viewer replaced by nomacs)?

missing the `settings.ini` in the portable version is an error in our packaging (which I fixed by now) - but it won't be there for the true setup (this is how nomacs recognizes that it is portable)

fitu996:
# Is there a reason why you chose *.exe instead of *.msi?
No. Because the only packaging method I have learned is inno setup. I do not know how to package it to a msi format. And as far as I know, inno setup do not support package as msi.
# Did I get it correct that 'local user only' does not require admin permission?
Yes. 'local user only' does not require admin permission.
# Is there a reason why you added all our licenses (from what I understand, this is not necessary)?
I add it because I think it is necessary. If it doesn't, I will remove it.
# Are file associations greedy? (is the user's default image viewer replaced by nomacs)?
It depends on user choice. Users can choose only set associations but not set as default, or set associations and set as default.

But in win10, set as default using installers is not possible. It is Microsoft that rejects it and there isn't any solution.

fitu996:
For `settings.ini` in the portable version. I know exactly what an empty `settings.ini` do. The installer will remove this file during installation.

fitu996:
By the way, could you answer this question?

[https://github.com/nomacs/nomacs/commit/8d0078dc63f32c23a4d72b3b372be54a3c8ac73f#comments](https://github.com/nomacs/nomacs/commit/8d0078dc63f32c23a4d72b3b372be54a3c8ac73f#comments)

diemmarkus:
> And as far as I know, inno setup do not support package as msi.

I googled it before and misunderstood the answers - you're right we can't package it as msi. from what I understand (I am not a sys admin), msi have advantages when it comes to silent install etc. is it straight forward to run a silent install with your setup?

> I add it because I think it is necessary. If it doesn't, I will remove it.

I personally don't like displaying licenses during install (no one reads them anyway : )

> For `settings.ini` in the portable version. I know exactly what an empty `settings.ini` do. The installer will remove this file during installation.

ah ok, it won't be there from my local install (so it's better to remove the line)

> But in win10, set as default using installers is not possible. It is Microsoft that rejects it and there isn't any solution.

I know - had troubles there too : )

fitu996:
Installer after removing the licenses: [https://mega.nz/#!qVt2VC4L!Uf-88827t-B4xawHHXs9K5xuLqMtDw9x50W7CO0tcxg](https://mega.nz/#!qVt2VC4L!Uf-88827t-B4xawHHXs9K5xuLqMtDw9x50W7CO0tcxg)

Inno setup support silent install: [https://jrsoftware.org/ishelp/index.php?topic=setupcmdline](https://jrsoftware.org/ishelp/index.php?topic=setupcmdline)

Try my installer with `installer.exe /?` you will get the full help.

fitu996:
Any other requirements?

If you find inno setup is difficult to use, then I can maintain this installer. Just tell me about all changes in file associations \(such as add new format, changes of icons, description, etc.\) and new language support.

Or if you have IRC or something similar to it. We can have a further talk.

diemmarkus:
> Any other requirements?

not so far... thanks!
I'll create the setup on my machine (in the end it's just calling your script) - still I'd be happy if you could further maintain the scripts 👍

diemmarkus:
>     2\. install all unofficial language support for inno setup

I am having an issue with this now. I don't want to install the unofficial language packages into appveyor. Is it possible, that the setup script detects missing languages and gives a warning instead of an error? (see [this appveyor build](https://ci.appveyor.com/project/diemmarkus/nomacs/builds/33751826))

fitu996:
Sorry but inno setup seems have no built-in support for this feature. It can be implemented manually but it will make the setup profile too complacated and difficult to read.

diemmarkus:
ok - I removed unsupported languages for now...

PR #465, opened 2020-04-13
Let the user choose the root directory of the file explorer

ivanradanov:
The problem I had with the file explorer was that the root of it was at / (on unix-likes) so you would need to go a lot of levels down until you get to your images and then you would need to scroll left/right in the explorer which was cumbersome.

So my proposed solution is to let the user choose the root directory, the default is / on unix likes and whatever qt thinks the root of the filesystem is in windows, perhaps it might be better to change to ~ on unix-likes and My documents or something in windows.

Example:
2020-04-14-041632_254x211_scrot

Tested only on linux, thought it should work elsewhere as well.

### DIFF
--- a/ImageLounge/src/DkGui/DkBaseWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkBaseWidgets.cpp
@@ -354,6 +354,41 @@ DkLabelBg::DkLabelBg(QWidget* parent, const QString& text) : DkLabel(parent, tex
 	setObjectName("DkLabelBg");
 }
 
+// DkElidedLabel --------------------------------------------------------------------
+DkElidedLabel::DkElidedLabel(QWidget *parent, const QString &text)
+	: QLabel("", parent) {
+	setText(text);
+	setSizePolicy(QSizePolicy::Ignored, QSizePolicy::Minimum);
+}
+
+void DkElidedLabel::setText(QString text) {
+	content = text;
+	updateElision();
+}
+
+void DkElidedLabel::resizeEvent(QResizeEvent *event) {
+	updateElision();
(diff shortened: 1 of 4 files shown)


diemmarkus:
cool thanks! I'll test it on windows today...

diemmarkus:
thanks again. I tested it yesterday & only found one issue: if an image is loaded, the corresponding directory unfolds in the explorer. with your addition, it is not revealed if the image is outside the current root path. do you have an idea how we could fix this behavior?

ivanradanov:
Thanks for the review.

I suppose we could hook into an image loaded signal (have not looked into it yet so I don't know how easy it will be) and change the file explorer directory to the directory of the image that was loaded.

However, this could lead to some unwanted behaviour.
Consider you are accessing image /a/b/c/img and you have set your file explorer to /a/b because you are dealing with a lot of images in subdirectories of /a/b
If you accidentally click another tab and load the image /x/y/img, the explorer would load /x/y, and when you go back to the original tab, it will load /a/b/c/, not /a/b/ which could be annoying, not sure to what degree.

I was thinking of an entry "Visit directory in explorer" in the image's context menu that you could click to load the directory it is in to prevent unwanted directory changes as above, which has its own disadvantages I suppose.
The same entry could be nice for the directory context menu in the explorer, for faster switching. Which makes me think about an "Up" button for going to the parent directory in the explorer, but things might get a bit cluttered...?

diemmarkus:
> However, this could lead to some unwanted behaviour.

you are right there... in the end, users can set the root directory in case they want the old behavior back. 

Now I'm only thinking if we should change the default path on windows to be root (at least I have hardly any images in my user folder on Windows).

Issue #292, opened 2018-12-18
Add .avif (AV1 Image File Format) support


EwoutH:
The AV1 Image File Format (AVIF) is a new image format based on the AV1 video codec. It would be great if support for this format could be added.

The AVIF spec can be found here: https://aomediacodec.github.io/av1-avif/
Netflix published a bunch of test images: http://download.opencontent.netflix.com/?prefix=AV1/Chimera/AVIF/

diemmarkus:
I'll have a look at it - thanks!

EwoutH:
[libavif](https://github.com/joedrago/avif) tagged a new release, [0.1.3](https://github.com/joedrago/avif/releases/tag/v0.1.3). See [Colorist](https://github.com/joedrago/colorist) for an example [implementation](https://github.com/joedrago/colorist/blob/master/lib/src/format_avif.c).

novomesk:
I just installed nomacs and I observed that it opens AVIF and AVIFS files.
I develop a Qt Image Plugin with AVIF support and I think that maybe nomacs loaded the *.AVIF files because it is a QT application.
![nomacs](https://user-images.githubusercontent.com/52529860/76639268-1e64d300-654e-11ea-8db0-a44a6f6ec135.png)

diemmarkus:
> I develop a Qt Image Plugin with AVIF support and I think that maybe nomacs loaded the *.AVIF files because it is a QT application

for sure - nomacs uses any qt image plugins that are available on the system. is your image plugin open source? if yes, I will add it to the default installer on windows.

novomesk:
Yes, my plug-in is free.
I am preparing a new improved version but it is not public yet.
So far you can see how the old version of my plug-in looks like, I shared sources here:
https://bugs.kde.org/show_bug.cgi?id=416941
The old version of my plug-in works with libavif 0.5.7 but doesn't work with 0.6.x (New version will work with 0.6.x).
I can notify you when the new version is ready.
You will need to prepare two libraries: [libaom](https://aomedia.googlesource.com/aom/) and [libavif](https://github.com/AOMediaCodec/libavif). cmake is needed in the configuration process. Libaom needs to be not very old, at least v1.0.0-errata1-avif.

diemmarkus:
> I can notify you when the new version is ready.

yes please!

novomesk:
The new version of my plug-in:
http://188.121.162.14/avif/qt_avif_image_plugin_2020-03-20.zip

[...]

PR #499, opened 2020-06-11
Enable AVIF export when supported.

novomesk:
This patch allow saving \*.avif files.
Image sequences based on AVIF (\*.avifs) can be only opened, so I added it only to app_p.openFilters
Related to: https://github.com/nomacs/nomacs/issues/292


diemmarkus:
thanks!

Issue #582, opened 2020-11-18
In AVIF picture with more than one frame only first one is shown


lucaspir7:
Like in this pic:

https://i.fiery.me/IpnR0.avif

It should work like a GIF or a WEBP pic bit it does not.
Only Chrome can open it properly.

novomesk:
It works when you change filename to `IpnR0.avifs`
Problem is that right now nomacs has enabled animations for **.avifs** extension but not **.avif**

In the first AVIF specification, the AVIF sequence had .avifs extension but that's changed recently. In the future, .avif extension will be used both for animated and non-animated images.

It is necessary to enable animations for the .avif extension.

lucaspir7:
Nice. Thanks for the info.

diemmarkus:
thanks!

PR #583, opened 2020-11-19
Enable animation for .avif extension

novomesk:
New AVIF specification draft indicate that ".avif" files can be animated too.
In the past, animated AVIF sequences had ".avifs" extension.

This patch enables animation for .avif extension while keep it enabled for older .avifs files.

Problem was reported in https://github.com/nomacs/nomacs/issues/582


diemmarkus:
thanks! 🥳

Issue #804, opened 2021-08-28
Option to keep date of modyfied images


starbrights:
Sometimes it is need to crop or rotate images. There should be an option not just to keep metadata, but also date of file. Often viewers sort files in order to the date - that fails now.

c0xc:
I have the same problem. Not just sometimes, but almost always because I almost never rotate (and crop) my photos on the same day and when I do, I still want to be able to sort them by date as I consider rotating (and sometimes cropping) to be "minor" changes.

I want to implement this change. I've already digged into Nomacs' behavior when rotating files (#799) which is related this issue because I want to be able to go through a collection of new photos and make minor corrections, i.e., rotate those that are sideways. With my first changes regarding the other bug, rotating an image is now an explicit change, so the user will be asked to save the changes when navigating away. I'm still trying to figure out how/when this "keep date" feature should be active.

It should definitely be opt-in as it would be counter-intuitive to keep the file date after an explicit change. But I don't think it should be a permanent option either because with my workflow, I would want to go through my photos and fix them retaining their dates, all within one session - but when I'm done, close Nomacs, eat lunch and open another image, I'd still expect it to have a new file date after rotating it. It could be from someone else's collection, so I would want to be able to see at first glance that I have changed this file.
So I'm thinking of a session-wide setting, maybe a checkbox/button that has to be clicked once, before editing the first image. What do you think?

leejuyuu:
@starbrights This should have been fixed by #837. Could you try the latest version?

leejuyuu:
Closing as completed by #837.

Issue #799, opened 2021-08-10
Nomacs silently modifies files on disk, destroying timestamp and metadata information.


PhilSalkie:
**Describe the bug**

If a user triggers a "Rotate" by pressing the "R" key, the file on disk is silently modified.  This changes the modification date on the file and changes the file size, all without the knowledge or consent of the user.  The change is immediate, silent, and irreversable, and the "Undo" function is ignored after an "R" key rotation.

**To Reproduce**

`ls -ltr imagefile.png`

Notice the time/date stamp and size.

`nomacs imagefile.png`

Click on image, press "R" key.  Image rotates as expected.
Close Nomacs.

`ls -ltr imagefile.png`

Notice that the time/date stamp is changed to just a few seconds ago, and that the file size may have changed as well.  

**Desktop (please complete the following information):**
 - OS:  Linux, MX-19.4
 - Nomacs V3.14

**Additional context**

There may be other operations which a user would consider "safe" which would destroy timestamp / metadata information on disk.  I have only just found out about the "R" key rotation - that discovery caused an immediate 

`sudo apt-get remove nomacs`

from all my systems, so I won't be looking for more hidden file mangling commands.

DJCrashdummy:
well... i was about to point out [the same "workaround"](https://github.com/nomacs/nomacs/issues/799#issuecomment-896407185) as @PhilSalkie to disable "Save Exif Orientation" in the advanced settings.

**if** there is a function to save changes without asking, this should be a separate (opt-in) option.
not to mention that it is pretty unfortunate that this auto-save is on by default (but this IMHO no-go gets sadly more and more common). :unamused:

c0xc:
Wouldn't it be better to have this option, but keeping the file's timestamp? I'm mostly annoyed be the feature because when I sort my files by date, they're out of order. But if the rotated file would have the old modification date, it could actually be quite useful...?

PhilSalkie:
It's a file viewer.  No viewer should _ever_ modify a file on disk
without the user's knowledge.

Think of forensics, think of file packaging, software version control,
file SHA256 sums - you've just changed the source material just by
looking at it, now the file doesn't match the original - and you don't
know that it's been changed.  Maybe a file is _supposed_ to be 90
degrees out because it's _supposed_ to be shown on an LCD display
that's mounted sideways (like a lot of video signage.)  You're working
on a system, you look at the image and rotate it to fit your screen,
now the image doesn't display properly anymore because Nomacs has just
[...]

PR #837, opened 2022-02-14
Make rotation and comment change explicit (*)

c0xc:
This is a work in progress.

I have removed the auto-save feature/bug, rotating an image is now an explicit change which means the image is marked with an asterisk (*) in the title. See issue #799 for details. Since file comments would've been saved automagically (or reset if you rotate or change the image), I have made that explicit too, so clicking on the save (comment) button now marks the image as edited in the same way.

I also want to add an option to declare rotating and cropping as "minor" changes, the file date would be retained after making a minor change (opt-in obviously). See #804.

Is there anyone left who'd be willing to spare some time to review or test my changes? Comments would be appreciated.

### DIFF
--- a/ImageLounge/src/DkCore/DkBasicLoader.cpp
+++ b/ImageLounge/src/DkCore/DkBasicLoader.cpp
@@ -112,44 +112,74 @@
 namespace nmc {
 
 // DkEditImage --------------------------------------------------------------------
-DkEditImage::DkEditImage(const QImage& img, const QString& editName) {
-	mImg = img;
-	mEditName = editName;
+
+DkEditImage::DkEditImage() : mNewImg(false), mNewMetaData(false) {}
+DkEditImage::DkEditImage(const QImage& img, const QSharedPointer<DkMetaDataT>& metaData, const QString& editName)
+           : mImg(img), mMetaData(metaData), mEditName(editName), mNewImg(true), mNewMetaData(false)
+{
+    //history edit item with modified image
+}
+
+DkEditImage::DkEditImage(const QSharedPointer<DkMetaDataT>& metaData, const QImage& img, const QString& editName)
+           : mImg(img), mMetaData(metaData), mEditName(editName), mNewImg(false), mNewMetaData(true)
+{
+    //history edit item with modified metadata
+}
+
+bool DkEditImage::hasImage() const {
+    //Every edit item has an image, but it may be the old/original one if only metadata has been edited
+    return !mImg.isNull();
+}
+
+bool DkEditImage::hasMetaData() const {
+    return mMetaData;
+}
+
+bool DkEditImage::hasNewImage() const {
+    return hasImage() && mNewImg;
+}
+
+bool DkEditImage::hasNewMetaData() const {
+    return hasMetaData() && mNewMetaData;
 }
 
 void DkEditImage::setImage(const QImage& img) {
-	mImg = img;
+    mImg = img;
 }
 
 QImage DkEditImage::image() const {
-	return mImg;
+    return mImg;
+}
+
+QSharedPointer<DkMetaDataT> DkEditImage::metaData() const {
+    return mMetaData;
 }
 
 QString DkEditImage::editName() const {
-	return mEditName;
+    return mEditName;
 }
 
 int DkEditImage::size() const {
-	
-	return qRound(DkImage::getBufferSizeFloat(mImg.size(), mImg.depth()));
+    
+    return qRound(DkImage::getBufferSizeFloat(mImg.size(), mImg.depth()));
 }
 
 // Basic loader and image edit class --------------------------------------------------------------------
 DkBasicLoader::DkBasicLoader(int mode) {
-	
-	mMode = mode;
-	mTraining = false;
-	mPageIdxDirty = false;
-	mNumPages = 1;
-	mPageIdx = 1;
-	mLoader = no_loader;
+    
+    mMode = mode;
+    mTraining = false;
+    mPageIdxDirty = false;
+    mNumPages = 1;
+    mPageIdx = 1;
+    mLoader = no_loader;
 
-	mMetaData = QSharedPointer<DkMetaDataT>(new DkMetaDataT());
+    mMetaData = QSharedPointer<DkMetaDataT>(new DkMetaDataT());
 }
 
 bool DkBasicLoader::loadGeneral(const QString& filePath, bool loadMetaData, bool fast) {
 
-	return loadGeneral(filePath, QSharedPointer<QByteArray>(), loadMetaData, fast);
+    return loadGeneral(filePath, QSharedPointer<QByteArray>(), loadMetaData, fast);
 }
 /**
  * This function loads the images.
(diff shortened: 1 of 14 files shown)


novomesk:
@c0xc This is valuable piece of work. Are you able to finish it in near future?

I think it would be complicated to merge it later.

Also try to rebase the branch so we see how appveyor Win64 build works.

c0xc:
@novomesk This change is actually complete. I've let it open for testing and also because I thought I'd want to add a feature (for minor modifications) but I'll postpone that.

Sure, I can rebase it on master again. Then I'm marking this as ready for review. And I'll merge it with my crop revert to see if there are conflicts (at that point, we could push things in an "integration" branch for people to test it before it goes into master, if necessary).

diemmarkus (commented):
Hey,

thanks for the change. Two things: auto saving metadata on rotation is currently an option in settings (default on). Does this PR change that setting accordingly? Second: jpgs are newly encoded if we mark images as edited (and therefore lose quality). Is that addressed here?

c0xc:
I cannot mark this as ready yet because it seems I've overlooked something that causes the exif flag to be reset before it's saved to disk. I'll fix that when I have the time and then update this branch.

@diemmarkus Thanks for showing up, your feedback is highly appreciated!

I am aware of this setting. I have not changed the default value, instead I've removed the mechanism that automatically saves the metadata and instead moved it into the regular save workflow (I've also changed the file comment feature in the same way, btw. in some cases it used to reset the file comment). So I've focused on removing the automatic/implicit file update mechanism and making it explicit because implicitly changing a file without asking the user can be dangerous and feels like a bug.
So the setting is still there, after rotating it should change the flag or save the rotated image, but you're right (your second point), if people only want to have the exif orientation changed without reencoding the image, then I guess I'll have to keep and use both states, image edited and metadata changed, the user should be asked in both cases. I cannot use the "trick" that was used before to make it look that way because the ImageLoader that's responsible for rotating would reset the edit history, which essentially caused another bug because people couldn't undo a rotation, yet the metadata was still saved to disk implicitly, changing the files.
So I'll change it to work with both states, but nothing should be saved without asking the user.

(I've actually had the idea to add a setting, opt-in, either permanent or runtime, with which the user could explicitly allow certain types of edits to be made without having to confirm the filename again, to be able to go through a directory of photos or scans and rotate and crop them quickly, regardless if jpeg or bmp, but there'd still have to be a visual indicator that the image has been edited. I'll work on that separately, later, if I have enough time...)

c0xc:
Hi, I have finally had the time to rewrite my changes and update this PR. I hope you guys are still interested.

@diemmarkus As for your comment, I have changed the history to also keep track of metadata changes and if the user has only changed the metadata after rotating the image, the save routine will only update the metadata. It will not save the rotated pixmap, which is only used for display in the gui.

I have not yet marked this PR as ready at the moment because:
* My changes are quite extensive, it would be great if someone wanted to help with testing (or with a code review, if someone has a lot of spare time). Otherwise, I'll poke around until I decide it's good to go.
* Since I've removed the routine that automatically saved all metadata changes in the background, there could still be a couple of routines that change metadata without adding a history item. Adding a history item is now mandatory. I'll have to find such calls.
* Finally, I also want to merge my revert of the crop feature (issue #855), unless anyone disagrees, because personally I would not want to keep working on the version that's based on the new-yet-unfinished crop feature.

Also, I've had to fix how the file comments are handled, so they're also reset on undo now.

@novomesk Feel free to checkout and build my branch to test it, if you want.

novomesk:
The Windows installer with your changes is available at:
https://ci.appveyor.com/project/diemmarkus/nomacs/builds/44331449/artifacts

Try to ask some Windows users to test it what they think.

c0xc:
During my tests, I have found issues when metadata edits are combined with regular edits, so I'll have to make some additional changes. (I never thought I'd have to modify so many parts of the program to fix exif rotation making it explicit...)

novomesk:
Hi @c0xc ,
do you have some progress on this feature?
I'd like to do some changes and I'd prefer not to affect your work.

c0xc:
Hi @novomesk 
Yes, I have basically fixed the last bug and still need to finish something, which should not take too long, I hope I'll have enough spare time next week. Then, I'll mark this PR as ready.

But I have not found someone who'd be willing to check or test my changes, so I hope I haven't overlooked anything. I've had to touch a lot of places to find all the routines that are involved when saving and to extend the history. Of course, if someone finds a bug after my changes have been merged, I'd fix it as soon as I can but as you know, it sometimes takes weeks until I have the time.

c0xc:
Hi @novomesk

I have made some minor adjustments, one was a bug caused by a race condition which caused the modified metadata to be reset right before it would be saved to disk. That's what makes this difficult: Nomacs uses a lot of containers (image loader, image container, basic loader, metadata), then it changes one in the middle to point to a new file or just clears it while other routines are run in the background, possibly interfering with each other... I think if each container would stick with its file until it's destroyed and removing those clear and reload calls (i.e., preventing a reload if it's already loaded with something else) might prevent bugs like that and make the code much more obvious and readable. But that's another story, like reverting the crop feature (will be in another PR).

I have rebased my branch onto the current master and marked this PR as ready.
I think it works ok now. Rotated images are not secretly saved anymore and now there's a real history that allows you to undo a rotation.

There's just one minor glitch that I've found, but it's not serious. It only happens when opening the settings tab, switching back to the image (at which point Nomacs reloads the image containers), then modifying/rotating it and navigating away. The previous image is loaded in the background while the new one is being saved, so when navigating back to that image, the previous one will be shown (restart helps). I'm not sure how I could easily fix that without changing all container classes to keep their state (almost like, atomic) as I mentioned above. I have added some comments explaining what might happen.
But I've seen worse glitches even in Nomacs (like changing the saturation and going to the next image, the modified saturation value is still shown...).

@novomesk How do you feel about merging this? If you have detailed questions about my changes, you can email me.

Delphox:
Hey I installed this PR's [Windows artifact from AppVeyor](https://ci.appveyor.com/project/diemmarkus/nomacs/builds/44780513), and it crashes when trying to load the recent files. Clearing the recent files list from settings.ini prevents a crash until it gets populated again. Here's the crash dump: [nomacs.exe.6092.zip](https://github.com/nomacs/nomacs/files/9612382/nomacs.exe.6092.zip)

novomesk:
Indeed, viewing recent files triggers the crash. Thank you for testing @Delphox 

@c0xc please investigate the problem.

c0xc:
@Delphox Thanks for testing! Unfortunately, I cannot reproduce the bug. I can start Nomacs with the recent files view, even if some of them are missing. However, I was able to cause it to crash when opening a huge directory with thousands of files. This crash seems to be related with the Exiv2 module. I'll have to investigate that. Does it only happen when opening a huge directory (or when clicking the first of the three buttons, labeled "Load the Directory", on an item in the recent files view)?

Delphox:
@c0xc I did more testing and it seems to have to do with certain images rather than just recents, or quantity (although quantity makes it more likely to run into one that crashes, of course). I'll upload two blank examples, one jpeg and one png, that causes recents to crash.

[Pictures.zip](https://github.com/nomacs/nomacs/files/9640925/Pictures.zip)

c0xc:
@Delphox Thanks for the sample, that helps! I was able to reproduce the crash with it (btw. it actually was the same crash). It was caused when copying the metadata object to put it into the history. Metadata is loaded in one of two ways and in one case, the reload would fail. I've fixed that in a way so that I don't need this reload at all anymore. This should also help in case an image is deleted and then modified.
Would you please test it again?

Delphox:
@c0xc No crashes here! Loaded with my regular settings full of all sorts of recent files. Good work! 😉

c0xc:
@Delphox That's good! Have you also tested my new feature(s)? Mainly: Rotate an image without Nomacs automatically saving the rotated file without asking you (if you've disabled exif rotation in the settings, enable it first). (Very annoying if you sort your files my modification time.)
Or rotate an image and then undo it.

Delphox:
> @Delphox That's good! Have you also tested my new feature(s)? Mainly: Rotate an image without Nomacs automatically saving the rotated file without asking you (if you've disabled exif rotation in the settings, enable it first). (Very annoying if you sort your files my modification time.) Or rotate an image and then undo it.

Yup, works as intended!

Delphox:
@c0xc I used your version for a while longer and realized it can't view webps. Can you check?

novomesk:
Unfortunately I am still getting crashes with the new build from `Sun Sep 25 20:42:48 2022`

c0xc:
@novomesk I have noticed that the Exiv2 module doesn't want to recreate some objects (when copying), so I've added a workaround for that. But I don't know if that's what caused the crash in your case. If it's still happening, can you say when it's happening, if affects just some file types or something else?

@Delphox It seems Nomacs used to have a custom WEBP loader but it's been removed as Qt has support for WEBP now, but that might depend on how your Qt build was configured (esp. relevant if you're building Nomacs manually).

novomesk:
Now it doesn't crash for me but let us keep testing for a while.

Delphox:
> @Delphox It seems Nomacs used to have a custom WEBP loader but it's been removed as Qt has support for WEBP now, but that might depend on how your Qt build was configured (esp. relevant if you're building Nomacs manually).

@c0xc I've been using the autobuild from AppVeyor. Interestingly, the latest version that fixed crashes for @novomesk also fixed webps for me, so I guess the two problems were sorta related? I'll keep using it as well to see if any other problems arise.

novomesk:
@c0xc Are you going to work on crop feature?

I want to focus on dealing with deprecated code warnings so the code will be prepared on future updates.

c0xc:
Thanks for merging! This was a massive change.

@Delphox Could be, it depends on the image (don't have the time to check atm), the main thing is that it's working now. If you find a bug, let me know.

@novomesk Actually, I had already prepared the crop revert but I'll have to merge it with this change now. I'll open a separate pull request for that.

PR #866, opened 2022-07-07
Translate exif metadata into human readable text

l-berg:
exiv2 provides a function to translate raw exif values into a human readable form.
This PR extends getNativeExifValue with the argument "humanReadable" to use human readable strings in the Meta Data Info panel.

For example, this translates Exif->Fujifilm->WhiteBalance from **0** to **Auto**.

### DIFF
--- a/ImageLounge/src/DkCore/DkMetaData.cpp
+++ b/ImageLounge/src/DkCore/DkMetaData.cpp
@@ -278,7 +278,7 @@ DkMetaDataT::ExifOrientationState DkMetaDataT::checkExifOrientation() const {
 	if (mExifState != loaded && mExifState != dirty)
 		return or_not_set;
 
-	QString orStr = getNativeExifValue("Exif.Image.Orientation");
+	QString orStr = getNativeExifValue("Exif.Image.Orientation", false);
 
 	if (orStr.isEmpty())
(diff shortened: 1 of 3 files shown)


novomesk:
Another question:

Will this work with older versions of exiv2 or we need to bump the requirements of the exiv2 version?

novomesk:
@c0xc Please take a look at this and tell us if you have any objections.

VJSchneid:
> Another question:
> 
> Will this work with older versions of exiv2 or we need to bump the requirements of the exiv2 version?

To translate the exif values it was only necessary to call the ``operator<<(std::ostream&, exiv2::value&)`` instead of ``value.toString()``. This operator seems to be very well supported as it can be find in [very old versions in the exiv2 repository](https://github.com/Exiv2/exiv2/blob/v0.5/src/value.hpp#L199).
I did not try to build nomacs using some older exiv2 version but from what I see this shouldn't be an problem.

novomesk:
Thanks for contribution!

c0xc:
> @c0xc Please take a look at this and tell us if you have any objections.

So this would print "No, compulsory" instead of "16" for Flash or "JPEG (old-style)" instead of "6" for Compression. Good idea, I like it!

Issue #850, opened 2022-04-10
Revert crop feature to v3.16 (until there's a better one)


c0xc:
**Is your feature request related to a problem? Please describe.**
Hello `<void>`. I am about to revert the new crop tool from v3.17 because it does not allow precise cropping anymore. If anyone has strong negative feelings about this, please leave a comment here.

In the old version you would select a section of the image, which you were able to zoom to make a precise cut. In the new version, it basically works the other way around, so you'd have a white grid centered in the middle of the window and instead of moving that section, you are moving the full image in the background. If you try to zoom out, the grid moves. Sometimes the grid jumps around, I haven't yet figured out when and how, exactly. But it often jumps around, so you'd end up cropping a different section of the image. All I can say is that cropping an image, something which used to be so simple that I never really thought about it, is now very difficult or, if the selection keeps jumping around, impossible.

**Describe the solution you'd like**
I understand that a lot of effort has been put into this new tool and I'm hesitant to revert someone else's work. But to be honest, v3.16 had a crop feature that worked just fine, you could easily select an area, zoom in to make sure you're not cutting something important off, zoom back out and confirm. I still want to be able to crop like that.

**Describe alternatives you've considered**
I've compared both versions. Obviously, my first thought about fixing the new crop feature is to make it work like the old one. But it seems like this would require much more work than reverting the old one. If I have more time later on, I might try to fix it, but for now, I'd rather improve other things and revert this, at least in the master branch.
There are several minor annoyances that could also be improved. For example, in v3.16 when clicking outside of the cropped area, you'd accidentally rotate it. I'm not sure why rotating is integrated into the crop feature at all but it could probably be fixed to rotate only if the corners are clicked. On the other hand, the new crop tool in v3.17 has a slider at the bottom, which only allows rotating up to 45 degrees, apparently. Then, there's a separate rotate feature (Edit menu), which shows a slider in the panel at the right side of the window, but the slider cannot be adjusted using the keyboard (rotating one more degree...) and when clicking on it, it jumps around because there's a preview underneath it which changes in size.
Either way, some of those things could probably be improved but those improvements would not be worth it if the new crop tool stayed the way it is now.

**Additional context**
https://github.com/nomacs/nomacs/issues/345
https://github.com/nomacs/nomacs/issues/547

c0xc:
Video showing the new, buggy crop feature, how the section moves when zooming:
[...]

PR #897, opened 2022-10-17
Revert crop feature to last working state

c0xc:
This reverts the old crop feature from version 3.16.1711 as announced in issue #850. In short: The new crop feature doesn't allow precise cropping anymore, mostly because you can't zoom anymore.

It's important to note that, because this affects many commits, I've tried to identify all commits after the last working state that somehow affected the crop feature and reverted them. In other words: If there were any unrelated improvements mangled into those commits (which should not be the case), those improvements could now be lost.


c0xc:
@novomesk fyi unfortunately this is currently stale because the code cleanup and refactoring caused many merge conflicts and I currently don't have the time to go though all of them. The draft state I have in my branch (from last year) works well as far as I can tell but it doesn't include the refactoring.

novomesk:
Closing because changes are implemented now.

PR #900, opened 2022-10-24
revamed the manpage to include all the arguments of --help

mmahmoudian:
In this PR I've done few things:
1. very minor changes to the manpage format to have proper tags
2. added all the flags defined in the `--help` to the manpage
3. Added the website and github repo URLs to the manpage
4. updated the "update date"

On a Linux machine the manpage can be viewed without compiling via:

```sh
man ./ImageLounge/xgd-data/nomacs.1
```

and it should look like this (of course the colors are because of me using `bat` as my `${MANPAGER}`):

image


novomesk:
Looks good but I would not add reference to the nomacs.org website because practically, we cannot update it.

PR #925, opened 2023-02-22
Revert crop feature to last working state (2nd attempt)

c0xc:
This is the 2nd attempt of my previous PR:
https://github.com/nomacs/nomacs/pull/897

The first PR couldn't be merged due to a long list of conflicts. I've applied a new strategy and was able to get those conflicts fixed while keeping my [autosave bug](https://github.com/nomacs/nomacs/pull/837) changes. At first glance, it seems to build now but others should test it as there's a chance I missed a few things going through all of those conflicts. Both of my changes (crop revert + autosave bug) as well as changes that were made since should be briefly tested to make sure they're not lost. I don't have much more time for tests or fixes now as I'm currently busy with new projects.


novomesk:
Try to use `clang-format` on changed .cpp and .h files, it will unify the code style.

For example:
`clang-format -i --style=file DkMetaData.cpp`

c0xc:
I ran `clang-format` on those 4 files.

novomesk:
I think all cpp,h files need clang-format to see less differences.

c0xc:
Well, I've run clang-format on those other files that show a lot of differences but it only eliminated a couple of diff lines. Do you have a specific file in mind that's showing too many diffs? Keep in mind I had to go through all old commits, identify those that contain changes of the crop feature and revert all of them, so naturally that's a lot (btw. this process took more than a week).

novomesk:
I am trying to look into this.
Are the changes in
`scripts/utils/fun.py` and
`scripts/utils/version.cache`

necessary?

c0xc:
I increased the version number and fixed a minor bug in this Python module you mention that's used to write the version to the source files because it would sometimes append a linebreak, breaking the code. My change prevents it from doing that.

c0xc:
Well, the Python fix is a oneliner but I can open a separature PR for it if you prefer that.
Just one question: If I remove my commit which increased the version to 3.18.0, it would not stay the same but instead it would be decreased to 3.16.1795. I thought that would not only be confusing but could also cause trouble for packagers. That's why I changed it and since there were many changes, I also increased the minor version to 18.

So do you really want me to remove this commit, going back to version 3.16.1795?

novomesk:
Make a new PR with the change in the python script.

Increase the version number in the new PR too, but not 3.18.0 yet because the even numbers were used for stable releases. 3.17.newnumber is fine so far.

Let me know if https://github.com/nomacs/nomacs/pull/930 works for you. I'd prefer to merge that one after you test it.

c0xc:
Sure, here you go:
https://github.com/nomacs/nomacs/pull/932
Let me know if that's ok and if I should rebase it on your current working branch.

I've removed the commit from my feature branch in this PR.

I didn't test the Windows binary because I don't have a suitable Windows vm ready at the moment, but I cloned your [feature branch](https://github.com/novomesk/nomacs/tree/pse/crop_revert_rebase_20230222) and built that. It seems to work fine, the (old) crop feature is restored and my rotation fix is still in there, too.

It would be great if someone who dislikes the new buggy crop feature would test it. My [bug report](https://github.com/nomacs/nomacs/issues/850) only got a single upvote but no comments. I did see a couple of complaints on the [website](https://nomacs.org/new-cropping/) but I guess it's unlikely any of those users will come here and test it.

novomesk:
Closing because changes are implemented and merged with slight modifications.

PR #930, opened 2023-03-08
Revert crop feature to last working state

novomesk:


novomesk:
@c0xc Please take a look at this, I continued with your changes, made a little clean-up and formatting changes.

This is the installer to test:
https://ci.appveyor.com/api/buildjobs/ejjqy7nj04qbkduc/artifacts/installer%2Fnomacs-setup-x64.msi

It would be good to find someone to test it.

Mine181:
I've tested this dev build for a few minutes, and while most things work well, there are a few issues I've noticed:
- Extracting images from an archive will make a folder based on the archive format, which isnt indicated in the extract window, and overwrites any images with the same file name (if an image was called image1.png for instance)

- Restarting the application prompts a "Do you want to save your tabs?" window, yet starts a new instance of the program without need of confirmation

- Number of threads can be set to a value exceeding that of the CPU's number of threads.

- Deleting a pasted screenshot from the clipboard does not work yet doesn't display an error.

- Fullscreen view shows the properties of the previous image if a non-saved screenshot is pasted

- Closing the only/last tab and then Choosing the wallpaper option crashes the program

- Right clicking a selected image deselects it

Another thing I've found but don't understand the reason for is that after randomly selecting some options, the program wouldnt load images when selecting it from a directory or when cycling through with the arrows.

novomesk:
Thank you @Mine181 for the testing.

What is good that it seems there are no new problems.

c0xc:
@novomesk I also thought about squashing my commits at some point but I decided against it because it would make it much more difficult to find out later which commits were reverted.

@Mine181 Thanks for your effort. During my many weeks trying to understand the sometimes overengineered code, I also found things like copy&paste are buggy. But have you tested the crop feature? If you have any current build at hand which already has the broken crop, compare it to this one.
There's a short video:
https://github.com/nomacs/nomacs/issues/850

PR #940, opened 2023-04-23
Initial port to Qt6

v-tyrtov:
The code now is compatible with Qt6

Corresponding PR for plugins is needed https://github.com/nomacs/nomacs-plugins/pull/40

Tested only under Linux environment

To build with Qt5 define
QT_VERSION_MAJOR=5
during cmake configuration stage


v-tyrtov:
Thanks for the comments. 
And thanks for merging https://github.com/nomacs/nomacs/pull/946

I had some time to fix most of warnings for Qt 5.15 and made the commits more readable.
https://github.com/nomacs/nomacs/pull/948

Not sure that I'll have enough time for significant contributions but still prefer to move the project to Qt6.

Could you please clarify about the plugins - wouldn't it be enough to merge https://github.com/nomacs/nomacs-plugins/pull/40 ?

novomesk:
> Could you please clarify about the plugins - wouldn't it be enough to merge [nomacs/nomacs-plugins#40](https://github.com/nomacs/nomacs-plugins/pull/40) ?

I don't have access to https://github.com/nomacs/nomacs-plugins
It should be forked to new one.

v-tyrtov:
Thanks for the clarification. 
In such case wouldn't it be better to merge the plugins to the main repo?
So we'd be able to get rid of submodules (at least partially).
It seems that there is no any sense to keep the plugins separately if they aren't used anywhere else. 

What do you think?

PR #946, opened 2023-05-10
Fix warnings and partially improve obsolete code

novomesk:


novomesk:
Hi @c0xc ,
if you have some time, can you test these changes if you encounter some regression or new problem.
They are mostly edits from @v-tyrtov but not all of them yet (he wanted to add Qt6 support).

PR #982, opened 2023-09-03
Fix build with exiv 0.28

listout:
With exiv 0.28 we're getting error such as:

> ImageLounge/src/DkCore/DkMetaData.h:151:23: error: ‘AutoPtr’ in ‘class Exiv2::Image’ does not name a type
>   151 |         Exiv2::Image::AutoPtr loadSidecar(const QString& filePath) const;
>       |                       ^~~~~~~
> ImageLounge/src/DkCore/DkMetaData.h:160:23: error: ‘AutoPtr’ in ‘class Exiv2::Image’ does not name a type
>   160 |         Exiv2::Image::AutoPtr mExifImg;
>       |                       ^~~~~~~

...and so on.

First discovered in Gentoo Linux.

Bug: https://bugs.gentoo.org/906488

### DIFF
--- a/ImageLounge/src/DkCore/DkMetaData.cpp
+++ b/ImageLounge/src/DkCore/DkMetaData.cpp
@@ -69,7 +69,11 @@ QSharedPointer<DkMetaDataT> DkMetaDataT::copy() const
         // ImageFactory::create(type) may crash even if old Image object has that type
         try {
             // Load new Exiv2::Image object
+#if EXIV2_TEST_VERSION(0, 28, 0)
+            Exiv2::ImageType i_type = mExifImg->imageType();
+#else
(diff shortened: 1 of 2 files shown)


listout:
Also can someone make a 3.17 tag for nomacs/nomacs-plugins. Helps in maintaining the package for us.

novomesk:
Thank you for this patch, it is a valuable feedback.

However, it won't be accepted in this form because the appveyor build failed (it uses old version of exiv2 and MSVC on Windows).

Meanwhile, I am preparing alternative fix in https://github.com/nomacs/nomacs/pull/983

I don't have rights for https://github.com/nomacs/nomacs-plugins , most probably there won't be updates there. Only if we create some fork.

Issue #951, opened 2023-06-08
Build broken with exiv2 0.28.0


cbinner:
Building version 3.17.2282 fails with the new exiv2 version I got through the repository updates today. This is the error message:

`In file included from /home/xxx/.cache/yay/nomacs/src/nomacs-3.17.2282/ImageLounge/src/DkGui/DkControlWidget.cpp:32:
/home/xxx/.cache/yay/nomacs/src/nomacs-3.17.2282/ImageLounge/src/DkCore/DkMetaData.h:158:19: error: 'AutoPtr' in 'class Exiv2::Image' does not name a type
  158 |     Exiv2::Image::AutoPtr loadSidecar(const QString &filePath) const;
      |                   ^~~~~~~
/home/xxx/.cache/yay/nomacs/src/nomacs-3.17.2282/ImageLounge/src/DkCore/DkMetaData.h:167:19: error: 'AutoPtr' in 'class Exiv2::Image' does not name a type
  167 |     Exiv2::Image::AutoPtr mExifImg; // TODO std::unique_ptr<Exiv2::Image> (and all other *::AutoPtr)
      |                   ^~~~~~~
make[2]: *** [CMakeFiles/nomacsCore.dir/build.make:239: CMakeFiles/nomacsCore.dir/src/DkGui/DkControlWidget.cpp.o] Error 1
make[2]: Leaving directory '/home/xxx/.cache/yay/nomacs/src/nomacs-3.17.2282/build'
make[1]: *** [CMakeFiles/Makefile2:242: CMakeFiles/nomacsCore.dir/all] Error 2
make[1]: Leaving directory '/home/xxx/.cache/yay/nomacs/src/nomacs-3.17.2282/build'
make: *** [Makefile:156: all] Error 2
make: Leaving directory '/home/xxx/.cache/yay/nomacs/src/nomacs-3.17.2282/build'
==> ERROR: A failure occurred in build().
    Aborting...`

BTW the installation instructions on the website could use a little correction, in Arch Linux nomacs has been moved to the AUR.

iq2luc:
Hello,

the attached patch [nomacs-fix-exiv2-0.28.patch.txt](https://github.com/nomacs/nomacs/files/11703502/nomacs-fix-exiv2-0.28.patch.txt) fixes the issue on my side.
There is also a variant of this patch for `nomacs-3.17.2206` (see [AUR](https://aur.archlinux.org/packages/nomacs#comment-918456)).

[...]

Issue #952, opened 2023-06-11
Nomacs NOT Working at Arch Linux


velan001:
In few days nomacs are not opening in my system . I will check where is the problem by opening nomacs through terminal the output is nomacs: error while loading shared libraries: libexiv2.so.27: cannot open shared object file: No such file or directory

My System is Arco Linux based on Arch.

FabioLolix:
@velan001 every time you get an error like
 `error while loading shared libraries: libXXX.so.XXX: cannot open shared object file: No such file or directory`
mean you need to rebuild the pacakge, the AUR pkgbuild have already been updated with a patch for building with the new exiv2 version when you posted

also don't use caps lock when writing

iq2luc:
Hello @velan001, see https://github.com/nomacs/nomacs/issues/951#issuecomment-1584107223 for a working PKGBUILD.

FabioLolix:
This has been resolved and can be closed

PR #983, opened 2023-09-04
Update metadata code

novomesk:
@c0xc 
I would appreciate if you have some time to try to build this and make few tests (if there is no regression, crash etc...)


cbinner:
You may want to close #951 and #952, those are duplicates.

PR #1000, opened 2023-10-28
Fix language name for zh_tw

peepeetee:
Fix language name for zh_tw

As is, the string referrers to zh_cn, which is probably a conversion alias

- Fork [nomacs](https://github.com/nomacs/nomacs.git) and create your branch from `master`
- Pull requests are only accepted if they pass [travis](https://travis-ci.org/nomacs/nomacs)

also have a look at our [CONTRIBUTING.md](https://github.com/nomacs/nomacs/blob/master/CONTRIBUTING.md)


scrubbbbs:
"簡體中文" translates to "Simplified Chinese"
"正體中文" translates to "Traditional Chinese"

zh_tw locale is for traditional Chinese,  so the change makes sense.

PR #1015, opened 2024-01-22
Add batch processing transform options: resize zoom and crop center

bigjim80:
Added 2 new features to the batch processing "Transform" page:

1. A new resize option called "Zoom". This resizes the image to a given rectangle, by taking the largest possible part of the image to have no black borders. This means that if the aspect ratio of the image is different than the given rectangle, a piece of the image (top/bottom or left/right) will be cut away.
2. The possibility to crop the image centered instead of a fixed x and y top-left coordinate.

Example of use: to convert holiday pictures to show on a desktop without black borders.

nomacs


novomesk:
@leejuyuu Would you like to test this new/proposed feature? I need someone to confirm that everything works and the old features are not affected.

If you are using Windows, there is installer or portable version available at https://ci.appveyor.com/project/diemmarkus/nomacs/builds/49003740/artifacts

@bigjim80 Thank you for you work. However, this is your first contribution to nomacs. I'd like to know if you are willing to help in the future too. For example if you can assist in the future in case there is an issue with the batch processing code which was altered, etc.

bigjim80:
@novomesk Thank you for considering this pull request! Of course, if it causes future issues or enhancements, I'll be very glad to help to fix them. If there is anything I could do, feel free to ask but my c++ experience for desktop platforms (Qt and other libraries) is not that great. Also time is quite limited unfortunately.  But I like nomacs very much so I will support where I can...

leejuyuu:
I tested the new features (on Arch Linux). The resize and crop works, but note that I haven't used batch processing before so there might be things that I did not catch. Here's some questions that I have:

1. The new "Zoom" resize feature comes with crop by default, is it clear to put it in the resize block? I think we need documentation for this.
2. When "Zoom" is set, the "Crop Rectangle" setting is ignored. This might confuse the users. Also, this makes "Zoom" always place the crop rectangle at the center of the image, and the user is never able to change its position.
3. Using "Zoom" with 90-degree rotation produces black edges.
4. "Crop to center" makes the "x" and "y" in the "Crop Rectangle" have no effect. It would be nice to disable/gray out the x and y options or make "Crop to center" a button that inputs the correct x and y values automatically.
5. Maybe "Crop to Center" to match the other options?

bigjim80:
Thanks for the testing & feedback!

1&2&5: You are totally right, using zoom and crop at the same time makes less sense and is confusing indeed. My suggestion is to disable (gray out or hide) the cropping controls when "zoom" is selected. Maybe there is a better name instead of "zoom" to auto-describe the resize & crop-to-fit function.

3: This is a bug indeed, i can fix that.

4: I thought of this, but the 4 controls of the cropping rect are grouped into 1 custom control, so it would need modifications to that control (DkRectWidget) as well. Can be done of course.

Would you like me to commit these extra fixes? Or do you prefer to drop this pull request entirely?

leejuyuu:
Thanks for the reply! It would be great if potentially conflicting options could be disabled.

Not sure whether this is feasible, but if you plan to modify the crop rectangle controls, would it be possible to separate the resize and crop-to-fit functions? Maybe like providing a set of inputs to specify an aspect ratio and a button that updates the width and height to the maximum possible size. Then the resize can behave as previously. This pops up after reading your reply but sounds more complicated. I'm totally fine with changing "zoom" to a more descriptive name and simply disabling the crop options.

As for whether to drop this PR, I am fine with either choice. @novomesk what's your preference?

novomesk:
Add a new commit to this PR and we can squash them during merging.

leejuyuu:
Thanks for the patch! Maybe it's just me, is it usually clear that a check box is false when it is grayed out, even if it is still checked? This happens when
- Check crop rectangle -> check crop center -> uncheck crop rectangle
- Check crop rectangle -> select zoom

bigjim80:
You're right, that is indeed more clear to show that checkbox has no effect.
The same happens when 'Crop to center' is checked, then the x and y spinboxes are disabled but still showing a value that is not used.
I've fixed both in this commit.

PR #817, opened 2021-10-15
Add:On Linux system support open image in the Trash

ZTFtrue:
Please check the following before submitting a *pull request*:

- [x]  Fork [nomacs](https://github.com/nomacs/nomacs.git) and create your branch from `master`
- Pull requests are only accepted if they pass [travis](https://travis-ci.org/nomacs/nomacs)

also have a look at our [CONTRIBUTING.md](https://github.com/nomacs/nomacs/blob/master/CONTRIBUTING.md)

-------------------------------

PR #1030, opened 2024-02-15
Add:On Linux system support open image in the Trash

ZTFtrue:
Add a feature on Linux system to support open image in the Trash. Support KDE, gnome, lxde-gtk3, xfce4. 

- Fork [nomacs](https://github.com/nomacs/nomacs.git) and create your branch from `master` [checked]
- Pull requests are only accepted if they pass [travis](https://travis-ci.org/nomacs/nomacs)  [checked]

also have a look at our [CONTRIBUTING.md](https://github.com/nomacs/nomacs/blob/master/CONTRIBUTING.md)

### DIFF
--- a/ImageLounge/src/DkCore/DkUtils.cpp
+++ b/ImageLounge/src/DkCore/DkUtils.cpp
@@ -62,7 +62,6 @@
 #include <QUrl>
 #include <QtConcurrentRun>
 #include <qmath.h>
-#include <QMessageBox>
 
 #include <QSystemSemaphore>
 #pragma warning(pop) // no warnings from includes - end
--- a/ImageLounge/src/DkCore/DkUtils.cpp
+++ b/ImageLounge/src/DkCore/DkUtils.cpp
@@ -62,6 +62,7 @@
 #include <QUrl>
 #include <QtConcurrentRun>
 #include <qmath.h>
+#include <QMessageBox>
 
 #include <QSystemSemaphore>
 #pragma warning(pop) // no warnings from includes - end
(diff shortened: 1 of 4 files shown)


novomesk:
Hello @ZTFtrue ,

In case there is some issue with this code in the future, are you willing to help finding fixes/solutions?
Do you have plans regarding other contributions to nomacs?

ZTFtrue:
I'm willing to help finding fixes/solutions for this code in the future.

I don't think I have enough time to regarding other contributions, but if i
have free time, I will contribute code to nomacs.

ZTFtrue:
> The link to trash items seems to be specific to the file manager instead of DE. I am on Arch with Sway WM and nautilus behaves differently from dolphin. Without your patch, Dolphin resolved the path for me when I attempted to open a trashed image with Nomacs, but Nautilus did not.

gnome_dolphin
 
That's because Dolphin can return real path in Gnome DE(just for Gnome DE)

-------------

> Did you test with the default file managers for each DE mentioned? (Nautilus for Gnome, Dolphin for KDE, PCManFM for LXDE, and Thunar for XFCE4) Nautilus, PCManFM, and Thunar use gio/gvfs to manage trash, Dolphin uses a different mechanism that's called KIO slaves

Yes, I tested it, this is the result.

gnome dolphin: <https://github.com/nomacs/nomacs/assets/13563890/ad273907-d610-4e17-bf1e-b0c0e323c975>  
gnome dolphin: <https://github.com/nomacs/nomacs/assets/13563890/98ef372e-ca18-4506-930d-462a0ac5e0c7>  
gnome nautilus: <https://github.com/nomacs/nomacs/assets/13563890/e5a3d809-26e2-4cf4-8a92-543daaeaeeb0>  
Kde dolphin: <https://github.com/nomacs/nomacs/assets/13563890/a07908a2-8f70-413e-8edb-a3186de48e74>  
kde nautilius: <https://github.com/nomacs/nomacs/assets/13563890/56636173-b4cb-4495-a399-ab51eddebb59>  
kde PCManFM: <https://github.com/nomacs/nomacs/assets/13563890/c61a1b01-1879-40bc-b40c-675483d4fb9c>  
Kde Thunar: <https://github.com/nomacs/nomacs/assets/13563890/3fe517a9-4f1a-4261-b162-8708084ad793>  
lxde PCManFM: <https://github.com/nomacs/nomacs/assets/13563890/0a0ea565-d38b-4ae7-bf68-ac9c83585e31>  
XFCE4 Dolphin: <https://github.com/nomacs/nomacs/assets/13563890/ff80c769-5a9f-4928-a630-6b79013bd21f>  
XFCE4 Thunar: <https://github.com/nomacs/nomacs/assets/13563890/d1c28d72-8948-4422-8d4e-236865ee8d1a>

leejuyuu:
> That's because Dolphin can return real path in Gnome DE(just for Gnome DE)

I do not think so. I am not using Gnome.

ZTFtrue:
> I do not think so. I am not using Gnome.

`just for Gnome DE` is a mistake.

 Dolphin will return the real path in some DE, But this feature will not deal with the real path.

leejuyuu:
I'm sorry for sounding grumpy. Allow me to explain some of my thoughts.

This feature feels like a "nice-to-have" for me. The problem can already be solved by restoring the trashed image, then nomacs can open it as a normal file. Note that I'm pretty biased since I have stopped using graphical file managers for a while.

Adding support for opening items from the Trash of these file managers will couple nomacs to the handling of the URL schemes, fuse mount points, etc. GTK and KDE applications can understand these URLs because they use `gio` and `kio` libraries internally. I could not find clear documentation on how the trash URLs or paths are formatted and resolved, so I suspect that it is expected to use the libraries directly and leave those as implementation details. I feel concerned about using something abstracted away from us by parsing the URLs. 

However, if we understand how these things work, we could probably also support other schemes like `smb://` instead of just `trash://`. I'm currently thinking on how things does not work as described in [this reddit comment](https://www.reddit.com/r/kde/comments/mqgfo4/comment/gui9g3h/?utm_source=share&utm_medium=web2x&context=3).

ZTFtrue:
the first pr: https://github.com/nomacs/nomacs/pull/817
I've read the source code for `KDE` and gnome's default image viewer in the past, and I can't find any clear documentation on how the trash URLs or paths are formatted and resolved, unless using the `gio` or `kio` libraries, but this makes difficult to develop and use.

[So I tested this feature in any DE and file manager I know so that it can support as many DE as possible.]( https://wiki.archlinux.org/title/Desktop_environment) And, I found that there are only two situations, KDE environment and Gnome environment ( Maybe, because many desktop  fork of GNOME).

----------------------

For  other schemes,  maybe we should develop new features.

--------------------
Add On:

`gio info` can't support `trash URLs`

leejuyuu:
> I can't find any clear documentation on how the trash URLs or paths are formatted and resolved, unless using the gio or kio libraries

This is what I am worrying about. If it is an implementation detail hidden behind some libraries, it could be changed, and nomacs would need to track and adapt to that change. Also, there might exist different situations that have not been considered in this PR.

scrubbbbs:
It is an odd feature that made it's way into Linux file managers. I think it is the file manager's job to do the conversion to an ordinary path or symlink if it is going to expect anything to work with it, as KDE has done with dolphin. In which case there really isn't anything we need to do here.

Dolphin, PCManFM (Xfce), Krusader all work. The GNOME/GVFS based ones are broken, perhaps by design. Nautilus will refuse to use nomacs even if it is the default application. Perhaps it sees the "%f" in the .desktop entry and says "you don't take a URL argument so it won't work -- sorry to break every application written in the last 30 years"

The right way to support it is to get them to fix their code or else use their API  in nomacs. If they can detect that a URL isn't supported then they can send a properly formatted path too.

scrubbbbs:
Closing since we probably can't support this due to popular desktops/file managers in a constant state of flux. It would just become another broken feature in a few years. If there was a stable API for this that would be the way to go.

Issue #1034, opened 2024-02-23
Compiling Nomacs from downloaded zip gets "not a git repository" error


LinuxBeaver:
`-- The C compiler identification is GNU 13.2.1
-- The CXX compiler identification is GNU 13.2.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/lib64/ccache/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/lib64/ccache/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test COMPILER_SUPPORTS_CXX11
-- Performing Test COMPILER_SUPPORTS_CXX11 - Success
-- Performing Test COMPILER_SUPPORTS_CXX0X
-- Performing Test COMPILER_SUPPORTS_CXX0X - Success
-- A release build (non-debug). Debugging outputs are silently ignored.
-- Found PkgConfig: /usr/bin/pkg-config (found version "1.8.0") 
-- Checking for module 'exiv2>=0.27'
--   Found exiv2, version 0.27.6
-- exiv2 version by pkg-config: 0.27.6
-- exiv2 contains exiv2.hpp, using it
-- Found OpenCV: /usr (found version "4.7.0") found components: core imgproc 
-- Checking for module 'libraw>=0.12.0'
--   Found libraw, version 0.21.2
-- Found TIFF: /usr/lib64/libtiff.so (found version "4.4.0")  
-- QUAZIP: QuaZip-1 for Qt5 found
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Traceback (most recent call last):
  File "/home/contrast/Desktop/nomacs-master/ImageLounge/../scripts/versionupdate.py", line 141, in <module>
    update(args.inputfile, args.copy)
  File "/home/contrast/Desktop/nomacs-master/ImageLounge/../scripts/versionupdate.py", line 42, in update
    l = add_git_tag_string(l)
`

The warning `Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).` got my attention. 

While compiling my disk only had 700mb of diskspace. I am assuming this is the cause and if I free up more disk space it will work.

[...]

PR #1037, opened 2024-02-28
Version git revision

leejuyuu:
Building Nomacs from the zip source archives
downloaded from Github fails because the `versionupdate.py` script
calls Git command, which assumes the working direction is a child of
a Git repo.

Move the revision tag logic to the CMake script. Check for the existence
of a `.git` directory to fix the issue.

Also, include the commit date and check whether there are modified files
in the repository. These could be useful when reporting and searching
for bugs. Assemble the result into a constant string for use.
Fixes https://github.com/nomacs/nomacs/issues/1034.

The NOMACS_GIT_TAG was defined but not referenced anywhere. Add a new
line in the version info of the about splash screen to display the
updated revision string, which contains the commit hash, the commit
date, and whether the source has modifications that have not been
committed when building. This could be helpful for reporting bugs.


leejuyuu:
The "about nomacs" now displays the Git revision
image

When built from the zip, it will display "unknown"
image

But, strangely, the appveyor build gets a non-existing git hash
image

ImageLounge/CMakeLists.txt
```diff
@@ -162,8 +162,39 @@ IF(NOT ENABLE_PLUGINS)
 	LIST(REMOVE_ITEM GUI_HEADERS ${CMAKE_CURRENT_SOURCE_DIR}/src/DkPluginManager.h)
 ENDIF(NOT ENABLE_PLUGINS)
 
+# Try to get revision from Git head
+set(NOMACS_REVISION "unknown")
```
novomesk, line 166: Instead of "unknown" can we use empty string?

leejuyuu: It was updated.

PR #1049, opened 2024-03-16
Use Exiv2::CommentValue to format Exif.Photo.UserComment metadatum

pinotree:
Converting the UserComment exif metadatum to string will result in its direct/quasi-internal string representation of libexiv2, which may include a `charset=...` prefix with the charset of the value.

Since we want the actual content/value of `UserComment`, and the `Exiv2::Value` held by `Exiv2::Exifdatum` is `Exiv2::CommentValue`, then cast it to call `comment()`. The result is converted to `QString` using `QString::fromStdString()`, which converts std::string as UTF-8 string.

For further details, see also the exiv2 ticket:
https://github.com/Exiv2/exiv2/issues/1258


novomesk:
Hello @pinotree ,
Thank you for the willingness to contribute.
In case there is some problem with this change in the future, would you be willing to help finding fixes?

PR #1050, opened 2024-03-16
Change request URLs to use HTTPS

leejuyuu:
Nomacs used HTTP to request for things like version files for auto updates. Because the version file contains URL, if the file gets tempered, the user could end up downloading malicious files. Modify the URLs used in the code so that requests use HTTPS by default, which should be safer.


novomesk:
It seems it works, we can merge when you make it ready.

leejuyuu:
@novomesk  This is ready!

Issue #944, opened 2023-05-04
Wrong Coordinates in Menubar when using Crop (V 3.16.1709)


SystemSpam:
Hello.

I found NoMac Image Lounge when I was looking for a cross-platform alternative for IrfanView.
I like your tool very much, but I have found an unsightly behavior when i mark a area for crop.

When I select an area with the "Crop"-Function, other coordinates are stored in the menu bar than the real start coordinates of the marker  (see Picture attached)
I use this manual Crop to detect the Coordinates for a later Batch-Processing, but the Crop-Area doesn't match.

Is this intentional or can it be adjusted so that the correct marked start coordinates?

![NoMac_Bugreport_Coordinates](https://user-images.githubusercontent.com/132557010/236157934-a894c2a2-800e-4be9-b1e3-9bf2247c59ca.png)

with best regards

Robert

novomesk:
I don't remember. I only know that crop was modified several times during last years.

leejuyuu:
Got it. This inconsistency feels unintended, so I will mark this as a bug and investigate for now.

haferburg:
@leejuyuu I attempted to fix it myself. Feel free to take a look.

PR #1070, opened 2024-05-03
Fixes wrong coordinates in crop toolbar (#944)

haferburg:
Bug: With Crop active, after creating/changing the rectangle by dragging, the coordinates displayed in the crop toolbar.

Fixed by: Set top left to top left.


haferburg:
I'm only able to verify that this fixes the bug. I have no idea if this change breaks anything else.

But when you look at the changes, setting top left to top left seems like the right thing to do.

leejuyuu:
Thanks! I've tested this patch, the crop tool still works as normal as I know. 
This will fix #944.

leejuyuu:
`DkEditableRect` is only used by `DkCropWidget`, so I think it should not break other functionalities.

PR #1082, opened 2024-05-31
Fix broken zip thumbnails

scrubbbbs:
This fixes the bug where zip thumbnails do not load at all. This is due to failing the test DkUtils::hasValidSuffix(). With globs, QRE uses file path matching for the "*" part of the glob -- "the transformation is targeting file path globbing, which means in particular that path separators receive special treatment". Since the zip paths are  encoded to something bizarre, QRE used this way won't accept them.

Well that's the backstory, but I think it should be rewritten so I'm submitting a simplified version that should do the job. I believe it will because the globbing patterns are normalized to be "*.something", always lowercase and nothing else. Now if they had other globbing foo in them this patch would not work for those cases.

I also reviewed all usages of hasValidSuffix() and found one spot that could be refactored.


scrubbbbs:
@novomesk thank you. I think I will proceed with the next phase of zip file fixes/missing features.

Issue #1068, opened 2024-05-02
Status bar displays wrong zoom value instead of 100%


tkzv:
**Describe the bug**
When I change image zoom (with up/down arrows or mouse wheel), when it becomes 100%, the status bar displays the previous value. Though, when I set 100% zoom with Ctrl-1, status bar correctly displays 100%; the gray panel at lower left also displays the correct value.

**To Reproduce**
1. Open an image.
2. Make sure status bar is visible.
3. Press Ctrl-1 to reset zoom to 100%.
4. Press the up arrow — the image is displayed at 150%, the panel at lower left and the status bar correctly indicate 150% zoom.
5. Press the down arrow — the image is displayed at 100%, the panel at lower left shows correct 100%, but the status bar still says 150%.
6. Press the down arrow — the image is displayed at 50%, the panel at lower left and the status bar correctly indicate 50% zoom.
7. Press the up arrow 2 times — zoom changes to 75%, then to 100%, the panel at lower left correctly shows those values, but the status bar says 75% for both.
8. Change zoom with the mouse wheel — same as above: correct values in status bar for all but 100%.

**Expected behavior**
The status bar should always display the correct magnification value.

**Desktop:**
 - OS: Gentoo Linux
 - Version 3.16.224-r6

**Additional context**
There were no errors in console.

leejuyuu:
Thanks for reporting, reproduced with nomacs 3.17.2295 (94e9adaa) on Arch Linux.

PR #1084, opened 2024-06-01
Fix status bar displaying wrong zoom

leejuyuu:
Fixes https://github.com/nomacs/nomacs/issues/1068.

There are two branches in `DkViewPort::zoom()` that changes the zoom but does not call `DkStatusBarManager::setMessage()`, so the status bar is not updated in some cases.

Add the call in the common function to fix this.

### DIFF
--- a/ImageLounge/src/DkGui/DkViewPort.cpp
+++ b/ImageLounge/src/DkGui/DkViewPort.cpp
@@ -509,6 +509,7 @@ void DkViewPort::resetView()
     controlImagePosition();
 
     emit zoomSignal(mWorldMatrix.m11() * mImgMatrix.m11() * 100);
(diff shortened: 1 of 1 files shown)


scrubbbbs:
Can confirm this is correct.  `zoom()` sometimes calls into `resetView()` and returns early, in which case no status message is set. 

There are two other spots where zoom signal is sent but no status message; are they also a problem?

Other than that,  I would refactor the redundant code but the current code is full of stuff like that.

leejuyuu:
> There are two other spots where zoom signal is sent but no status message; are they also a problem?

@scrubbbbs Could you point out where they are?

scrubbbbs:
Search for `emit zoomSignal`

leejuyuu:
The `emit zoomSignal` calls in `DkViewPort` all reach a `setMessage` currently, so I think it's fine. As for refactoring, I currently don't have enough familiarity to the codebase to ensure that I do not cause more trouble than solving problems, so let's fix the issue first.

PR #1087, opened 2024-06-05
Fix use-after-free in DkMetaDataDock (Debian #1051579)

scrubbbbs:
Fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1051579

When metadata dock is enabled, nomacs crashes some time after all tabs are closed. This is due to to use-after-free on the item model.

It seems like the intent of the incorrect code was to prevent a null image from causing an update to the view, so I have corrected as such. If the intent was to empty/reset the view that could be addressed in a separate PR.


novomesk:
I see that AppVeyor build failed. Try to add into appveyor.yml

`GIT_CLONE_PROTECTION_ACTIVE: false` after `environment:`

scrubbbbs:
The appveyor fix seems to work, when that is applied would you like me to rebase to restart appveyor?

novomesk:
yes, rebase.

Issue #1093, opened 2024-06-07
Overloaded signals not found when using Qt6


leejuyuu:
This issue is for tracking multiple issues that related to missing signals when using Qt6.

### Problem

In Qt5, some signals were overloaded, and these overloads were removed in Qt6.
For example, in Qt5 `QButtonGroup::buttonClicked` could be either `QButtonGroup::buttonClicked(QAbstractButton *button)` or `QButtonGroup::buttonClicked(int id)`, but in Qt6 there are  `QButtonGroup::buttonClicked(QAbstractButton *button)` and `QButtonGroup::idClicked(int id)` instead. The function signature changes caused some slots expecting `buttonClicked(int id)` to fail to connect.

Nomacs uses `QMetaObject::connectSlotsByName()` to simplify many signal slot connections, however, this method catches invalid connections and only warns in runtime. Therefore the connection failures were not noticed until people started seeing regressions. 

### Issues caused by this

- #1062
- #1076

### Proposed changes

I am working on removing all the `connectSlotsByName()` and start reconnecting the signal and slots explicitly. This way, any signature or name mismatch will be captured at compile time, and I will be able to fix them easily.

Issue #1076, opened 2024-05-17
Nomacs doesn't crop an image at a specified aspect ratio


BaumGuard:
**Describe the bug**
When trying to crop an image at a certain aspect ratio, the ratio chosen isn't being applied to the selection area. This means that one can freely select an area without the selected ratio. As a workaround one can enter the ratio in the text fields next to the drop-down menu.

**To Reproduce**
1. Open an image in nomacs
2. Change into cropping mode (either by right click -> Edit -> Crop Image or by clicking the corresponding symbol in the bar on the top)
3. Select an aspect ratio (e.g. 16:9)
4. Select an area in the image

**Expected behavior**
Normally the area selected should have the specified ratio, but the ratio isn't applied to the selection and to the text fields next to the drop-down menu

**Desktop (please complete the following information):**
 - OS: Manjaro Linux
 - nomacs Version: 3.17.2295

leejuyuu:
Thanks for reporting, confirmed on Arch Linux, nomacs 3.17.2295 (94e9adaa).

pbentkowski:
I have the same problem. I compiled from the git repo on 6 Jun 2024 with **no** plugins.

- OS: Tuxedo OS 3 
- nomacs Version: 3.17.2297 (cc3bf6d0)
- OpenCV: 4.5.4
- Qt:  6.7.0

pbentkowski:
A follow-up… I compiled Nomacs the same day, also on Tuxedo OS 3, but on a computer where the OS is installed from scratch. And **cropping works**! The difference seems to be the version of the Qt library.

- OS: Tuxedo OS 3
- nomacs Version: 3.17.2297 (bac800e5)
- OpenCV: 4.5.4
- Qt: 5.15.13

Probably, because of other project, the first computer is set to preffer Qt6.

leejuyuu:
@pbentkowski Thanks for testing this! I confirm that this is caused by changing from Qt5 to Qt6. This is caused by a missing signal connection.

```
[WARNING] QMetaObject::connectSlotsByName: No matching signal for on_ratioBox_currentIndexChanged(QString)
```

I am working on looking through all the auto signal connections, hopefully this can be fixed soon!

scrubbbbs:
@leejuyuu are you removing them? I think you should as they break when renaming the widget. Also could replace them with the new signal/slot syntax which catches mistakes at compile time.

leejuyuu:
Yes, I just happened to mentioned in #1093 :)

[...]

Issue #1062, opened 2024-04-12
nomacs could not change theme through Setting Tab


LloydLore:
I am a new user for nomacs v3.172295. The default theme is "Dark-theme", and I'd like to use "Light-theme". 
Issue Description: Could not change theme through menu/settings tab, 
Reproduce Step: 
1. Edit --> Settings, (Ctrl+Shift+P, could enter setting tab as well)
2. In settings tab, select General tab, and choose the option "Color Settings" from "Dark Theme" to "Light Theme"
3. The bottom line of this tab will prompt "Please Restart nomacs to apply changes"
4. Click the prompt, the application restarts, and the theme doesn't change
5. Also, the option "Color Setting"'s value doesn't change.

Walkaround:
I could manually change the settings resides in path "$HOME/.config/nomacs/'Image Louge.conf', the config field is "themeName312", change this value will result in application theme change. 

BTW, this application is awesome, thank you!

Aqa-Ib:
Can reproduce, but in my case the default theme is light and I can't change it to neither dark nor system.
Thanks!

leejuyuu:
It seems that this bug is intermittent. Some times the theme and config file is written correctly.

leejuyuu:
@LloydLore @Aqa-Ib I had a harder time reproducing this on 3.17.2296 (515812d) when trying to fix this. Could you try newer version as well?

A possible cause is that `QSettings` does not write to the file before the program closes. Maybe we can call `sync()` manually to try to make file writes earlier.

LloydLore:
@leejuyuu , I am using Arco Linux while it seems the package manager didn't update to the newest version. I will try it once the package is updated.

Aqa-Ib:
I have tried with latest git (installed through AUR package nomacs-git) and still the same here.
Probably these warnings are relevant:
```
[WARNING] QMetaObject::connectSlotsByName: No matching signal for on_themeBox_currentIndexChanged(QString)
[WARNING] QMetaObject::connectSlotsByName: No matching signal for on_keepZoom_buttonClicked(int)
[WARNING] QMetaObject::connectSlotsByName: No matching signal for on_loadGroup_buttonClicked(int)
[WARNING] QMetaObject::connectSlotsByName: No matching signal for on_saveGroup_buttonClicked(int)
[WARNING] QMetaObject::connectSlotsByName: No matching signal for on_openDefault_clicked()
[...]

PR #1096, opened 2024-06-09
Remove QMetaObject::connectSlotsByName()

leejuyuu:
Issues have been found when migrating from Qt5 to Qt6 due to overloaded signals were separated into different names. These were unnoticed because nomacs use `QMetaObject::connectSlotsByName()` to simplify many signal slot connections, which only warns in runtime. See https://github.com/nomacs/nomacs/issues/1093

 Remove all the `QMetaObject::connectSlotsByName()` calls and manually    connect the signals to make things explicit and allow compile time check    of the signatures, so similar problems can be catched early in the    future. The slots are renamed so that they do not follow Qt connectSlotsB    rules anymore. Any signature changes in the signals from Qt5 to Qt6 were    fixed to pass compiler.

Closes https://github.com/nomacs/nomacs/issues/1093
Fixes https://github.com/nomacs/nomacs/issues/1062
Fixes https://github.com/nomacs/nomacs/issues/1076

### DIFF
--- a/ImageLounge/src/DkCore/DkBasicWidgets.cpp
+++ b/ImageLounge/src/DkCore/DkBasicWidgets.cpp
@@ -358,8 +358,8 @@ void DkColorChooser::init()
     mAccepted = false;
 
     colorDialog = new QColorDialog(this);
-    colorDialog->setObjectName("colorDialog");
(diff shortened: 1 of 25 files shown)


scrubbbbs:
Looks good. I compiled and tested a few modules. I can't test everything, but reading the patch all the changes seem straightforward.

Issue #853, opened 2022-05-02
Building for Windows with MinGW


eddiezato:
Currently building for Windows is set up in cmake files for MSVC only. It would be good to have it for MinGW also.

I have tried several times to build `nomacs` in `MSYS2/CLANG64` environment, but I don't seem to have enough knowledge to solve all the problems that arise.

ChiaYen-Kan:
Hi:

i create a patch file for build with msys in x86_64 mode

not perfect but it can work, may be someone can make it to more perfect 

patch file:
[0001-fix-for-build-with-msys2.patch](https://github.com/nomacs/nomacs/files/11684606/0001-fix-for-build-with-msys2.patch)

build log:
[nomacs-build.log](https://github.com/nomacs/nomacs/files/11684574/nomacs-build.log)

eddiezato:
@ChiaYen-Kan thanks!
I was finally able to build.

eddiezato:
@ChiaYen-Kan Hey, sorry to bother, can you update your patch file? Want to try to build from the current master with Qt6.

ChiaYen-Kan:
@eddiezato
It seems impossible at the moment
In this project is depend some qt5 function, that's not exist in qt6, and no solution in qt6
and the original patch is base Unix/Linux config, but Unix/Linux config seem not yet full move to qt6(also include plugins project)

https://forum.qt.io/topic/131934/is-there-an-alternative-to-qwintaskbarbutton-in-qt-6/2
https://doc.qt.io/qt-6/extras-changes-qt6.html

ChiaYen-Kan:
@eddiezato 
msys2 + Qt6, drop the Qt5 feature

patch file, see the build log to patch all component
[0-main-0001-fix-build-with-msys2-Qt6.patch](https://github.com/nomacs/nomacs/files/14447706/0-main-0001-fix-build-with-msys2-Qt6.patch)
[1-3rd-party-imageformats-0001-fix-for-qt6.patch](https://github.com/nomacs/nomacs/files/14447705/1-3rd-party-imageformats-0001-fix-for-qt6.patch)
[2-plugins-0001-fix-for-qt6.patch](https://github.com/nomacs/nomacs/files/14447707/2-plugins-0001-fix-for-qt6.patch)

log
[build.log](https://github.com/nomacs/nomacs/files/14447700/build.log)

no test, and you want to build with msys2, please push the nomacs team support it

eddiezato:
[...]

PR #1097, opened 2024-06-09
Add Cmake cross-compilation target for Mingw/Windows

scrubbbbs:
Adds necessary cmake files to compile the Windows version on a Linux/ Unix host. Also includes a binary packaging script to make a Windows portable package.

This is intended for MXE as the cross compiling environment, since its the easiest method. However, there isn't much specific to MXE so it could support other methods, or maybe native Mingw on Windows.

The build recipe is roughly  (untested; based  on  my bash history):

```shell
apt install wine

cd ~/
git clone <nomacs url>
git clone <mxe url>
cd mxe

# note: mxe may have apt packages to sidestep compiling things
# note: this is for qt5
make MXE_TARGETS=`x86_64-w64-mingw32.shared` qtbase qtimageformats qtwinextras opencv quazip tiff exiv2 libraw

export MXE_DIR=~/mxe
export MXE_TARGET=x86_64-w64-mingw32.shared
export PATH="${MXE_DIR}/usr/bin:$PATH"

alias pkg-config=${MXE_TARGET}-pkg-config
alias qmake=${MXE_TARGET}-qmake
alias cmake=${MXE_TARGET}-cmake

cd ~/
mkdir build-nomacs
cd build-nomacs

cmake -D QT_VERSION_MAJOR=5 -D ENABLE_TRANSLATIONS=ON -D ENABLE_HEIF=ON -D ENABLE_AVIF=ON  \
-DENABLE_JXL=ON -D USE_SYSTEM_QUAZIP=OFF -D CMAKE_BUILD_TYPE=Debug ~/nomacs/ImageLounge

make -j8
make collect # build a runnable package

wine nomacs-mingw/nomacs.exe # run nomacs under Wine (actually works OK, but slower)
```


novomesk:
Can try to investigate/check how this impact those who would like to build nomacs under MSYS2 ?

If we can support the MSYS2, that would be great.

Take a look at https://github.com/nomacs/nomacs/issues/853 , there is some patch there.

scrubbbbs:
Moved some stuff around and added Msys2 support for Qt  5, easy to make a running build but will need some tweaks for packaging; mainly another strategy for pulling all the dependent dlls.

I also fixed a problem with configuring the Unix/Linux build with Quazip-1 (was missing WITH_QUAZIP define). And I replaced tabs with spaces in that file as well (very annoying!).

Note that Qt 6 is not working on either target (mxe or msys2/mingw), (not a priority right now?) but should be straightforward to add.

The recipe for Msys2/mingw
```shell
export target=mingw-w64-x86_64

pacman -S $target-qt5-base $target-qt5-svg $target-qt5-winextras $target-qt5-tools \
          $target-libraw $target-libtiff $target-exiv2 $target-opencv $target-cmake $target-gcc

export PATH=/ming64:$PATH

cd <build-dir>
cmake  -D  QT_MAJOR_VERSION=5  -D ENABLE_TRANSLATIONS=ON -D  USE_SYSTEM_QUAZIP=ON <nomacs-dir>/ImageLounge

cmake --build . --parallel 8
```

Issue #464, opened 2020-04-10
[Regression] Nomacs no longer respects KDE/Qt font DPI


geckolinux:
Hi there, I'm using Nomacs 3.14 on openSUSE Tumbleweed (rolling) with Qt 5.14.1.

I've used it for years, but for some reason it has recently stopped respecting my KDE font DPI setting of 108. Sorry that I didn't really notice when this happened, but I know that it used to work, probably within the past month or two something has changed. Here's a comparison of its UI with Kate, which behaves normally:

![Screenshot_20200410_091947](https://user-images.githubusercontent.com/2692138/78997615-876b4680-7b0c-11ea-8874-d14fbef0f2e0.png)

I can work around this with `QT_SCALE_FACTOR=1.2 nomacs`, but this never used to be necessary.

Thanks a lot!

diemmarkus:
thanks for reporting. this could be related to #450 (though I don't know what causes these issues)

geckolinux:
@diemmarkus Thanks for the reply. Also possibly related to #385 ?

mesaprotector:
Hope it's not a breach of etiquette to mention a four year old issue, but I just ran across this today. While I can run nomacs with the relevant environment variable to get the fonts to look correct, this causes zoom percentages to be wrong (ex. a full-screen screenshot will only actually fill the screen at 107%). 

Not sure what causes this either, but I can try different settings if anyone would like. I'm not much of a programmer.

leejuyuu:
Could not reproduce the font DPI with Kubuntu 22.04, so I cannot verify what #1102 is fixing. Forced font DPI to 144, the display scaling is 100%.

Top-left: 3.17.2297 (06671c64)
Top-right: 3.19.0 (d768ccda)
Bottom: 3.12 (from ubuntu repo)

![image](https://github.com/nomacs/nomacs/assets/46868116/edec94ff-2827-4411-b3dd-82007526c263)

leejuyuu:
About the image scaling, is it correct that image at 100% in nomacs on system display scaled to 150% should look the same when 150% in nomacs on system that displays 100%?

scrubbbbs:
The correct way to rescale nomacs UI is to use QT_FONT_DPI=96*scale_factor.

QT_SCALE_FACTOR will have issues as it (by-design) multiplies any system scale factor and nomacs' internal scale factor based on DPI.

scrubbbbs:
[...]

PR #1102, opened 2024-06-18
Enable HiDPI support in Qt and fixup the main view

scrubbbbs:
We want to enable HiDPI in Qt so that the UI is properly scaled for different display resolutions. The following changes are a first step in that direction:

- Do not disable Qt's internal HiDPI scaling; we should be relying on it for everything besides image/thumbnail display.
- Main view adjusts for scaling so 100% is actual pixels
- QT_SCALE_FACTOR works correctly; 100% is still actual pixels
- Dialogs using main view logic also adjusted

Remaining issues:
- old device-scaling/dpi logic can be refactored or removed
- Icons and thumbnail resolution are incorrect
- Plugins must be patched as well (patch forthcoming)

### DIFF
--- a/ImageLounge/src/DkCore/DkBaseViewPort.cpp
+++ b/ImageLounge/src/DkCore/DkBaseViewPort.cpp
@@ -302,7 +302,11 @@ QSize DkBaseViewPort::getImageSize() const
         return mSvg->defaultSize().scaled(size(), Qt::KeepAspectRatio);
     }
 
-    return mImgStorage.size();
+    // HiDPI: Pretend the image is smaller and keep scaling/translating logic the same
+    // At 100%, this gives mImageRect.size()*dpr == img.rect().size(), so Qt disables device scaling
+    qreal deviceScale = devicePixelRatioF();
+
+    return mImgStorage.size() / deviceScale;
 }
 
 QRectF DkBaseViewPort::getImageViewRect() const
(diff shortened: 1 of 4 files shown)


scrubbbbs:
Fixes https://github.com/nomacs/nomacs/issues/464

leejuyuu:
The icons looks pixelated indeed. This is scaled at 150%. Kubuntu 22.04, Qt 5.15.3.

Left: 3.17.2297 (06671c64)
Right: 3.19.0 (d768ccda)

image

Is the "Main view adjusts for scaling so 100% is actual pixels" taking effect, I feel both sides look the same.

scrubbbbs:
@leejuyuu Thank you for reviewing.  

I have been testing by setting QT_SCALE_FACTOR environment variable, which AFAIK is effectively what you get on a HiDPI display. In actuality my display is around 110 DPI which makes everything a bit too small for my taste.

Here are some screenshots before and after how it should look at 2x scaling. I am using a 1x1 pixel test pattern so the slightest amount of scaling will show up as a Moire pattern.

_I am now noticing the image scaling works as I intended but not the UI elements which seem to be over-scaled, so I've marked this PR as draft until I can git this resolved._

master 6d39242 1x
image

master 6d39242 2x
image

hidpi #1102 2x
image

leejuyuu:
Am I understanding correctly that you want QT_SCALE_FACTOR to scale only the UI elements but not the image? Is this the usual behavior of image viewers? It's more intuitive to me that everything scales together, I am no expert on these things though.

scrubbbbs:
@leejuyuu yes, that is correct. My feeling is that this is expected behavior, though I haven't researched other viewers in this regard. Also, it should make anti-aliasing look better as there isn't any additional scaling to mess it up. In addition it is a nice feature to be able to view the image with no transformations applied, if for no other reason than to verify the image quality or that artifacts were not introduced by the viewer or any plugins/editing operations.

scrubbbbs:
After further research I have determined this is not the correct solution, so I'm closing it.

QT_FONT_DPI=xx is the correct environment variable to change the UI scale which has all the desired effects (100% scale is physical DPI, etc).

PR #1114, opened 2024-06-29
Remove text based connections

leejuyuu:
Remove text based signal connection macros SIGNAL and SLOT in favor of function pointer based syntax. This enables compile time check of the signal and slot name and signatures, and makes the methods searchable via LSP.


novomesk:
Please rebase so we test together with my latest commits.

leejuyuu:
This has been rebased.

Issue #1116, opened 2024-06-30
CI libquazip for Qt6


leejuyuu:
Ubuntu 22.04 currently does not have libquazip for Qt6. We wait until the github action 24.04 image is out of beta, and then we can use libquazip1-qt6-dev.

PR #1117, opened 2024-06-30
Remove libquazip5-dev for Qt6 build

leejuyuu:
libquazip5-dev is for Qt5. Ubuntu 22.04 does not have libquazip for Qt6, so we need to wait for 24.04 to enable it.

Tracking in https://github.com/nomacs/nomacs/issues/1116.


leejuyuu:
The check step failed should not matter, this commit did not touch cpp sources.

PR #1121, opened 2024-07-09
Give default focus to central widget

### LINKED ISSUE #629: Keyboard shortcut issues in frameless mode
Hi! 👋 I love to use frameless mode. But I have small issue with it. Every time I open an image, navigation with keyboard arrows does not work until I click on the image with a mouse. And I have to press Escape 3 times to close image preview, on first click nothing happens, on second click dimmed background is being closed and only then image itself closes. 

_This does not present in regular window, everything works as expected._

### LINKED ISSUE #843: nomacs focus when opening
I'm using nomacs 3.16.1709 x64 on Windows and when opening a picture, I can't use the arrow keys to browse the other pictures without first clicking anywhere inside "nomacs". 

Would it be possible to give nomacs the focus when opening a picture?

### LINKED ISSUE #908: Nomacs is not responsive to keystrokes until an image has been selected.
**Describe the bug**
When first opening Nomacs and selecting a folder, keystrokes do nothing until the mouse has clicked somewhere in the main image panel.

**To Reproduce**
1. Open Nomacs.  It shows a list frequently used folders
2. Click on one of the folders.  Nomacs shows the image as expected.
3. Press any keyboard shortcut (right, left, up, down, 1-5, etc)
4. Nothing happens.
5. Click anywhere in the main image panel.
6. Keystrokes now work as expected.

**Expected behavior**
Nomacs should automatically initialize the main image panel to receive keystrokes without the user clicking on it first.

**Screenshots**
None

 - OS: Artix
 - Version git

**Additional context**
Alt-Tab away and back again, and the file explorer panel begins to receive keystrokes.

scrubbbbs:
On startup, gives focus to the central widget viewport, if there is one. This prevents any panel/dock from taking keyboard focus.

Other calls to setFocus/setFocusPolicy that were working around this are removed.

Fixes #843
Fixes #908
Fixes #629

### DIFF
--- a/ImageLounge/src/main.cpp
+++ b/ImageLounge/src/main.cpp
@@ -67,6 +67,7 @@
 #include "DkSettings.h"
 #include "DkTimer.h"
 #include "DkUtils.h"
+#include "DkViewPort.h"
 
 #include "DkDependencyResolver.h"
 #include "DkMetaData.h"
@@ -319,6 +320,9 @@ int main(int argc, char *argv[])
     if (parser.isSet(slideshowOpt)) {
         cw->startSlideshow();
     }
+
+    if (cw->hasViewPort())
+        cw->getViewPort()->setFocus(Qt::TabFocusReason);
 
 #ifdef Q_WS_MAC
     nmc::DkNomacsOSXEventFilter *osxEventFilter = new nmc::DkNomacsOSXEventFilter();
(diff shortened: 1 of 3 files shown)


ImageLounge/src/main.cpp
```diff
@@ -320,6 +320,9 @@ int main(int argc, char *argv[])
         cw->startSlideshow();
     }
 
+    if (cw->hasViewPort())
+        ((QWidget *)cw->getViewPort())->setFocus(Qt::TabFocusReason);
```
novomesk, line 324: You can use here

`cw->getViewPort()->setFocus(Qt::TabFocusReason);`

when you add

`#include "DkViewPort.h"`

scrubbbbs: My thought at the time was to avoid the dependency as it isn't needed. The worst case is the cast becomes invalid due to refactoring and crashes the app at startup; not in a place you'd have to uncover. But if you prefer the other way, no problem.

Issue #550, opened 2020-08-24
Auto Full Screen & Esc to Close Program (Not from GUI view from Fullscreen View)


Bonethugs1980:
This is not a major problem, just a couple of minor 'quality of life' adjustments.  Two things of note:

1. When you press Esc button (from the settings 'Close on Esc'), it goes back to the GUI view - which is fine.  There is no option to close the program entirely from Esc...

2. There is no option to open and go straight into fullscreen mode either from clicking/opening a file directly or from opening the file from within the program 'Open'.  It always goes to GUI view first before pressing F11 or selecting 'View'>'Fullscreen' or double click on the image (dependent on setting).

*If possible for upcoming release(s) (not important right now), could an extra option be 'Esc closes program' (when selecting the this potential option it will untick 'Close on Esc' or 'Exit program in fullscreen with Esc'

*If possible have the program auto open to fullscreen from the start of opening a new applicable/supported image file.

If this is already in the roadmap or of things to do, please ignore and I can close if already planned.

shankarsivarajan:
On version 3.17.2206, "Close on Esc" closes the program completely.

Bonethugs1980:
Thanks for letting me know, Shall I just close this and mention the full screen from start issue.  Or leave this open and update the heading and comments as such?

shankarsivarajan:
I don't know the etiquette around this either, but I'd leave it open and edit it. For future reference, I think it's good practice to make each bug/feature request a separate issue so they can be closed independently.

Bonethugs1980:
Just installed the beta version you mentioned 3.17.2206.  This does 'Esc to Close' the same way.  Is there a misunderstanding?

So, similar to FastStone Image Viewer (not sure if I can mentioned here).  It does Esc to Close from fullscreen, where as nomacs does it back to the GUI view, which is one step extra instead of just closing it outright from fullscreen view.

shankarsivarajan:
Ah, I see what you mean. I didn't understand the term "UI View"; I thought you meant that Esc closes the image but not the program, like Ctrl+W does. My bad.

diemmarkus:
> 1. When you press Esc button (from the settings 'Close on Esc'), it goes back to the GUI view - which is fine.  There is no option to close the program entirely from Esc...

true, it only closes nomacs if not in fullscreen - that's actually a bug.

[...]

PR #1122, opened 2024-07-11
Startup: fix non-maximized window at startup on GNOME

scrubbbbs:
On some platforms (GNOME/X11) and potentially other X11 window managers, the maximized state of the main window will not be restored.

Since Qt emulates showMaximized() on X11, it is likely this is not limited to GNOME and could affect other WMs now and in the future.

This corrects the problem on GNOME and may also prevent other WMs from breaking in the future.

I have also added a more complete "ensure the window is visible/active" check as sendPostedEvents() is not quite enough to fully realize the window.

---

I have tested the patch on Windows as well and it has no effect as expected.  Also GNOME/Wayland doesn't have this problem, nor the various fullscreen problems (e.g. #550) , which this does not address.

Qt documentation regarding X11 WM issues:
https://doc.qt.io/qt-5/application-windows.html#x11-peculiarities

### DIFF
--- a/ImageLounge/src/main.cpp
+++ b/ImageLounge/src/main.cpp
@@ -265,11 +265,20 @@ int main(int argc, char *argv[])
     } else
         w = new nmc::DkNoMacsIpl();
 
+    qInfo() << "maximized:" << w->isMaximized() << "fullscreen:" << w->isFullScreen();
+
     // show what we got...
     w->show();
 
-    // this triggers a first show
-    QCoreApplication::sendPostedEvents();
+    bool maximized = w->isMaximized();
+
+    while (!w->isActiveWindow())
+        qApp->processEvents();
+
+    // Qt emulates showMaximized() on some platforms (X11), so it might not work.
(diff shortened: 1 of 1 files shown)


SR-G:
I'm running ARCHLINUX + GNOME SHELL
I was using the "AUR" flavor and since a few weeks it was insanely allowing to use (nomacs was never starting maximized / never remembering the last window status at startup).

Today i removed the AUR flavor and installed the package from source - it's a bit better, however : 
- in a multi-monitor setup, the last monitor being used is never remembered (nomacs is always opening on monitor #1 - this is a consistent behavior = last monitor being used is not remembered anymore (it was in the AUR version)) 
- the "last window status being remembered" (including in my case just being "maximized") is most of the time remembered, but NOT always (i have no idea why - like i open then close 1, 2, 3, ... 5 images : all are maximied, then the 6th one from same same folder, without having done anything else before, is NOT maximized anymore, then it's remembered again) - but this is not consistent, most of the time, it's remembered fine (and it's already way better than before)

PR #1136, opened 2024-07-22
Frameless: fix bugs and improve mode switching

### LINKED ISSUE #765: 'Private Mode' inastance becomes standard instance
**Description**
'Private Mode' instance becomes standard instance after activating 'Pseudocolor Function'.

**Severity**
Minor

**System**
OS: Windows 10 Pro
Version 20H2

**Steps to reproduce the behavior**
1. Open image
2. Activate 'Pseudocolor Function' (Ctrl+U)

**Expected behavior**
'Private Mode' instance  is displayed with pseudocolor toolbar.

**Video**
https://user-images.githubusercontent.com/37305513/126499340-fade0af8-6d1e-4bc7-afdd-3d5b45d9dcd1.mp4

**Actual behavior**
Standard instance is displayed with pseudocolor toolbar.

### LINKED ISSUE #781: Incorrect exit from ‘Full screen’
**Steps To Reproduce**  
1.	Open nomacs app;
2.	Open image;
3.	Right-click image -> Activate Frameless;
4.	Right-click image -> Activate Full screen;
5.	Right-click image -> Deactivate Full screen.

**Expected behavior**  
Menu bar is not displayed. 
  
**Actual behavior**  
Menu bar is displayed.  
  
**Screenshots**  
image4
   
**Environment**  
 - Windows 10 Home [x64] ver 21H1;  
 - nomacs ver 3.16.1709 [x64].

### LINKED ISSUE #1032: Start with full screen mode
Hello, I need it to open in full screen mode when I launch Nomacs. and when I press Esc, it closes the app (not to get out of full screen mode).

### LINKED ISSUE #1134: Frameless (Win11) obstructed by what appears like a Mask on WideScreen Monitor.
See Video.

I'm having a weird issue where I cannot drag a frameless window on my ultra wide monitor because it gets cut off almost like there are masks on the left/right side of my monitor over it. Disappears.

https://github.com/user-attachments/assets/efc28b76-b3f2-4857-83ea-93a7c79f059d

Win11X64, Latest Pre-Release build of Nomacs

scrubbbbs:
PR #1128 introduced a regression that broke frameless mode on Windows.  This patch resolves that issue and cleans up various aspects of mode switching.

See commit message for list of changes.

### DIFF
--- a/ImageLounge/src/DkCore/DkSettings.cpp
+++ b/ImageLounge/src/DkCore/DkSettings.cpp
@@ -374,6 +374,7 @@ void DkSettings::load(QSettings &settings, bool defaults)
     app_p.showRecentFiles = settings.value("showRecentFiles", app_p.showRecentFiles).toBool();
     app_p.useLogFile = settings.value("useLogFile", app_p.useLogFile).toBool();
     app_p.defaultJpgQuality = settings.value("defaultJpgQuality", app_p.defaultJpgQuality).toInt();
+    app_p.appMode = settings.value("appMode", app_p.appMode).toInt();
 
     QStringList tmpFileFilters = app_p.fileFilters;
     QStringList tmpContainerFilters = app_p.containerRawFilters.split(" ");
@@ -535,6 +536,7 @@ void DkSettings::save(bool force)
 {
     DefaultSettings s;
     save(s, force);
+    qDebug() << "saved settings to:" << s.fileName();
 }
 
 void DkSettings::save(QSettings &settings, bool force)
@@ -600,7 +602,7 @@ void DkSettings::save(QSettings &settings, bool force)
     // always save (user setting)
     settings.setValue("defaultJpgQuality", app_p.defaultJpgQuality);
     settings.setValue("appMode", app_p.appMode);
-    settings.setValue("currentAppMode", app_p.currentAppMode);
+    // settings.setValue("currentAppMode", app_p.currentAppMode);
 
     settings.endGroup();
     // Global Settings --------------------------------------------------------------------
@@ -802,8 +804,6 @@ void DkSettings::save(QSettings &settings, bool force)
     sync_d = sync_p;
     meta_d = meta_p;
     resources_d = resources_p;
-
-    qDebug() << "settings saved to" << settingsPath();
 }
 
 void DkSettings::setToDefaultSettings()
@@ -1078,8 +1078,9 @@ void DkSettingsManager::init()
 
     param().load(settings, true); // load defaults
 
-    int mode = settings.value("AppSettings/appMode", param().app().appMode).toInt();
-    param().app().currentAppMode = mode;
+    // TODO currentAppMode is obsolete, now set in main() after argument parsing
+    // int mode = settings.value("AppSettings/appMode", param().app().appMode).toInt();
+    // param().app().currentAppMode = mode;
 
     // init debug
     DkUtils::initializeDebug();
--- a/ImageLounge/src/DkGui/DkNoMacs.cpp
+++ b/ImageLounge/src/DkGui/DkNoMacs.cpp
@@ -136,8 +136,6 @@ DkNoMacs::DkNoMacs(QWidget *parent, Qt::WindowFlags flags)
     am.createMenus(mMenu);
     am.enableImageActions(false);
 
-    mSaveSettings = true;
-
     mOpenDialog = 0;
     mSaveDialog = 0;
     mThumbSaver = 0;
@@ -162,7 +160,7 @@ DkNoMacs::DkNoMacs(QWidget *parent, Qt::WindowFlags flags)
     mOldGeometry = geometry();
     mOverlaid = false;
 
-    resize(850, 504);
+    resize(986, 686); // default window size, fits the settings tab
     setMinimumSize(20, 20);
 
     //// fun fact
@@ -204,7 +202,7 @@ void DkNoMacs::init()
     readSettings();
     installEventFilter(this);
 
-    if (DkSettingsManager::param().app().appMode != DkSettings::mode_frameless) {
+    if (DkSettings::normalMode(DkSettingsManager::param().app().appMode) != DkSettings::mode_frameless) {
         DkToolBarManager::inst().showDefaultToolBar(DkSettingsManager::param().app().showToolBar);
         showMenuBar(DkSettingsManager::param().app().showMenuBar);
         DkStatusBarManager::instance().show(DkSettingsManager::param().app().showStatusBar);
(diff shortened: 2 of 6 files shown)


scrubbbbs:
Tested with Qt5 on Linux/Windows and Qt6 on Linux. Any help to test on Mac OS as well as Qt6/Windows is appreciated.

ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -578,44 +612,74 @@ void DkNoMacs::toggleFullScreen()
...
+    // -maximized->fullscreen doesn't cover the screen (Qt 6.5.1 / Windows 10)
+    // -maximized->fullscreen->maximized->normal loses normal size (Qt 5.15.13 / Linux / GNOME)
+    // #define USE_WINDOWSTATE 1
+
+#ifdef USE_WINDOWSTATE
     setWindowState(windowState() | Qt::WindowFullScreen);
+#else
+    mWasMaximized = isMaximized();
+    showFullScreen();
+#endif
```
leejuyuu, line 638: According to [the source code of `showFullScreen()`](https://github.com/qt/qtbase/blob/6c036012b5f2304a05af29f29daa7582603f79ae/src/widgets/kernel/qwidget.cpp#L3038-L3049), it would seem that `Qt::WindowMaximized` and `Qt::WindowFullScreen` does not mix. We should use the higher level `showFullScreen()` and remove the `setWindowState` branch. This should be safe because I do not see `USE_WINDOWSTATE` defined anywhere else.

ImageLounge/src/DkCore/DkSettings.cpp
```diff
@@ -600,7 +602,7 @@ void DkSettings::save(QSettings &settings, bool force)
     // always save (user setting)
     settings.setValue("defaultJpgQuality", app_p.defaultJpgQuality);
     settings.setValue("appMode", app_p.appMode);
-    settings.setValue("currentAppMode", app_p.currentAppMode);
+    // settings.setValue("currentAppMode", app_p.currentAppMode);
```
leejuyuu, line 605: Should we remove this as `currentAppMode` is replaced?

ImageLounge/src/DkCore/DkSettings.cpp
```diff
@@ -1078,8 +1078,9 @@ void DkSettingsManager::init()
 
     param().load(settings, true); // load defaults
 
-    int mode = settings.value("AppSettings/appMode", param().app().appMode).toInt();
-    param().app().currentAppMode = mode;
+    // TODO currentAppMode is obsolete, now set in main() after argument parsing
+    // int mode = settings.value("AppSettings/appMode", param().app().appMode).toInt();
+    // param().app().currentAppMode = mode;
```
leejuyuu, line 1083: Should we remove this as currentAppMode is replaced?

ImageLounge/src/main.cpp
```diff
@@ -234,44 +231,75 @@ int main(int argc, char *argv[])
...
             qWarning() << "illegal mode: " << pm << "use either <default>, <frameless> or <pseudocolor>";
+    }
+
+    if (parser.isSet(fullScreenOpt))
+        mode = nmc::DkSettings::fullscreenMode(mode);
 
-        nmc::DkSettingsManager::param().app().currentAppMode = mode;
+    if (!nmc::DkSettings::modeIsValid(mode)) {
+        qWarning() << "invalid mode:" << mode;
+        mode = 0;
```
leejuyuu, line 255: Suggest changing to 
```suggestion
        mode = nmc::DkSettings::mode_default;
```

leejuyuu:
There's an regression that I am not sure whether is introduced by this PR. Found on Win10, with the portable CI artifact of 50319582:

1. Open an image. The toolbar icons are shown.
2. Switch to frameless mode.
3. Switch back to normal mode. The toolbar is missing. It is shown in 3.17.2295.

leejuyuu:
Confirm this fixes #765, #781, and #1134. I think this should not close #1032 because there is currently no way (that I know of) to exit the app directly from fullscreen mode, so the mode can never be persisted.

ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -2074,16 +2153,21 @@ DkNoMacsFrameless::DkNoMacsFrameless(QWidget *parent, Qt::WindowFlags flags)
...
-
     chooseMonitor(false);
 #endif
-    show();
 
 #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     connect(mDesktop, &QDesktopWidget::workAreaResized, this, &DkNoMacsFrameless::chooseMonitor);
     connect(am.action(DkActionManager::menu_view_monitors), &QAction::triggered, this, &DkNoMacsFrameless::chooseMonitor);
+#endif
 #endif
```
leejuyuu, line 2171: This would disable this block completely because CHOOSE_MONITOR is not defined anywhere. Is it necessary and safe to do so in this PR?

ImageLounge/src/DkGui/DkNoMacs.cpp
scrubbbbs: Thanks for looking into this, I left it in for testing but there is no need.

ImageLounge/src/DkCore/DkSettings.cpp
scrubbbbs: currentAppMode is still needed in a few places (should be removable with some refactoring), I'm commenting out since it could be useful to know when refactoring. At that time it should be removed.

ImageLounge/src/DkCore/DkSettings.cpp
scrubbbbs: See my next comment

scrubbbbs:
> I think this should not close #1032 because there is currently no way (that I know of) to exit the app directly from fullscreen mode, so the mode can never be persisted.

It is possible with ESC in frameless mode or Alt-F4 in any mode. If there is a keybinding to File/Exit that also works. It is also possible for window manager to trigger it, independent of anything we do in nomacs.

I think ESC should continue to exit fullscreen mode as that is expected behavior for apps in general.

scrubbbbs:
> There's an regression that I am not sure whether is introduced by this PR. Found on Win10, with the portable CI artifact of 50319582:

Good catch. I'm also seeing it on the Linux build. Probably because mode switch saves settings at the time of app exit, which has the toolbar hidden. I'm working on a fix.

ImageLounge/src/DkGui/DkNoMacs.cpp
scrubbbbs: It is safe to do, but not necessary. I am not aware of the reasoning around this design choice, but I think it may be a good thing to deprecate:

- It's pretty atypical behavior for a viewer application
- The frameless window might show on a different monitor from the normal window; one might not remember or be confused as to why that is.
- If you clicked "Remember my choice" in choose monitor dialog, you might have trouble changing monitors; I can't seem to find the context menu anywhere.

leejuyuu:
> There's an regression that I am not sure whether is introduced by this PR. Found on Win10, with the portable CI artifact of 50319582:

Confirm this is now fixed.

leejuyuu:
Another possible regression is that switching between modes sometimes seems to reset settings. Found on Win10, with the portable CI artifact of 50343050


https://github.com/user-attachments/assets/89dda770-4fe7-405b-a078-7b0f85ca8cd4

ImageLounge/src/DkGui/DkNoMacs.cpp
leejuyuu: I agree with the reasoning. However, if this deprecation is unnecessary for this fix to work, maybe we should split this change into another PR. That PR could remove this whole block instead of hiding part of the code by a macro, and unused functions like `chooseMonitor` could also be cleaned up.

scrubbbbs:
> Another possible regression is that switching between modes sometimes seems to reset settings. Found on Win10, with the portable CI artifact of 50343050

The settings loading/saving seems straightforward in code. But the way nomacs is restarted, it is possible to have two instances open at the same time. Maybe that puts a race condition on the settings file which causes the new instance to see a partially written settings file. I think should be a new issue, nothing has changed regarding how nomacs restarts.

ImageLounge/src/DkGui/DkNoMacs.cpp
scrubbbbs: Agreed, I'll back out this change. I think that wraps things up unless you have anything else?

scrubbbbs: After further reflection I've ported chooseMonitor to Qt6 to avoid any functional bifurcation. The new implementation works on both Qt versions.

Issue #1118, opened 2024-07-04
Title bar doesn't follow dark theme in Windows 11


Solarflame5:
**Describe the bug**
Apps can set their title bar color to dark in Windows 11, nomacs doesn't support this and always shows a light title bar.

**To Reproduce**
1. Set nomacs theme to dark
2. Title bar stays light

**Expected behavior**
Title bar turns dark

**Screenshots**
![image](https://github.com/nomacs/nomacs/assets/47642904/4ab16b97-861f-4b56-85b7-06de4cc719a8)

**Desktop (please complete the following information):**
 - OS: Windows 11 Pro
 - Version: 23H2

scrubbbbs:
This is possible by using internal windows APIs. Example [here](https://github.com/scrubbbbs/cbird/blob/3c0f82e51e737dadef4db94a89374922d7361d8b/src/qtutil.cpp#L642).

leejuyuu:
>  The same issue existed in the Blender software, and it was fixed in release 3.3. I don’t have any information about the code or programming, but I am adding the relevant links here in case they are helpful. Thank you for your great app! ❤️
> 
> [Blender 3.3 Release Notes](https://developer.blender.org/docs/release_notes/3.3/user_interface/#:~:text=Windows%3A%20title%20Bar%20will%20respect%20operating%20system%20%22Dark%20Mode%22%20setting.)
> [Fixing Issue](https://projects.blender.org/blender/blender/commit/ddbac88c08ef)
> [Fixing Issue](https://archive.blender.org/developer/D14847)

_Originally posted by @mahdishalchian in https://github.com/nomacs/nomacs/issues/1140#issuecomment-2440756242_

scrubbbbs:
This will be fixed in #1140 however Qt 6.8.0 or later is required which isn't in the appveyor workflow yet.

Issue #885, opened 2022-09-04
Meta Data Info panel, glitch on text selection


hydropix:
When you want to select a part of a text from the metadata, the text is split with an offset. I think that the font has a different setting in the letter space, because the offset increases with the length of the text.
![image](https://user-images.githubusercontent.com/5154373/188307021-b19cdd28-9bc8-45a2-b7f2-30190888f1ae.png)

Issue #699, opened 2021-06-11
Console font color looks uncomfortable for eyes in dark mode


leafyboi:
**Describe the bug**
Console font color looks uncomfortable for eyes in dark mode.

**To Reproduce**
1. Click on 'Ctrl + Alt + L' for open console.
2. Click on 'Ctrl + Shift + P' for open settings.
3. Change the color theme in color setting on 'Dark Theme'.
Result: Console log color can't be changed. 

**Expected behavior**
Add color change setting of console please.)

**Desktop (please complete the following information):**
 - OS: [Windows 10 Pro, build 19042.985]
 - Version [3.16.1709 64-bit]

**Additional context**
It is not bug, but blue color looks so uncomfortable for eyes as for me and I could not find a setting where you can change the color of the console.

Issue #617, opened 2021-02-22
Windows behind nomacs showing through when viewing settings in frameless view


BillH99999:
When I view settings in frameless view it puts the tabs at the top of the screen and the settings slightly below (see area enclosed in red in screenshot).

![image1](https://user-images.githubusercontent.com/26846819/108775742-60481880-7516-11eb-83ae-81c592882ad9.jpg)

In between any open windows or the desktop shows through.  This isn't so bad when it is a single color desktop, but If I have other windows open behind nomacs, then that shows through and is very distracting.  I'm not sure this is a bug, but is there a way to prevent what is behind nomacs from showing through?  If not, can this be an enhancement to a future release?

Love the program.  Thanks!

bill

PR #1140, opened 2024-07-26
Theme: New Light and Dark Themes

scrubbbbs:
I ported the dark theme to light theme and made a lot of  improvements.  This fixes a few issues (see commit message). Update: dark theme now ported from new light theme.

Looking for testers and feedback on choices made here. There are only a few changes to source for minor fixes, so it can be tested by replacing the theme file in older versions.

Todo:
- ~checkbox svg for manipulator widget checkbox~
- ~Qt 6 port~
- ~Radio buttons do not follow theme colors~

Default light theme:
image

Frameless  mode:
image

Light theme with user background color:
image


scrubbbbs:
New default dark theme (based on new light theme)
image

With dark purple background
image

With dark red background. Maybe useful for astronomy.
image

novomesk:
Please rebase so that AppVeyor runs successfully.

leejuyuu:
At 02b2b28 #617 persists, on Arch with Sway WM, Qt 5.15
image

leejuyuu:
For #699, I feel that the blue font color is the cause that's unconfortable to read. Does the color encode meanings like log levels? Do you think it might be better to just print them in white?

image

leejuyuu:
I could not reproduce the glitch of #885, so not sure whether it is fixed or not. The new version still has some weirdness; sometimes the text blurs a little when selecting and deselecting. Not sure it's related though.

scrubbbbs:
> At [02b2b28](https://github.com/nomacs/nomacs/commit/02b2b287a6647e2b7273080fc1950907c830aae4) #617 persists, on Arch with Sway WM, Qt 5.15

This is really bizarre. My first thought is you tested it without `make install` which is needed to load the new theme (the theme won't load from the build directory if you run `./nomacs`. In the meantime I'll check if it is a Wayland issue.

scrubbbbs:
> not reproduce the glitch of #885, so not sure whether it is fixed or not

Nice catch. I checked it again and you are right, it is not fixed. The easy way to see it is double-click a wide text field to select all, then single-click on it. The glitch appears  in Both Qt5 and Qt6.

scrubbbbs:
> For #699, I feel that the blue font color is the cause that's unconfortable to read. Does the color encode meanings like log levels? Do you think it might be better to just print them in white?

The colors are for error levels, so warning is yellow, critical is red etc. The hard-coded colors don't work well, either too dark on the dark theme, or too light on the light theme. 

The black background is the quick fix. It is possible to add the log colors to the theme but a new method to pass colors around would need to be devised - the log text lines are rich text and that doesn't support stylesheets.

I think the colors are good as they help draw attention to issues that might matter (the yellow and red), so we should try to keep them. As for the black background, it is not ideal but the console is more of a debugging tool.

leejuyuu:
> you tested it without `make install` which is needed to load the new theme (the theme won't load from the build directory if you run `./nomacs`. 

That's the case. Can confirm #617 fixed now.

image

I feel the background and border color of the directory entries are a bit too dark.

ps_20240908125612

The left margins of the menu entries when there's icon are missing or too small.

ps_20240908125730

The Qt controls in the new theme look off.

image

Here's how they look originally.

mahdishalchian:
image
Hello
I'm not sure if I'm expressing this in the right place, but the top bar of the software is white while in dark mode. In most applications, the dark mode also changes the color of the top bar.

scrubbbbs:
>  In most applications, the dark mode also changes the color of the top bar.

The title bar is drawn by the window manager, not by nomacs. I'm not sure about Linux, but there is a private/undocumented API for Windows, might be supported now. See #1118

mahdishalchian:
The same issue existed in the Blender software, and it was fixed in release 3.3. I don’t have any information about the code or programming, but I am adding the relevant links here in case they are helpful. Thank you for your great app! ❤️

[Blender 3.3 Release Notes](https://developer.blender.org/docs/release_notes/3.3/user_interface/#:~:text=Windows%3A%20title%20Bar%20will%20respect%20operating%20system%20%22Dark%20Mode%22%20setting.)
[Fixing Issue](https://projects.blender.org/blender/blender/commit/ddbac88c08ef)
[Fixing Issue](https://archive.blender.org/developer/D14847)

scrubbbbs:
> It looks nice. The only thing I don't like so far is that I don't see menu border in light theme:

Thanks for reviewing. Fixed this in today's patch. I am looking at fonts right now, I think that can be improved a bit, mainly the default 10pt is too small by today's standards. Any opinions on this?

scrubbbbs:
> The Qt controls in the new theme look off.

I had to re-style some of the controls because they didn't follow the theme background color (they used the native qt widget theme). On some controls this is not possible without also replacing the pixmaps drawn on the control with our own versions, which isn't going to match the system controls at all.

I think this is the only way to get a consistent look between platforms, and the look I've been trying to emulate is the flat/unshaded/minimal look from the dark theme. At the same time I'm trying to minimize the change set and ensure the controls are scaled properly for HiDPI. We could try to emulate widgets of a well-known theme like Windows 11 or KDE Plasma, but that would obviously be more involved.

If there is a specific area I can improve I'm open to suggestions. I can already see the up/down arrows on spin boxes are not always nice (better at higher dpi), this is on my list.

scrubbbbs:
Today's patch:
- revisited radio button and checkbox to make them align vertically and closer to text baseline
- bump most font sizes by one point

scrubbbbs:
> The left margins of the menu entries when there's icon are missing or too small.

This appears to be a QT bug with stylesheets. You will notice it is also present with the old dark theme, and could change depending on QT version. Any attempt to modify the background color of QMenu will result in loss of margins/padding... and trying to fix/restore the padding comes with worse issues IMO.

Below is a canonical example I cooked up to investigate. You'll notice "Bug 2" which doesn't seem to have a solution, which feels like a bigger issue to me.

https://github.com/user-attachments/assets/31da4026-61bd-4d0f-a9c1-40555d96e558

scrubbbbs:
> I feel the background and border color of the directory entries are a bit too dark.

Thanks, I  fixed this today as well, I'm going to fix the spin boxes and after that call it good with known issues. I don't want to hold the patch too long as it is already too big IMO.

leejuyuu:
> I think this is the only way to get a consistent look between platforms, and the look I've been trying to emulate is the flat/unshaded/minimal look from the dark theme.

Is this really necessary? I was not aware of the "flat" design because I have been using the light theme. To be frank, I personally like the default Qt controls more.

> default 10pt is too small by today's standards

The 11pt font size looks fine. However, the icons seem relatively too large. Would it be possible to reduce their sizes slightly and give the buttons some padding?

image

> The left margins of the menu entries when there's icon are missing or too small.
> I feel the background and border color of the directory entries are a bit too dark.

These two are fixed for me. Thanks.

### Other things I noticed when trying the dark theme:

The chevron icon of the dropdowns is black for the dark theme.

image

One of the buttons on the batch processing button seems to get an extra border.

ps_20241117015331


When the style of the tab gets flattened, it becomes less obvious which is active.


ps_20241117015742

While using the defaults, the active tab looks slightly above others because of its border and shadows.

image

It's worse for the dark theme because the color is inverted. It feels like "e.jpg" gets highlighted with lighter color and a bottom border, but the active tab is settings.

image

scrubbbbs:
> > I think this is the only way to get a consistent look between platforms, and the look I've been trying to emulate is the flat/unshaded/minimal look from the dark theme.
> 
> Is this really necessary? I was not aware of the "flat" design because I have been using the light theme. To be frank, I personally like the default Qt controls more.

Besides the consistent look and feel, it is necessary to style all of the default controls so that custom colors can work reliably.

The reason is that setting the application palette (or any widget's palette) isn't predictable, since the style plugin (Qt widget rendering  part) is not required to use them, and often won't. So for example on Windows we end up with buttons and tabs that do not take the custom background color (or any shade of it).

In today's patch I've added this capability (setting the application palette). You can see the result on Windows is not ideal
image

While on Linux with the Fusion style plugin, not too bad since it doesn't render native widgets
image

I'm considering if we should also add an option to choose the style plugin, it's easy to support and gives the user more options. Although power users have always been able to do this by setting it in the environment (`QT_QPA_PLATFORMTHEME`).

leejuyuu:
> You can see the result on Windows is not ideal

I get your point now.

> add an option to choose the style plugin

Just curious, if this is possible, can we make the windows version default to a style plugin that plays well with the backgrounds, so there's no need to reset the widget styles?

scrubbbbs:
> > add an option to choose the style plugin
> 
> Just curious, if this is possible, can we make the windows version default to a style plugin that plays well with the backgrounds, so there's no need to reset the widget styles?

We could. Qt would have to be compiled with this style plugin enabled or we build it separately ourselves. My cross-compiled Windows build has the Fusion style and it looks identical to the Linux version.

The main issue, I'm finding it is pretty tricky to remap the system palette even for well-supported style plugins. While  small changes usually look OK, there is often a loss of contrast, and there is a point where it loses all contrast unless we invert the palette, which can change the 3d look of widgets (maybe Fusion is OK for this, not sure). The stylesheet widgets do not suffer from this issue since they use alpha  blending.

I don't feel the result will ever be as good as having a stylesheet theme which was designed for this purpose (recoloring). The level of control simply isn't there. However if it turns out to be good enough perhaps it can be the default.

scrubbbbs:
> When the style of the tab gets flattened, it becomes less obvious which is active.

I put a border around the active tab, which matches the bottom border of inactive tabs

image

image

leejuyuu:
> The main issue, I'm finding it is pretty tricky to remap the system palette even for well-supported style plugins.

Understood, let's proceed as you planned.

> I put a border around the active tab, which matches the bottom border of inactive tabs

Thanks, this makes it a lot clearer.

At one moment, all the icons became lighter. I could not reproduce it though. Also, does the widget theme only work with the system theme?

image

The dropdown item becomes white when using light theme and windows widget style.


https://github.com/user-attachments/assets/41efe786-f729-4603-ab5d-3012a67de566

scrubbbbs:
> At one moment, all the icons became lighter. I could not reproduce it though.

The icon color can't be changed without restarting. I haven't been into the icon coloring code yet, but I feel it needs to be thrown out and replaced with our own QIconEngine that supports altering SVGs. This would allow SVGs in the CSS to also be colorized properly.

However, there is a bug where two restarts are needed to get the icon color to show correctly. I'm not sure what all of the cases are to trigger the bug, but one is changing the platform theme from light to dark while nomacs is open.

> Also, does the widget theme only work with the system theme?

It is applied with all themes; however the dark/light nomacs theme redefine most of the widgets we are using. Any controls or properties we did not override in our .css files will fall back to the style plugin. For example the spacing around controls is not something I'm overriding right now.

> The dropdown item becomes white when using light theme and windows widget style.

Nice catch, this seems to be on any combo box that is focused, when de-focused it draws correctly.

scrubbbbs:
> The dropdown item becomes white when using light theme and windows widget style.

This appears to be a qt bug, I have [filed a report](https://bugreports.qt.io/browse/QTBUG-131761). I'm wondering if we should lock the dark/light theme to the fusion style for better consistency.

scrubbbbs:
Today's push fixes #1118 on Qt 6.8.0 and later, and also correctly responds to external theme changes. The "windowsvista" style plugin, however can no longer be colorized (it looked pretty bad so I think this is OK). The "windows11" plugin can be, provided the user is on windows 11 or later. I've tested on Windows 10,11, and MacOS and I think this is ready. I wait to hear for any final feedback, will aim to merge this weekend.

leejuyuu:
I noticed that this version shows the "Please restart" message when changing themes while f1cb0e does not. Is this intended?

image

scrubbbbs:
> I noticed that this version shows the "Please restart" message when changing themes while [f1cb0e5](https://github.com/nomacs/nomacs/commit/f1cb0e56faeb13603a936cf3179bbf9f11760c5d) does not. Is this intended?

Yes, it seems that dynamic switching can be unpredictable so I added this back as it can't hurt.

leejuyuu:
> > I noticed that this version shows the "Please restart" message when changing themes while [f1cb0e5](https://github.com/nomacs/nomacs/commit/f1cb0e56faeb13603a936cf3179bbf9f11760c5d) does not. Is this intended?
> 
> Yes, it seems that dynamic switching can be unpredictable so I added this back as it can't hurt.

It feels confusing. I thought "The theme did change, should I still restart the program?" when I saw that. Could you make the message only show when the dynamic switching fails?

scrubbbbs:
> > > I noticed that this version shows the "Please restart" message when changing themes while [f1cb0e5](https://github.com/nomacs/nomacs/commit/f1cb0e56faeb13603a936cf3179bbf9f11760c5d) does not. Is this intended?
> > 
> > Yes, it seems that dynamic switching can be unpredictable so I added this back as it can't hurt.
> 
> It feels confusing. I thought "The theme did change, should I still restart the program?" when I saw that. Could you make the message only show when the dynamic switching fails?

Well it doesn't fail in any consistent manner. Sometimes a control on a particular plugin on a particular qt version won't paint correctly when playing with the colors and switching between themes/plugins etc and restart fixes that up.

Besides that, I did forget that the icon color will not be applied until restarting. So dynamic switching for the moment is giving a preview of what it all looks like, not the final result.

It's more like a "restart may be needed" not a "restart is required" situation. We could try to communicate that with a dialog? I am open to ideas.

leejuyuu:
> I did forget that the icon color will not be applied until restarting. So dynamic switching for the moment is giving a preview of what it all looks like, not the final result.

Maybe change the message to something like "Restart nomacs (recommended to fully apply the new theme)"?

scrubbbbs:
Today's push handles palette changes more correctly which fixed some bugs I noticed on Windows 10, most obvious of which is that I wasn't picking up the windows dark theme at startup when the System theme was selected or when switching between nomacs themes back to the System theme.

It also makes Fusion the default style plugin on windows.
I came to this decision since:
- on Windows 10, the default windowsvista plugin will not use the windows dark palette
- windowsvista plugin also cannot be recolored at all. A hack is needed to get any color changes, but even then the result is poor (e.g. button/menu colors remain gray).
- on >= Windows 11, the "windows11" plugin looks a bit off to me (menu bar spacing is quite bad IMO)

leejuyuu:
All my questions have been addressed. Thanks!

PR #1146, opened 2024-08-06
Add UserComment to Metadata Ribbon

eddiezato:
Makes the Metadata Ribbon able to display Exif.Photo.UserComment


scrubbbbs:
This is already possible if the current image contains that key.  I guess it's fine to add though.

Issue #797, opened 2021-08-08
Scan Folder Recursive no longer works


drawson1:
I've just updated to version 3.16 as part of an openSUSE upgrade; recursive scanning no longer works.

If I have:

Folder
   | - sub1
   | - sub2

and I select "Folder", only the files in "sub1" are found and displayed; the program never recurses into "sub2"

derkork:
Same problem on Windows, version 3.16.

anarcat:
> I've just updated to version 3.16 as part of an openSUSE upgrade; recursive scanning no longer works.

could you clarify what version you updated *from*?

anarcat:
@novomesk i think this is a dupe of #297.

leejuyuu:
Closing as duplicate, please continue the discussion at #297.

Issue #297, opened 2018-12-28
Scan folder recursively misbehaviour


keinerweiss:
Scenario 1:
1. Activate "File -> Scan folder recursively"
2. Click "File -> Open a directory"
3. Chose a directory with one image and at least 2 subfolders with each one image in it.
4. Click "Open"
5. The preview tiles show only the image in the selected folder and none of the subfolders

Scenario 2:
1. Activate "File -> Scan folder recursively"
2. Click "File -> Open a directory"
3. Chose a directory with zero images and at least 2 subfolders with each one image in it.
4. Click "Open"
5. The preview tiles show only the image in the first subfolder (not from the second subfolder)

diemmarkus:
hi,

thanks for reporting. the preview tiles do not properly handle recursive scans. this is because it might become very complex (i.e. if the root folder is selected) and due to design issues in our loader. I will keep this issue open, but it might take a while because we need to redesign the loader to properly fix this behavior.

best,
markus

ggoebel:
It as been 20 months. Any updates?

Current status on arch linux running nomacs v3.17.0 with "Scan folder recursively" set appears to be that it displays images found in one of the directories that was found recursively.

blissend:
Noticing similar behavior with it not including subfolders. Any plans to implement now that it's been a couple of years?

Korb:
nomacs 3.17.2206 [x64] (Aug 13 2020, Portable), OpenCV 4.3.0, Qt 5.14.2, Microsoft Windows [Version 10.0.19045.2846] — same.

Do I understand correctly that "[Scan Folder Recursive no longer works](https://github.com/nomacs/nomacs/issues/797)" (#797, Aug 8, 2021) discusses the same problem of not being able to display all images from all subfolders of the current folder, as implemented in IrfanView?

anarcat:
seeing the same issue here, still present in 3.19.

anarcat:
it looks like the code that's supposed to do this never worked. from the original implementation (https://github.com/nomacs/nomacs/commit/0da007234e5513ae1d8a2545dfe5360e24e67f0f) shipped in 1.6.4, this line was already commented:

https://github.com/nomacs/nomacs/blame/ad0c650ae56d14ef4d1fac94ff249e3987915263/ImageLounge/src/DkImage.cpp#L1601

[...]

PR #1148, opened 2024-08-18
Draft: properly recurse into subdirectories

anarcat:
It seems the original code did everything to do recursion into
subfolders except it explicitly stopped at the first folder with
images. This made browsing folders weird; it would find some
subfolder (any subfolder, not necessarily the first one listed in the
file explorer) and show *that*.

We change that behavior and just bite the bullet: we take *everything*
the QDirIterator gives us and throw that in the QFileInfoList, which
is really a QList. According to the documentation:

"This operation is relatively fast, because QList typically allocates
more memory than necessary, so it can grow without reallocating the
entire list each time."

https://doc.qt.io/qt-6/qlist.html#append

That said, obviously that list can grow to be absolutely huge, but
doing otherwise would need to refactor the loader into something that
returns an iterator (which raises the question of why we're not just
passing along that QDirIterator, probably because we're sorting
things...)

This also doesn't work so well for directories that have multiple
layers of subdirectories. From what I can tell from experience, the
images in the final list come out of order, and all mixed up
together. For example, if you have:

Photos/2024/01/01/foo.jpg
Photos/2024/01/01/bar.jpg
Photos/2024/01/02/quux.jpg

quux.jpg might be inserted between foo and bar.

On the other hand, if you have:

Photos/2024-01-01/foo.jpg
Photos/2024-01-01/bar.jpg
Photos/2024-01-02/quux.jpg

... the new code seems to handle this fine.

Either way, this *feels* to me like an improvement to the current
code, but I'll let the maintainers decide of that.

Closes: #797
Closes: #297

### DIFF
--- a/ImageLounge/src/DkCore/DkImageLoader.cpp
+++ b/ImageLounge/src/DkCore/DkImageLoader.cpp
@@ -1707,11 +1707,10 @@ QFileInfoList DkImageLoader::updateSubFolders(const QString &rootDirPath)
     // find the first subfolder that has images
     for (int idx = 0; idx < mSubFolders.size(); idx++) {
         mCurrentDir = mSubFolders[idx];
-        files = getFilteredFileInfoList(mCurrentDir,
-                                        mIgnoreKeywords,
-                                        mKeywords); // this line takes seconds if you have lots of files and slow loading (e.g. network)
-        if (!files.empty())
-            break;
+        // this takes seconds if you have lots of files and slow loading (e.g. network)
+        files.append(getFilteredFileInfoList(mCurrentDir,
+                                             mIgnoreKeywords,
+                                             mKeywords));
     }
 
     return files;


anarcat:
> This also doesn't work so well for directories that have multiple layers of subdirectories. From what I can tell from experience, the images in the final list come out of order, and all mixed up together. For example, if you have:
> 
> Photos/2024/01/01/foo.jpg Photos/2024/01/01/bar.jpg Photos/2024/01/02/quux.jpg
> 
> quux.jpg might be inserted between foo and bar.
> 
> On the other hand, if you have:
> 
> Photos/2024-01-01/foo.jpg Photos/2024-01-01/bar.jpg Photos/2024-01-02/quux.jpg
> 
> ... the new code seems to handle this fine.

actually, that's not true: the order depends on whatever the directory loader figures it should send us first. so the sorting doesn't happen at the right level.

loading 35k+ images also takes a couple of seconds, without any progress, which is a little concerning, but it works!

scrubbbbs:
Thanks for the contribution. I think you've identified the issue thoroughly, but not the motivation. The QDirIterator loop will take a very long time (perhaps minutes) to return for large directory structure on a network volume, and most of the time spent will be for directories you won't immediately (or ever) see. However, in order for sorting to work correctly you pretty much have to take everything.

Ideally we would have a separate thread to build the file list so we could start seeing thumbnails immediately, and they would periodically re-sort as more files are added. I think this is trickier than just using QDirIterator because you'd like to take into account the sorting mode (e.g. visit modified directories first if sorted by modtime).

As a simple modification to this patch I'd like to see a progress dialog with a cancel button show up if the loop is taking longer than a few seconds. This will prevent any "frozen app" dialog from showing up as well.

There is also the issue with updating the file list. Nomacs periodically rescans for changed files. If the file list is huge however, we need to make sure this won't cause the progress dialog to pop up all the time.

> For example, if you have:
> Photos/2024/01/01/foo.jpg
> Photos/2024/01/01/bar.jpg
> Photos/2024/01/02/quux.jpg
> 
> quux.jpg might be inserted between foo and bar.

Files can be sorted by name (the default), but we don't have an option for path. This would have to be added. It could be the new default since it also covers sorting by name.

I can see a use case for sorting by name even with scan-recursive enabled, so we'd have to keep that.

anarcat:
On 2024-08-18 05:49:35, scrubbbbs wrote:
> Thanks for the contribution. I think you've identified the issue thoroughly, but not the motivation. The QDirIterator loop will take a very long time (perhaps minutes) to return for large directory structure on a network volume, and most of the time spent will be for directories you won't immediately (or ever) see. However, in order for sorting to work correctly you pretty much have to take everything.

I would argue that if you're trying to recurse a large tree of
directories, *over the network*, yes, it will fail, just as it fails
now.

> Ideally we would have a separate thread to build the file list so we could start seeing thumbnails immediately, and they would periodically re-sort as more files are added. I think this is trickier than just using QDirIterator because you'd like to take into account the sorting mode (e.g. visit modified directories first if sorted by modtime).

Right, so I see the proposed implementation is not optimal, but from my
perspective it *works* and does *something* whereas the *current*
implementation... well, it basically does nothing.

So we go from "recursive just doesn't work" to "recursive hangs". Do you
*really* want to see only the former?

I don't have the cycles to rewrite all of this to have a parser thread
and progress. This requires way more intimate knowledge into both Qt
(which I last touched before Qt had major versions) and nomacs itself,
so I am almost certainly not going to be able to implement what you are
suggesting.

For what it's worth here, my full collections (36000+ photos) loads in
about 30+ seconds, before starting to display thumbs, depending on the
filesystem cache. For me it's an acceptable compromise to have a full
timeline view, but it is "hang-y": having at least a "cancel" button
would be important, I guess.

Loading this years' images (1600 so far) is almost instant.

Perhaps, as a stopgap, we could keep the limit to the number of folders
parsed and log that error?

I'll update the MR summary with a checklist as well.

anarcat:
i should also mention that if you send me a few pointers to similar sections of the code (about posting progress to the user and threading, in particular), i might be able to take a stab at this.

anarcat:
> For what it's worth here, my full collections (36000+ photos) loads in about 30+ seconds, before starting to display thumbs, depending on the filesystem cache. For me it's an acceptable compromise to have a full timeline view, but it is "hang-y": having at least a "cancel" button would be important, I guess.

actually, that's with an extra patch that removes the folder count restriction:

```diff
commit 0aa0484066a85785b19246a272ac6ab866fa591b
Author: Antoine Beaupré <anarcat@debian.org>
Date:   Sun Aug 18 01:23:37 2024 -0400

    remove artificial 100 folders limit
    
    This was breaking my "yearly display" which would hit more than 100
    folders (one per day or so).
    
    Naturally, if you use a layout like YYYY/MM/DD, you're going to hit
    that limit much faster anyway.
    
    If we *do* want to keep such a limit, we should at least make it
    configurable. I'm not sure why that limit is there anyways:
    considering the amount of debugging code leftover, I would assume it's
    a safety switch.

diff --git a/ImageLounge/src/DkCore/DkImageLoader.cpp b/ImageLounge/src/DkCore/DkImageLoader.cpp
index a0b14f52..15596632 100644
--- a/ImageLounge/src/DkCore/DkImageLoader.cpp
+++ b/ImageLounge/src/DkCore/DkImageLoader.cpp
@@ -1680,9 +1680,6 @@ QStringList DkImageLoader::getFoldersRecursive(const QString &dirPath)
             subFolders << dirs.filePath();
             nFolders++;
 
-            if (nFolders > 100)
-                break;
-
             // getFoldersRecursive(dirs.filePath(), subFolders);
             // qDebug() << "loop: " << dirs.filePath();
         }
```

so just this patch here actually works fairly well, performance-wise.

scrubbbbs:
> So we go from "recursive just doesn't work" to "recursive hangs". Do you *really* want to see only the former? 

I'd prefer something that was stable and sometimes incorrect over something that could crash/hangs.

> I don't have the cycles to rewrite all of this to have a parser thread and progress. This requires way more intimate knowledge into both Qt (which I last touched before Qt had major versions) and nomacs itself, so I am almost certainly not going to be able to implement what you are suggesting. 

Forget threads; yes that is getting too complex too quickly. QProgressDialog is simple to implement and I will be happy to add it. If you want to take a stab at it pseudo code is:

```c++
QProgressDialog dialog(this);
for (dir : dirIterator) {
  dialog.setText("scanning dir recursive %1", dirPath);
  if (!dialog.isVisible() && elapsedTime > 3000) dialog.show();
  qApp->processEvents();
  if (dialog.isCancelled()) break;
  fileList += getFileInfo(dir);
}
```

anarcat:
On 2024-08-18 12:23:49, scrubbbbs wrote:
>> So we go from "recursive just doesn't work" to "recursive hangs". Do you *really* want to see only the former? 
>
> I'd prefer something that was stable and sometimes incorrect over something that could crash/hangs.

As I said, the patch, as it is here, is actually stable and sometimes
incorrect, it shouldn't hang unless the subdirs have a huge number of
files in them (which is already an issue with the current parser).

My hangs were only happening because I removed the "100 subfolders"
limit currently in the code: I didn't actually submit that patch here,
and therefore the code is stable here, even with my 36000 image
directory (but of course incorrect).

>> I don't have the cycles to rewrite all of this to have a parser thread and progress. This requires way more intimate knowledge into both Qt (which I last touched before Qt had major versions) and nomacs itself, so I am almost certainly not going to be able to implement what you are suggesting. 
>
> Forget threads; yes that is getting too complex too quickly. QProgressDialog is simple to implement and I will be happy to add it. If you want to take a stab at it pseudo code is:
>
> ```c++
> QProgressDialog dialog(this);
> for (dir : dirIterator) {
>   dialog.setText("scanning dir recursive %1", dirPath);
>   if (!dialog.isVisible() && elapsedTime > 3000) dialog.show();
>   qApp->processEvents();
>   if (dialog.isCancelled()) break;
>   fileList += getFileInfo(dir);
> }
> ```

Are there progress dialogs in place in nomacs now? I would hate to be
the one introducing a nasty "never ending progress bar" (because that's
essentially what this will be, necessarily).

But thanks, that certainly looks very promising! I would show it within
500ms that said, 3 seconds seems like an awful time to let the user
hang without canceling, hell 100ms is a long time in UI interaction...

scrubbbbs:
> Are there progress dialogs in place in nomacs now? I would hate to be
the one introducing a nasty "never ending progress bar" (because that's
essentially what this will be, necessarily).

There are a few instances of QProgressDialog already. If it provides some information (path, file count etc) that is changing over time that would make it feel less indeterminate.

There is an unobtrusive progress animation (blue lines just under the toolbar) for loading images but no cancel button there.

> But thanks, that certainly looks very promising! I would show it within
500ms that said, 3 seconds seems like an awful time to let the user
hang without canceling, hell 100ms is a long time in UI interaction...

Well if it's too short it will flash on the screen all the time so you could annoy people either way. But yeah maybe 3 is too long. We'll have to use it for a while to see how it feels.

I did notice one issue with your patch, if I open a directory I get all the thumbnails (recursive scan), however if I open a file and then go back to the thumbnail view (shift-T), the list of thumbnails has reset to the contents of that file's directory and I can't get back to the original list of thumbs; I can get back with file explorer but have to wait for everything to load again.

anarcat:
> Well if it's too short it will flash on the screen all the time so you could annoy people either way. But yeah maybe 3 is too long. We'll have to use it for a while to see how it feels.

got it, makes sense! let's split the difference and try 1 second? :)



> I did notice one issue with your patch, if I open a directory I get all the thumbnails (recursive scan), however if I open a file and then go back to the thumbnail view (shift-T), the list of thumbnails has reset to the contents of that file's directory and I can't get back to the original list of thumbs; I can get back with file explorer but have to wait for everything to load again.

hum. no idea what's going on there. that changed with the patch?

maybe the `files` object is kept across calls? should we reset it at the start of the loop there? doubtful...

scrubbbbs:
> hum. no idea what's going on there. that changed with the patch?

No, before the patch there was always only one dir loaded at a time in `DkImageLoader`. The thumbnail scene doesn't maintain have its own file list FWICT, it only responds to changes in the image loader's current directory.

> maybe the `files` object is kept across calls? should we reset it at the start of the loop there? doubtful...

It seems like the `DkImageLoader` only wants to handle one directory at a time (see usages of `mCurrentDir` throughout). If it sees that `mCurrentDir` doesn't reflect the currently loaded image's parent dir, it sends a signal that the dir changed and the `DkThumbsScene` is reloaded.

anarcat:
I've pushed another patch here that adds a progress dialog and improves on logging a bit, see if it answers your concerns about interactivity. I wonder if, with that, we can remove the folder limit here:

https://github.com/nomacs/nomacs/blob/master/ImageLounge/src/DkCore/DkImageLoader.cpp#L1683-L1684

> It seems like the `DkImageLoader` only wants to handle one directory at a time (see usages of `mCurrentDir` throughout). If it sees that `mCurrentDir` doesn't reflect the currently loaded image's parent dir, it sends a signal that the dir changed and the `DkThumbsScene` is reloaded.

I don't know what to do about that, I think I barely understand what the problem is...

anarcat:
so i ended up actually *compiling* this code and testing it, and the progress dialog never comes up, because even my largest directory never triggers the "one second" delay. i am thinking we should remove that 100 folder limit then.

anarcat:
> ugh. so removing that limit makes it hang on something _else_: just the directory crawler seem to be taking a couple seconds itself, in my case, but the other dialog does work, and does show progress, even if it's aligned weirdly.

so just for completion's sake, i've added a progress dialog on the *other* directory iterator, and now we have two overlapping dialogs. it's not great, but it really shows how bad our walker is: first it iterates over all files, recursively, just to get a list of folders, then it looks into each of those folders individually and loads, again, all files from there. it's pretty bad!

but at least we have visible feedback, it feels pretty responsive now.

anarcat:
i tried to look at merging `getFoldersRecursive` and `updateSubFolders` and this whole thing is just a mess. there's a lot of business logic that happens in `getFilteredFileInfoList`, so we'd need to refactor that function to be able to operate on an existing file list instead, or better yet, on individual filenames so we can incrementally filter those items.

i'll also note there's a commented out `sort` call in that cursed function that might explain why we can't get nice things here.. :)

ugh. about to run out of steam here, sorry.

scrubbbbs:
> i'll also note there's a commented out `sort` call in that cursed function that might explain why we can't get nice things here.. :)

It should be equivalent to sort everything once the file list was built - but it would have to be sorted by path and not filename as it is now. And of course the dir list would be sorted at that time as well.

> ugh. about to run out of steam here, sorry.

I don't know the code too well but I'm speculating we could do this in a series of smaller bites.

Eliminating `getFoldersRecursive` and adding the progress dialog seems to be a decisive move in the right direction. Rome wasn't built in a day and all that. I think I would merge that even if it doesn't play well with other features yet.

Next, I would try to remove `mSubFolders` as a member variable and see what falls out of that. I'm speculating if we are moving to a flat filesystem model then there should be no need to track subfolders at all.

Finally, `mCurrentDir` could be removed. This seems like the toughest thing to do, but almost seems required to get us away from the current dir filesystem model.

anarcat:
On 2024-08-20 09:36:01, scrubbbbs wrote:

[...]

> Eliminating `getFoldersRecursive` and adding the progress dialog seems to be a decisive move in the right direction. Rome wasn't built in a day and all that. I think I would merge that even if it doesn't play well with other features yet.

I think that's actually a second step because you can't do that without
refactoring the filtering function.

Honestly, the code as it is now works for me and might be mergeable as
is. The only problem is the twin progress dialogs, but that seems like a
lesser evil compared to the recursive feature not working at all.

> Next, I would try to remove `mSubFolders` as a member variable and see what falls out of that. I'm speculating if we are moving to a flat filesystem model then there should be no need to track subfolders at all.

Agreed.

> Finally, `mCurrentDir` could be removed. This seems like the toughest thing to do, but almost seems required to get us away from the current dir filesystem model.

Right.

The way i see it, the next steps after this MR are:

 1. factor filtering and deduplication out of `getFilteredFileInfoList` into a standalone
    functions (one that deals with individual files, another with file
    lists, for dedupe)

 2. eliminate `getFoldersRecursive` and the duplicate directory crawling
   (and duplicate progress dialog)

 3. cleanup member variables (`mSubFolders` and `mCurrentDir`)

 4. setup post-crawling sorting (and, why not, with EXIF date sorting?)

scrubbbbs:
I played with this a bit and committed some changes and additions in [my deep-fs branch](https://github.com/scrubbbbs/nomacs/tree/deep-fs). However, I do not think it is ready to be merged into master.

Nomacs has developed a reputation for being buggy and lacking polish. So I am hesitant to do anything that will introduce new issues or break existing functionality. The short list of issues on my mind now:

- Opening any image can can cause a very long pause (albeit with progress dialog) while all subfolders are scanned. IMO this is the biggest issue, and was not an issue before with subfolders enabled.
- "Filter Files" on thumbnail preview is broken
- (already mentioned) Going from image back to thumbnail preview reverts to the old behavior and the subfolder scan is lost.

If you want to continue working on it, I'd be happy to create a new branch in nomacs or we can keep this PR open and continue working from your branch.

scrubbbbs:
I'm closing since this has gone dormant. There is some current interest from @maurolacy in having a list-based mode for files (vs the current directory-based mode), I think this concept could be folded into that.

Issue #1139, opened 2024-07-25
Batch processing crop fails


leejuyuu:
i just installed 3.19.0, now the batch processing fails:
```
processing /Users/xxx/Downloads/FCI_hres_OPE__true_color_flames_20240723T033000.png
[Transform Batch] image 22272 x 22272 cropped to x12971 y3334 w1880 h1136
Original filename added to Exif
Could not save: /Users/xxx/Downloads/cropped/FCI_hres_OPE__true_color_flames_20240723T033000
Ui - a lot of things went wrong. Your original file can be found here: 
Unknown error
I did not delete the original because I detected 2 failure(s).
```

the same issue seems to also exist with the latest `master`.

_Originally posted by @sjoro in https://github.com/nomacs/nomacs/issues/1138#issuecomment-2250042688_

leejuyuu:
Confirmed on Arch, nomacs 3.19.0 (eca2377d).

Steps to reproduce:

1. Select one input file (I used png) in batch processing.
2. Setup Transform > Crop with a crop rectangle with some size (eg. 20 x 20).
3. On the Output page, select an empty directory, and set the settings as follows
![image](https://github.com/user-attachments/assets/2fd13a41-37a1-4171-a7b3-50c4fdc171a8)

4. Execute the batch processing, and observe the same message in the batch log.

leejuyuu:
@sjoro The batch processing functions when changing the "Keep Extension" in the "Filename" block to "Convert To". Please see if this works for you.

I see that the New Filename has no extension, could be the source of the problem.

leejuyuu:
@sjoro Thanks for testing! Does the crop rectangle position work correctly for you as well?

sjoro:
yes, works as expected.

PR #1150, opened 2024-08-19
Rewrite DkFileNameConverter

leejuyuu:
The issue https://github.com/nomacs/nomacs/issues/1139 is caused by `DkFileNameConverter` not parsing the input pattern correctly. For example, "<c:0>.<old>" is interpreted as "<c:0>" because that the regex "<.*>" gets the whole string instead of two groups. This rewirtes `DkFileNameConverter` to parse the pattern manually instead of using regex.

The signature of the converter is also changed, so we can create one converter and reuse it instead of creating one in each iteration.

Fixes https://github.com/nomacs/nomacs/issues/1139.

### DIFF
--- a/ImageLounge/src/DkCore/DkUtils.cpp
+++ b/ImageLounge/src/DkCore/DkUtils.cpp
@@ -981,11 +981,112 @@ QString DkUtils::stdWStringToQString(const std::wstring &str)
 }
 
 // DkConvertFileName --------------------------------------------------------------------
-DkFileNameConverter::DkFileNameConverter(const QString &fileName, const QString &pattern, int cIdx)
+DkFileNameConverter::DkFileNameConverter(const QString &p)
+    : mFrags{}
 {
-    this->mFileName = fileName;
-    this->mPattern = pattern;
-    this->mCIdx = cIdx;
+    const QStringView pattern(p);
+    qsizetype start = 0;
+    Token type = Token::Text;
+    std::vector<uint> numbers{};
+    FragType fragType = FragType::Text;
+
+    for (qsizetype i = 0; i < pattern.size(); i++) {
+        const QChar ch = pattern.at(i);
+
+        if (ch != '<' && ch != '>' && ch != ':') {
+            continue;
+        }
+
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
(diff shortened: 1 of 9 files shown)


ImageLounge/src/DkCore/DkUtils.cpp
```diff
@@ -1001,96 +1102,41 @@ DkFileNameConverter::DkFileNameConverter(const QString &fileName, const QString
...
+        case FragType::Text:
+            list.append(frag.text);
+            break;
+        case FragType::FileName: {
+            const qsizetype sep = file.lastIndexOf('.');
 
-        // zero padding
-        for (int idx = 0; idx < cNumZeros; idx++) {
-            result += "0";
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
```
scrubbbbs, line 1117: Could use QString:mid(0, len), works on both versions

ImageLounge/src/DkCore/DkUtils.cpp
```diff
@@ -981,11 +981,112 @@ QString DkUtils::stdWStringToQString(const std::wstring &str)
...
+    FragType fragType = FragType::Text;
+
+    for (qsizetype i = 0; i < pattern.size(); i++) {
+        const QChar ch = pattern.at(i);
+
+        if (ch != '<' && ch != '>' && ch != ':') {
+            continue;
+        }
+
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
```
scrubbbbs, line 1000: I think we should aim to avoid conditional code as much as possible unless there is a meaningful performance or functional impact. 

I believe QString:mid() works on either version here and also for left()/first()

leejuyuu: These are `QStringView`s, and the `mid` method is [deprecated in Qt6](https://doc.qt.io/qt-6/qstringview-obsolete.html#mid).

For `QString::mid()`, there are performance gains according to [the documentation](https://doc.qt.io/qt-6/qstring.html#mid) if the indices are within bounds. 

Because the API is different enough, I think we should drop support of Qt5 once things stabilize. Although these conditionals are ugly, they are easy to spot and can make removing Qt5 code as easy as removing one branch, and we are not losing things due to being locked to old APIs.

PR #1164, opened 2024-09-15
macOS startup fixes

hmueller01:
This PR fixes the problem in macOS that double clicking an image file did not open the file any more.
After a long debugging session (and adding log output to a file, as console output does not help if app is installed on macOS) I found that a deprecated Qt define was used: Q_WS_MAC
See also https://doc.qt.io/qt-5/macos-issues.html#compile-time-flags

Also added a flag to macdeployqt to always create the dmg file (and do not complain).


novomesk:
It seems the submodules in 3rd-party folder point to old commits.

hmueller01:
Hm,  I see. But how can I update this?
```
> git remote -v
origin  https://github.com/hmueller01/nomacs.git (fetch)
origin  https://github.com/hmueller01/nomacs.git (push)
upstream        https://github.com/nomacs/nomacs.git (fetch)
upstream        https://github.com/nomacs/nomacs.git (push)

> git fetch upstream
Fetching submodule 3rd-party/LibRaw
Fetching submodule 3rd-party/exiv2
Fetching submodule 3rd-party/expat
Fetching submodule 3rd-party/imageformats
Fetching submodule 3rd-party/opencv
Fetching submodule 3rd-party/quazip
Fetching submodule ImageLounge/plugins

> git merge upstream/master
Already up to date.

```

scrubbbbs:
> Hm, I see. But how can I update this?

I'm not sure, but maybe this. The submodule init/update is definitely needed. The amend/force might be needed.

```
git submodule init
git submodule update
git commit [-amend]
git push [-force]
```

hmueller01:
I do not know what went wrong. But nothing helped. At the end of the day I checked out the submodules manually to the commit id of nomacs master and committed this. This PR does not change the submodules any more. It might be better to squash the merge ...

ImageLounge/src/main.cpp
scrubbbbs, line 1: We already have logging to file. How is this different?

Can we drop `Q_WS_MAC` it and just use `Q_OS_MAC`... this is what we use everywhere else

hmueller01: Doh. You are right. I didn't know / see this. Removed code.

Dropped Q_WS_MAC.

README.md
```diff
@@ -247,23 +256,35 @@ cmake --build . --parallel 8
...
-````bash
+```bash
 docker login
 docker tag nomacs diemmarkus/nomacs
 docker push diemmarkus/nomacs:latest
-````
+```
+
+
+## Enable Debug output
```
novomesk, line 278: This part regarding Debug output is no longer valid.

hmueller01: Ok, you are right. I just looked in the code and if I understand that right debug messages are not created at all!

https://github.com/nomacs/nomacs/blob/84bfa5e6617d3fc133c921687f8f6fed96007c2f/ImageLounge/src/DkCore/DkUtils.cpp#L438

The Qt default behavior activates qDebug() output by not defining `QT_NO_DEBUG_OUTPUT` and this is done by `CMAKE_BUILD_TYPE=Debug`. But now I don't get qDebug() output even with this define. Have I overseen something?

hmueller01: I would suggest changing 
https://github.com/nomacs/nomacs/blob/84bfa5e6617d3fc133c921687f8f6fed96007c2f/ImageLounge/src/DkCore/DkUtils.cpp#L438
to
```
    QString p = "%{if-info}[INFO] %{endif}%{if-warning}[WARNING] %{endif}%{if-critical}[CRITICAL] %{endif}%{if-fatal}[ERROR] %{endif}";
#ifndef QT_NO_DEBUG_OUTPUT
    p += "%{if-debug}[DEBUG] %{endif}";
#endif
    p += "%{message}";

```

hmueller01: If you like I can do a new PR to not mix it up here. Just let me know.

novomesk: close this and make a new PR.

hmueller01: Ok, I have undone this change and will open a new PR addressing the debug issue. Would be nice to merge this now. Tx.

PR #1166, opened 2024-09-20
Enable debug output

hmueller01:
added possibility to output qDebug() messages, fixed logfile output

PR #1170, opened 2024-09-26
logToFile cleanup

hmueller01:
This fixes #1166 

But I found another file that uses the logToFile() function
https://github.com/nomacs/nomacs/blob/9a37a252804cf8934a27e4a37951732e19d4bc03/ImageLounge/src/DkGui/DkLogWidget.cpp#L100
This won't be formatted correctly atm. I do not understand where this get's used. Does anybody know?


hmueller01:
I found the way where logToFile() is used in DkLogWidget.cpp. It can be activated by Panels -> Show Log and redirects the log output to a panel. That easy. Wow. Why didn't I find this before ... :-)
Ok, output is in different colors. I think the added [INFO], etc. is not needed here. 
Thanks for merging!

Issue #711, opened 2021-07-06
[Usability]/[Adjustment panel] The picture jumps and twitches on the main screen and on the thumbnail.


RatatataBoom:
Steps to reproduce: 
1. Open file (any image)
2. Click the Adjustment menu.
3. Press "Rotation" or "Scale", "Threshold", "Saturation", etc.
4. Slowly move the scroll bar left or right. 

Expected result: 
The picture smoothly changes in accordance with the variable parameter on the main screen
and on the thumbnail

Actual result:
The picture jumps and twitches on the main screen and on the thumbnail.

Attachments: 
https://drive.google.com/file/d/1bJzhXVoifobPcbyQNvfWEkUOhhadmaOZ/view?usp=sharing

Environment:

Win10 (19042.1052)
Nomacs ver: 3.16.1709[x64]

leejuyuu:
Confirmed on 3.19.0 (eca2377d), Arch Linux.

Issue #665, opened 2021-05-23
While togging scale 'resize' image gets big for a split second


Ambidre:
**Describe the bug**
Image resized but when togging scale “resize” the image gets big for a split second

**To Reproduce**
1. Go to 'Adjustments'
2. Click on 'Resize'
3. Tog scale 'Resize'

**Expected behavior**
Image resized

**Desktop (please complete the following information):**
- OS: Windows 10 Education
- Version 3.16.1709 [x64]

leejuyuu:
I believe this has the same cause as #711.

PR #1192, opened 2024-11-18
fix: prevent viewport update when applying manipulator

leejuyuu:
The same manipulator were undoed before applying a new one and triggers
a viewport paintEvent, causing the image to change back to original when
dragging the rotation or scaling slidebar.

Modify the undo call to avoid triggering paintEvent.

Fixes https://github.com/nomacs/nomacs/issues/665.
Fixes https://github.com/nomacs/nomacs/issues/711.


leejuyuu:
Thanks for testing.

> However, the preview image can still be out of sync

I am aware of this. I plan to redesign the manipulator widgets. That would be in another PR though. Some plans:

- Make the preview actually the preview, i.e. dragging the slider does not update the viewport, only the preview.
- Add an "Apply" button to commit the effect/editing. Maybe a "Reset" to reset the slider.
- Remove the undo and redo buttons.

scrubbbbs:
It would be cool if the main view showed the real-time preview instead of having it occupy space in the manipulator and updating both of them.

PR #1195, opened 2024-11-24
Fix mac quazip from source

hmueller01:
Update `-D ENABLE_QUAZIP=ON -D USE_SYSTEM_QUAZIP=OFF` build on macOS. Was a bit outdated. Seems be unused for quite a while.


scrubbbbs:
Thanks! Works for me in all modes on 12.7. I noticed nomacs will crash trying to open a zip if quazip is disabled, I'm only getting on the MacOS version and it was also present before this patch.

PR #1202, opened 2024-11-27
Build: Disable version update script in default build

scrubbbbs:
versionupdate.py causes parts of nomacs to be recompiled/relinked even if the version did not change, which slows down development builds considerably.

This script is only useful when the incrementor is also enabled, therefore it is moved to that block in cmake.

Therefore, the default build will always use the version defined in CMakeLists.txt, and never the incrementor cache file.

Also adds cmake message with nomacs version and a warning if incrementor is enabled since that could publish the wrong version.

NOTE: This PR needs to be validated with VSCode build, need to check of the versions in the .exe files is present and correctly reflects nomacs version.


scrubbbbs:
I have verified that appveyor build version numbers are correct after this change, and bumping the version works as expected. The version is not being bumped here, but a newline was added to version.cache for test purposes.

Issue #1172, opened 2024-09-30
Linker warnings on macOS


hmueller01:
During compilation of the plugins I get these warnings:
```
[ 85%] Linking CXX shared library ../libpaintPlugin.dylib
ld: warning: ignoring duplicate libraries: '-lc++'
ld: warning: search path '.../nomacs.hmueller01/build/libs' not found
```
There is no libs folder within build. So the warning is correct. Is it created at Linux or Windows?
Otherwise
`${NOMACS_BUILD_DIRECTORY}/libs
`
at 
https://github.com/novomesk/nomacs-plugins/blob/18861dc71c015006bfc551090a30e96b4fa37c78/CompositePlugin/CMakeLists.txt#L52
could be removed ...

I didn't find the reason for the duplicate libraries: '-lc++' yet. Is it on other platforms as well?

scrubbbbs:
The lib folder is not on other platforms (864bb5b). I'm not seeing this on MacOS 12.7.6  / clang 14 / cmake 3.30.2. I am building out of tree in `~/tmp/build.nomacs` like so:

`cmake -DENABLE_QUAZIP=ON ~/src/nomacs/ImageLounge`

hmueller01:
Ok, 
`nomacs.hmueller01/build> cmake -D QT_VERSION_MAJOR=6 -D ENABLE_QUAZIP=ON --install-prefix /Applications ../ImageLounge/`
does not compile on my system (macOS Sonoma 14.7.1 / clang 15 / cmake 3.30.5)
```
...
[ 37%] Building CXX object CMakeFiles/nomacsCore-3.19.1.dir/qrc_nomacs.cpp.o
[ 38%] Building CXX object CMakeFiles/nomacsCore-3.19.1.dir/3rdparty/libqpsd/qpsdhandler.cpp.o
[ 39%] Building CXX object CMakeFiles/nomacsCore-3.19.1.dir/3rdparty/libqpsd/qpsdhandler_p.cpp.o
[ 40%] Building CXX object CMakeFiles/nomacsCore-3.19.1.dir/3rdparty/libqpsd/qpsdplugin.cpp.o
make[2]: *** No rule to make target `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.sdk/usr/lib/libbz2.tbd', needed by `libnomacsCore-3.19.1.dylib'.  Stop.
```
[...]

PR #1203, opened 2024-11-28
macOS: Fix missing libs warning

hmueller01:
see #1172 

Does not fix `ld: warning: ignoring duplicate libraries: '-lc++'` in all the plugins ...
ATM only for `libnomacsCore-3.19.1.dylib` and `nomacs.app/Contents/MacOS/nomacs`


scrubbbbs:
I'm thinking it is probably not a great idea to drop the warning, it must exist for a reason, and I don't understand what that reason is. Even though in this case (-lc++) it is innocuous, I'm wondering if this could result in future issues or mask problems that require our attention.

Also, there is another proposed solution in the issue you linked in #1172 which does not involve disabling the warning (reinstalling gcc was suggested).

hmueller01:
It just ignores double -l options. And if you write `gcc -lc++ -lc++ ...` who cares. This warning is useless.
Of course I tried the reinstalling of gcc but it didn't help in my case. But if you do not like the solution I'll remove this part of the PR.

scrubbbbs:
> This warning is useless

Maybe, maybe not. At least it can change the link order of the binary which can cause subtle changes. Since nomacs uses a shared library which is also linked to plugins, we need to pay extra attention to link warnings.

I'd like to remove it, thanks!

hmueller01:
Ok, fair enough. Removed that.

Issue #1001, opened 2023-11-04
Wrong files displayed when browsing quickly


tkzv:
**Describe the bug**
When I browse a directory by pressing and holding an arrow, sometimes images are displayed in wrong order: an image appears only briefly and is replaced by a previous one. For example, the folder has images from 1.png to 1000.png; I start at 1, Nomacs correctly displays 1, 2, 3, 4, briefly shows 5, then quickly replaces it with 4, then displays 6, briefly displays 7, returns to 6, displays 8, 9, 10... This displaying next file and reverting to previous one also creates unpleasant flickering. Meanwhile, the names in the title bar are incremented correctly, so, when the titlebar says "5.png" Nomacs displays 4.png.

This is rather unpredictable — there is no way to tell which files will be displayed and which will be rolled back this particular time. The problem is more likely to manifest when large images are displayed scaled down and/or when there are other processes heavily loading the hard drive and the CPU. I've encountered it for various file types: JPEG, PNG, WebP.

The problem also may manifest when browsing files by pressing (rather than holding) an arrow for each image, if you press it quickly.

Not reproduced when "Image Loading Policy" is "Skip Images".

**To Reproduce**
1. Create a series of large images with numbers. For example, in Linux:
```
for n in `seq -w 1 1000`; do magick -size 10000x5000 label:$n text-$n.webp; done
```
(requires bash, coreutils and ImageMagick)

2. Open "text-0001.webp" in Nomacs. Zoom out to something like 6%.

3. Press and hold right arrow. The displayed numbers would change something like this:
0001
0002
0003 (very briefly)
0002
0004
0005 (very briefly)
0004
0006
0007 (very briefly)
0006
0008
...

**Expected behavior**
The images should be displayed in their correct order: 0001, 0002, 0003, 0004, 0005, 0006, 0007, 0008...

**Screenshots**
Example of browsing:
[obs-capture.zip](https://github.com/nomacs/nomacs/files/13256994/obs-capture.zip)

**Desktop:**
 - OS: Gentoo Linux
 - Versions: kernel: 5.15.32-r1, dev-qt/qtcore-5.15.10-r2, x11-base/xorg-server-21.1.9, media-libs/opencv-4.8.0-r1, media-gfx/nomacs-3.16.224-r6

olodar:
I wondering how they still not care about this bug

scrubbbbs:
Can confirm, git master 6dfc779, Linux, X11, qt 5.15.10

Settings:
[...]

PR #1205, opened 2024-12-01
Fix the loading issue when browsing quickly

leejuyuu:
Fixes https://github.com/nomacs/nomacs/issues/1001.
Also some refactorings.


scrubbbbs:
Seems to work fine. I was unsure about removing the delay/timer but it behaves the same also with the slideshow animation.

leejuyuu:
> I was unsure about removing the delay/timer but it behaves the same also with the slideshow animation.

Do you have concerns about removing the timer?

The thing I am still working on for this PR is that the title still updates slightly faster than the image if I test in a (quite limited) VM.

scrubbbbs:
> > I was unsure about removing the delay/timer but it behaves the same also with the slideshow animation.
> 
> Do you have concerns about removing the timer?

When I see a timer like that my first though is it was for event compression; to maintain responsiveness or maybe to avoid using too much memory usage. I'm not seeing any responsiveness or memory issues so I think you are OK to remove it.

> The thing I am still working on for this PR is that the title still updates slightly faster than the image if I test in a (quite limited) VM and open the thumbnails panel.

Yes, I missed that. Easy to see when running under valgrind!

leejuyuu:
Regarding the lag between the title update and the image display, the order of operations seems correct (image loaded -> the new image is set in the image storage -> update title -> paint viewport). I think the problem is that the image is large, and we use the original one initially when high-quality anti-aliasing is enabled, so it always lags a bit behind the title update. One way is to display a fast down-scaled copy before the HQ anti-aliased image is ready, but this feels difficult given the current implementation. Better leave it in future improvements.

I optimized the image loading path instead. It might reduce the load and seem to make things a bit smoother.

This PR is ready for testing and review.

ImageLounge/src/DkCore/DkImageStorage.cpp
```diff
@@ -280,6 +287,86 @@ QImage DkImage::rotateImage(const QImage &img, double angle)
...
+    if (img.depth() > 32) {
+        // Ensure the data is 32-bit aligned
+        imgIn = img.convertToFormat(QImage::Format_ARGB32);
+    }
+
+    QImage imgOut = QImage(imgIn.size().transposed(), imgIn.format());
+
+    const QRgb *dataIn = reinterpret_cast<const QRgb *>(imgIn.constBits());
+    QRgb *dataOut = reinterpret_cast<QRgb *>(imgOut.bits());
+
```
scrubbbbs, line 302: I recommend using QImage::bytesPerline() for the row stride, or adding a check. This allows doing a zero-copy conversion from image buffers from another library (opencv or libraw, openGL, whatever, in a future-proof way.

ImageLounge/src/DkCore/DkImageStorage.cpp
```diff
@@ -280,6 +287,86 @@ QImage DkImage::rotateImage(const QImage &img, double angle)
...
+    angle = std::fmod(angle, 360);
+    if (angle < 0) {
+        angle += 360;
+    }
+
+    if (angle == 0) {
+        return img;
+    }
+
+    if (angle == 90 || angle == 180 || angle == 270) {
```
scrubbbbs, line 340: You might consider QImage::mirrored() if you are going to stack operations anyways.

scrubbbbs:
> (image loaded -> the new image is set in the image storage -> update title -> paint viewport).

If the other data (metadata/status bar/notes etc) is not updated until painting, is there a way to make the title use the same code path? I'm OK to leave it unsolved if the solution is complex and this doesn't give much benefit.

ImageLounge/src/DkCore/DkImageStorage.cpp
leejuyuu: I tried this, and it's actually slightly faster. It can work for 90 and 270 degrees.

ImageLounge/src/DkCore/DkImageStorage.cpp
leejuyuu: Not sure what you want to check for?

Following the docs, I ensure the image is 32-bit aligned, then cast the data pointer to `QRgb*`. The row stride is then the width.

image

leejuyuu:
> > (image loaded -> the new image is set in the image storage -> update title -> paint viewport).
> 
> If the other data (metadata/status bar/notes etc) is not updated until painting, is there a way to make the title use the same code path? I'm OK to leave it unsolved if the solution is complex and this doesn't give much benefit.

Not that I know of.  The metadata etc. are triggered by `imageLoaded`, while the title updates when `imageUpdated`, which is always emitted after `imageLoaded` and in other places like editing. The order of the slots getting called seems correct. But I do not see an easy way to delay the title further.

I also suspect that the window title update might be separated from painting the widgets.

ImageLounge/src/DkCore/DkImageStorage.cpp
scrubbbbs: > The row stride is then the width.

I'm reading that it must be 32-bit aligned *at minimum*, so it could be 32,64,128, etc. In which case there can be unused bytes at the end of each row that must be considered. The QImage constructors that take a pointer and bytesperline argument might produce such an image, we might want to use these constructors at some point when working with large images, since they don't have to make a copy.

So I believe you must check if  bytesPerLine()==width()*bytesPerPixel() if you are wanting to treat the image  buffer as a continuous block.

leejuyuu: > it must be 32-bit aligned at minimum, so it could be 32,64,128

Yes, it must be 32-bit aligned at minimum, and 32-bpp images are also 32-bit aligned, otherwise they cannot be casted to `QRgb*` safely. So, we only need format conversion if we have 64-bit image formats, as they can be 32-bit or 64-bit aligned depending on platforms.

> The QImage constructors that take a pointer and bytesperline argument might produce such an image, we might want to use these constructors at some point when working with large images, since they don't have to make a copy.

I would probably avoid using these constructors because the resulting `QImage` will not own the memory. I try leveraging the copy-on-write (or "implicit data sharing" from the Qt docs) property of the `QImage` as much as possible. When there's a need, I can allocate a `QImage` from a known size, and access the memory directly with `bits()` or `constBits()`, so the memory stays managed by the `QImage`.

scrubbbbs:
The fix seems good, I'm not seeing any issues.

Two issues with the benchmarks
- I cannot link the bench without "make install" first which is moderately annoying, but I can live with it
- I get a bunch of errors from Qt (both 5 and 6), like:

```
QPainter::begin: Paint device returned engine == 0, type: 3
QPainter::setRenderHint: Painter must be active to set rendering hints
QPainter::setWorldTransform: Painter not active
```
Which means the test is broken for me. I am on xorg, so maybe that's why it doesn't show up for you. QPainter is part of QTGui, which usually will require a window or some context to be present (maybe QApplication is enough here, IDK).

scrubbbbs:
It seems this patch has broken antialiasing. Antialiasing is not being re-applied when zooming or if the image size otherwise changes. (fit to window, toggle fullscreen, etc)

leejuyuu:
> The fix seems good, I'm not seeing any issues.
> 
> Two issues with the benchmarks
> 
>     * I cannot link the bench without "make install" first which is moderately annoying, but I can live with it
> 
>     * I get a bunch of errors from Qt (both 5 and 6), like:
> 
> 
> ```
> QPainter::begin: Paint device returned engine == 0, type: 3
> QPainter::setRenderHint: Painter must be active to set rendering hints
> QPainter::setWorldTransform: Painter not active
> ```
> 
> Which means the test is broken for me. I am on xorg, so maybe that's why it doesn't show up for you. QPainter is part of QTGui, which usually will require a window or some context to be present (maybe QApplication is enough here, IDK).

This seems odd. Did you run the benchmark inside the build directory?

leejuyuu:
> It seems this patch has broken antialiasing. Antialiasing is not being re-applied when zooming or if the image size otherwise changes. (fit to window, toggle fullscreen, etc)

Nice catch! Should be fixed now.

PR #1213, opened 2024-12-07
Doc: various cleanup and update build recipes

scrubbbbs:
- move common git clone stuff to the top
- add Arch, RedHat, FreeBSD, Haiku
- remove old Ubuntu package lists; omit 20.04 as it is EOL in 2025

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,19 @@
 # nomacs - Image Lounge 🍸
 
-nomacs is a free, open source image viewer, which supports multiple platforms. You can use it for viewing all common image formats including RAW and psd images. nomacs is licensed under the GNU General Public License v3 and available for Windows, Linux, FreeBSD, Mac, and OS/2.
+nomacs is a free, open source image viewer, which supports multiple platforms. You can use it for viewing all common image formats including RAW and psd images. nomacs is licensed under the GNU General Public License v3 and available for Windows, Linux, FreeBSD, Mac, Haiku, and OS/2.
 
 [![Build status](https://ci.appveyor.com/api/projects/status/ye6wd1hap4cebyo8?svg=true)](https://ci.appveyor.com/project/novomesk/nomacs)
 [![Downloads](https://img.shields.io/github/downloads/nomacs/nomacs/total.svg)](https://github.com/nomacs/nomacs/releases/latest)
 [![Crowdin](https://badges.crowdin.net/nomacs/localized.svg)](http://translate.nomacs.org/project/nomacs)
+
+## Get the Source
+
+```console
+git clone https://github.com/nomacs/nomacs.git
+cd nomacs
+git submodule init
+git submodule update
+```
 
 ## Build nomacs (Windows)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,11 +1,20 @@
...
 [![Downloads](https://img.shields.io/github/downloads/nomacs/nomacs/total.svg)](https://github.com/nomacs/nomacs/releases/latest)
 [![Crowdin](https://badges.crowdin.net/nomacs/localized.svg)](http://translate.nomacs.org/project/nomacs)
 
+## Get the Source
+
+```console
+git clone https://github.com/nomacs/nomacs.git
+cd nomacs
+git submodule init
+git submodule update
```
leejuyuu, line 15:
```suggestion
git clone https://github.com/nomacs/nomacs.git
cd nomacs

# nomacs uses submodules for third party libraries
git submodule init
git submodule update
```

README.md
```diff
@@ -48,63 +57,125 @@ python scripts/make.py "qt/bin" --lib-path "nomacs/3rd-party/build"
...
-## Build nomacs (Ubuntu)
+## Build nomacs (Linux/Unix)
 
-Get the required packages:
+Before you build nomacs, please note the following:
 
-- For Ubuntu 21.04 and above:
+- We recommend using the Qt6 version of nomacs. However, you may want to match the Qt version of your desktop environment (e.g. KDE in 24.04 is Qt5-based).
+- [kimageformats-plugins](https://github.com/KDE/kimageformats) provides additional formats (AVIF,HEIC/HEIF) and is sometimes only available for the Qt5 version. Otherwise, you must compile it separately.
+- Zip file support requires Quazip, which has varied support in distributions. If the system package is missing or does not work, you can set `USE_SYSTEM_QUAZIP=NO` to use the version in nomacs/3rdparty. However, you may also need to remove the any system quazip development package temporarily. (`llibquazip*-dev`)
```
leejuyuu, line 66:
```suggestion
- We recommend using the Qt6 version of nomacs. However, you may want to match the Qt version of your desktop environment (e.g. KDE in Ubuntu 24.04 is Qt5-based).
- [kimageformats-plugins](https://github.com/KDE/kimageformats) is an optional dependency that provides additional formats such as AVIF and HEIC/HEIF. The Qt version of the plugins should match the Qt version when compiling nomacs.
- Zip file support requires Quazip, which has varied support in distributions. If the system package is missing or does not work, you can set `USE_SYSTEM_QUAZIP=NO` to use the version in nomacs/3rdparty. However, you may also need to remove the any system quazip development package temporarily. (such as `llibquazip*-dev` for Ubuntu)
```

README.md
```diff
@@ -48,63 +57,125 @@ python scripts/make.py "qt/bin" --lib-path "nomacs/3rd-party/build"
...
+
+# qt5
+sudo apt install qtbase5-dev qttools5-dev libqt5svg5-dev qt5-image-formats-plugins libexiv2-dev libraw-dev libopencv-dev libtiff-dev libtiff-dev libquazip5-dev build-essential git cmake lcov libgtest-dev kimageformat-plugins
+```
+
+#### Arch
+
+```console
+# qt6
+sudo pacman --sync qt6-base qt6-imageformats qt6-svg qt6-tools quazip-qt6 kimageformats exiv2 libraw libtiff opencv kimageformats git cmake gtest base-devel make
```
leejuyuu, line 95: I think it's more often written as `-S` than `--sync`. Also, `make` is already in `base-devel`.

```suggestion
sudo pacman -S qt6-base qt6-imageformats qt6-svg qt6-tools quazip-qt6 exiv2 libraw libtiff opencv kimageformats git cmake gtest base-devel
```

README.md
```diff
@@ -48,63 +57,125 @@ python scripts/make.py "qt/bin" --lib-path "nomacs/3rd-party/build"
...
 
-This will by default place the source into ~/nomacs
-Go to the nomacs/ImageLounge directory and run `cmake` to get the Makefiles:
+### Configure nomacs
+
+Nomacs is configured with cmake. These cmake options are often needed:
+- QT_VERSION_MAJOR=[5|6] - Default 5, 6 recommended
+- ENABLE_QUAZIP=[ON|OFF] - Default OFF, ON recommended
+- USE_SYSTEM_QUAZIP=[ON|OFF] - Default ON, recommended
+- CMAKE_BUILD_TYPE=[Debug|Release|RelWithDebInfo] - Debug builds have more logging as well as debug symbols.
```
leejuyuu, line 130: I suggest removing the recommendations here because quazip is an optional dependency.

```suggestion
- ENABLE_QUAZIP=[ON|OFF] - Default OFF
- USE_SYSTEM_QUAZIP=[ON|OFF] - Default ON
- CMAKE_BUILD_TYPE=[Debug|Release|RelWithDebInfo] - For normal usage, choose `RelWithDebInfo`. For development, `Debug` builds have more logging as well as debug symbols.
```

README.md
```diff
@@ -48,63 +57,125 @@ python scripts/make.py "qt/bin" --lib-path "nomacs/3rd-party/build"
...
+
 
 ```console
+cd nomacs
 mkdir build
 cd build
-cmake ../ImageLounge/.
+cmake -D QT_VERSION_MAJOR=5 -D ENABLE_QUAZIP=ON -D USE_SYSTEM_QUAZIP=ON ../ImageLounge
```
leejuyuu, line 137: I suggest removing the `-D` options because (1) Qt6 is recommended above so make it consistent and (2) quazip is an optional dependency.

```suggestion
cmake ../ImageLounge
```

README.md
```diff
@@ -48,63 +57,125 @@ python scripts/make.py "qt/bin" --lib-path "nomacs/3rd-party/build"
...
+# running nomacs from ~/nomacs/build/
+LD_LIBRARY_PATH=.:$LD_LIBRARY_PATH ./nomacs
 ```
 
-note that you have to execute
+## Build and run tests
 
-```console
-sudo ldconfig
+Nomacs uses googletest (libgtest-dev), this is included in package lists above. A `check`target will be created if GoogleTest is present. To build and run tests:
```
leejuyuu, line 155: I would avoid using Ubuntu package names if this section is for generic Linux.

```suggestion
Nomacs uses [GoogleTest](https://github.com/google/googletest), which is included in the package lists above. A `check` target will be created if GoogleTest is present. To build and run tests:
```

README.md
```diff
@@ -48,63 +57,125 @@ python scripts/make.py "qt/bin" --lib-path "nomacs/3rd-party/build"
...
+# Some systems you also require changes to 
+# `/etc/ld.so.conf` to add `/usr/local/lib` 
+sudo ldconfig
+```
 
 ### For Package Maintainers
 
-- Set `ENABLE_TRANSLATIONS` to `true` (default: `false`)
-- Build all officially supported [plugins](https://github.com/nomacs/nomacs-plugins/)
+- Recommended additional packages: kimageformats provides HEIF/HEIC and AVIF image loading, and many others
```
leejuyuu, line 176:
```suggestion
- Recommended optional dependency: kimageformats provides HEIF/HEIC and AVIF image loading, and many others
```

README.md
```diff
@@ -158,7 +223,8 @@ You will now have a binary (`nomacs.app`), which you can test (or use directly).
 sudo make install
 ```
 
-If you want to have an independent bundle image (`nomacs.dmg`) you can create it by using
+If you want to have an independent bundle image (`nomacs.dmg`) you c
+an create it by using
```
leejuyuu, line 227:
```suggestion
If you want to have an independent bundle image (`nomacs.dmg`) you can create it by using
```

README.md
scrubbbbs: I don't feel RellWithDebInfo is worthwhile, as it will build debug symbols.. but I guess they could appear in crash reports. Is that your thinking?

README.md
```diff
@@ -48,63 +59,125 @@ python scripts/make.py "qt/bin" --lib-path "nomacs/3rd-party/build"
...
 
-## Build nomacs (Ubuntu)
+## Build nomacs (Linux/Unix)
 
-Get the required packages:
+Before you build nomacs, please note the following:
 
-- For Ubuntu 21.04 and above:
+- We recommend using the Qt6 version of nomacs. However, you may want to match the Qt version of your desktop environment (e.g. KDE in Ubuntu 24.04 is Qt5-based).
+- [kimageformats-plugins](https://github.com/KDE/kimageformats) is an optional dependency that provides additional formats such as AVIF and HEIC/HEIF. The Qt version of the plugins should match the Qt version when compiling nomacs.
```
novomesk, line 67: The original repo is https://invent.kde.org/frameworks/kimageformats

You can mention also JPEG XL (JXL) format which is also supported by kimageformats.

README.md
leejuyuu: Yes, stack traces will be available if the build has debug symbols.

leejuyuu:
Reading the dependency installation parts again, would it be more friendly to packagers that we separate the optional dependencies like image format plugins, quazip, and gtest?

scrubbbbs:
After testing a lot of different distros, I'm confident we can remove quazip as an optional dependency. 3rdparty/quazip works perfectly on any distribution in case there is a conflict with the quazip package.  Yes, the feature has a few bugs/quirks but nothing major.

Image format plugins should not be optional, it is a small package. kimageformats is a bit heavy so yeah, we should call it out to package maintainers.

I think gtest should be included if possible, I haven't seen any distro where it isn't available. Maintainers should be running tests to validate the build when they are able.

With these ideas in mind I updated the maintainers notes.

scrubbbbs:
> "optional" for me means packages that can be removed without making the software unable to compile or run

opencv, raw, and tiff are also optional. There are really 4 types of dependencies here

- optional build/development dependencies (don't affect the binaries at all) (gtest)
- optional run-time software/plugins (don't affect binaries, but we'll use them if we find them) (qt-imageformats-plugins, kimageformats)
- optional linked libraries (get linked to binaries)  (opencv, tiff,  raw,  quazip)
- alternate installation packages (e.g.  opencv-core,opencv-imgproc instead of the full fat opencv metapackage)

Instead of listing these for each OS, I think we should just have a section that explains this a bit and we'll list the full package list with the understanding some of them can be omitted. I'm going to draft that now, let me know what you think.

README.md
```diff
@@ -48,63 +59,146 @@ python scripts/make.py "qt/bin" --lib-path "nomacs/3rd-party/build"
...
+| LibRAW  |  No      | OpenCV  | `ENABLE_RAW`       | Enables reading RAW images
+| LibTiff |  No      | OpenCV  | `ENABLE_TIFF`      | Enable reading multi-page TIFF
+| Quazip**|  No      |         | `ENABLE_QUAZIP`    | Enable reading from zip files
+| OpenCV  |  No      |         | `ENABLE_OPENCV`    | RAW, TIFF, Adjustments, High-quality thumbnails, DRIF files, histogram, mosaic
+| nomacs/plugins| No | OpenCV  | `ENABLE_PLUGINS`   | Paint on image, Composite, Affine Transform, Fake Miniatures, Page Extraction
+| googletest    | No |         | `ENABLE_TESTING`   | Enables `make check` target for unit testing
+| QImageFormats | No |         | n/a                | Enables reading ICNS, MNG, TGA, TIFF, WBMP, WEBP
+| KImageFormats | No |         | n/a                | Enables reading AVIF, HEIF/HEIC, JXL, EXR, EPS and [more](https://github.com/KDE/kimageformats)
+
+(**) Quazip is not enabled by default as of nomacs 3.19.1, this may change in a future update.
```
leejuyuu, line 88: I think this is a breaking change and we should not do this unless necessary.

README.md
```diff
@@ -48,63 +59,146 @@ python scripts/make.py "qt/bin" --lib-path "nomacs/3rd-party/build"
...
-  sudo apt-get install debhelper cdbs qt5-qmake qttools5-dev-tools qt5-default qttools5-dev libqt5svg5-dev qt5-image-formats-plugins libexiv2-dev libraw-dev libopencv-dev cmake libtiff-dev libquazip5-dev libwebp-dev git build-essential lcov libzip-dev
-  ```
+This is the list of configurable packages and what they provide in nomacs. If you want to be sure a feature is included/excluded set the cmake option for that feature. By default, all features are enabled if the dependencies are met**.
 
-Clone the nomacs repository from GitHub:
+| Package | Required | Depends | cmake | Description
+| -       | -        | -       | -     | -
+| Qt      |  Yes     |         | `QT_VERSION_MAJOR` | Qt components: Core, Concurrent, Network, PrintSupport, SVG, Widgets, Core5Compat
+| LibRAW  |  No      | OpenCV  | `ENABLE_RAW`       | Enables reading RAW images
+| LibTiff |  No      | OpenCV  | `ENABLE_TIFF`      | Enable reading multi-page TIFF
```
leejuyuu, line 80: Do these two libraries depend on OpenCV?

leejuyuu:
> Instead of listing these for each OS, I think we should just have a section that explains this a bit and we'll list the full package list with the understanding some of them can be omitted. 

I'm okay with the current version. Thanks!

README.md
scrubbbbs: LibRAW definitely does. LibTiff I'm not sure, it was not clear from scanning the code for `WITH_LIBTIFF`, however it is a cmake dependency. Maybe that is outdated.

leejuyuu: Could you confirm? The libraw packages in Ubuntu and Arch also do not list opencv as a dependency.

https://www.libraw.org/docs/Install-LibRaw.html
image

README.md
scrubbbbs: I assume you mean making quazip default enabled. I'm not advocating that we do this anytime soon,  but  I do not think it would break much and the trend seems to be always enabling it.

d7e555f878597fdb was the last to disable quazip (it has gone back and forth a few times) , but it is not clear to me why  (@novomesk  might remember why). Since then we have updated 3rdparty/quazip.

- It is a somewhat unique feature, I believe one significant reason users select nomacs over other viewers
- Quazip1 seems to be present in all distros now
- Most maintainers  seem to be setting ENABLE_QUAZIP=ON already (they were on 3.17 too). 
- Quazip < 0.7.6 had a CVE which we resolved by updating the submodule
- We fixed the build so  USE_SYSTEM_QUAZIP works on everything again

README.md
scrubbbbs: To clarify, the dependency is in nomacs itself, not the library. In otherwords, OpenCV is integral to how we use LibRAW.

As for tiff,

`cmake -DENABLE_OPENCV=OFF -DENABLE_RAW=OFF -DENABLE_TIFF=ON  ../ImageLounge/`

Does not work; however it might be made to work easily. I did not check functionality, but simply removing the check from Unix.cmake and the includes in DkBasicLoader.cpp allowed it to compile. Now, for the page extraction plugin, which I believe takes advantage of multi-page TIFF support, that might be another story.

leejuyuu: >To clarify, the dependency is in nomacs itself, not the library. In otherwords, OpenCV is integral to how we use LibRAW.

The current wording did not look like that. I would probably invert the table if that is the intent.

```md
#### Build options and their dependencies

The following Qt components are necessary: Core, Concurrent, Network, PrintSupport, SVG, Widgets, Core5Compat.

There are other optional features that can be enabled during build:

| Option             | Depends on     | Requires        | Description
| ------------------ | -------------- | --------------- | -
| `ENABLE_RAW`       | LibRAW         | `ENABLE_OPENCV` | Enables reading RAW images
| `ENABLE_TIFF`      | LibTiff        | `ENABLE_OPENCV` | Enable reading multi-page TIFF
| `ENABLE_QUAZIP`    | Quazip**       |                 | Enable reading from zip files
| `ENABLE_OPENCV`    | OpenCV         |                 | RAW, TIFF, Adjustments, High-quality thumbnails, DRIF files, histogram, mosaic
| `ENABLE_PLUGINS`   | nomacs/plugins | `ENABLE_OPENCV` | Paint on image, Composite, Affine Transform, Fake Miniatures, Page Extraction
| `ENABLE_TESTING`   | GoogleTest     |                 | Enables `make check` target for unit testing

#### Runtime dependencies

- Corresponding libraries for the enabled options.
- QImageFormats: Enables reading ICNS, MNG, TGA, TIFF, WBMP, WEBP
- KImageFormats: Enables reading AVIF, HEIF/HEIC, JXL, EXR, EPS and [more](https://github.com/KDE/kimageformats)

```

Issue #718, opened 2021-07-06
Rotate icons should not be the same as the undo / redo icons


RatatataBoom:
Stepts to reproduce:
1. Open the program.
2. Open file (any type).
3. Click the Adjustment menu.
3. Сompare Undo / Redo icons in the Edit tool bar with icons in the Adjustments panel below the thumbnail and with the Rotate icon at the top of the Adjustments panel

Expected result: 
Icons are rendered differently according to their function.

Actual result:
The icons are identical

Attachments: 
https://drive.google.com/file/d/15TebJKE8ki5fwRO8oSi1msU9ZBccClKZ/view?usp=sharing

Environment:

Win10 (19042.1052)
Nomacs ver: 3.16.1709[x64]

PR #1229, opened 2024-12-24
Undo redo toolbar icons

leejuyuu:
Add undo redo icons and toolbar items.

Fixes https://github.com/nomacs/nomacs/issues/718.

### DIFF
--- a/ImageLounge/license/COPYING.BREEZE-ICONS
+++ b/ImageLounge/license/COPYING.BREEZE-ICONS
@@ -1,0 +1,213 @@
+Applies to 
+  - src/img/edit-undo.svg
+  - src/img/edit-redo.svg
+
+The Breeze Icon Theme in icons/
+
+    Copyright (C) 2014 Uri Herrera <uri_herrera@nitrux.in> and others
+
+    This library is free software; you can redistribute it and/or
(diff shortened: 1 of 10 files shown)


scrubbbbs:
I'm OK with removing the buttons, however ctrl-z does not work after performing an action until I click in the viewport. 

There should be something we can do to allow unused keyboard events to propagate from docked panels into to the main viewport. I think this doesn't work by default because they are at the same widget "level" as the central widget rather than being children (in the QMainWindow paradigm).

leejuyuu:
> I'm OK with removing the buttons, however ctrl-z does not work after performing an action until I click in the viewport.
> 
> There should be something we can do to allow unused keyboard events to propagate from docked panels into to the main viewport. I think this doesn't work by default because they are at the same widget "level" as the central widget rather than being children (in the QMainWindow paradigm).

The Ctrl+Z shortcut does not work as well before this branch.

We would need to forward the events explicitly or add some signals and slots to pass those events around. I have not grasped the widget tree enough to do this now, so I created https://github.com/nomacs/nomacs/issues/1234 to track this.

leejuyuu:
By the way, please don't squash merge if possible. I use git blame quite often, and it does not work very well when there are big combined diff.

Issue #1232, opened 2024-12-30
Crash when rotating 8-bit image


leejuyuu:
> I am getting either a segfault or exception thrown from cv::Mat() when constructing the image with 8-bit images. It makes sense since CV_8UC4 is 4 bytes per pixel, and lower bit depths are not being converted to 32-bit.
> 
> 
> 
> Would it  be possible to support common formats here (8,24-bit) without changing too much? If OpenCV supports 64-bit images maybe that could be added too.
> 
>  

 _Originally posted by @scrubbbbs in [843a889](https://github.com/nomacs/nomacs/commit/843a8898d9aca7a4535f3f387a34901c6bc2e9a9#r150796524)_

leejuyuu:
> CV_8UC4 means unsigned 8-bit int for each pixel and 4 channels, so the data is 32-bit. Here I used that because the color should not be relevant here, only that each pixel is 32-bit is important. This works because QImage uses at least 32-bit aligned data and we convert 64-bit images to 32-bit.
> 
> 
> 
> Could you attach the image that caused the crash? I can open an 8-bit image without issue. 

 _Originally posted by @leejuyuu in [843a889](https://github.com/nomacs/nomacs/commit/843a8898d9aca7a4535f3f387a34901c6bc2e9a9#r150798242)_

leejuyuu:
> An 8-bit image (256-color) like GIF for example has *one* 8-bit channel, hence the issue. So for your code to work as-is all qimage pixel formats must be promoted/demoted to 32-bit rgba-8888
> 
> 
> 
> This file has an exif rotation but nomacs will ignore it. However, if you use the rotate function in the toolbar, it will crash.
> 
> 
> 
> [Image](https://github.com/user-attachments/assets/aa88148f-153e-4391-b56e-fa2515017078)
> 
>  

 _Originally posted by @scrubbbbs in [843a889](https://github.com/nomacs/nomacs/commit/843a8898d9aca7a4535f3f387a34901c6bc2e9a9#r150799045)_

leejuyuu:
Confirmed at 26c81093, Arch Linux.

```
terminate called after throwing an instance of 'cv::Exception'
  what():  OpenCV(4.10.0) /usr/src/debug/opencv/opencv/modules/core/src/matrix.cpp:434: error: (-215:Assertion failed) _step >= minstep in function 'Mat'


Thread 1 "nomacs" received signal SIGABRT, Aborted.
[...]

PR #1233, opened 2024-12-31
fix: fix crash when rotating 8-bit image

leejuyuu:
I misunderstood the "alignment" in the QImage docs. Here it means that the start of each row is at least 32-bit aligned, while the elements are not necessarily so. This means we cannot treat 8-bit images as QRgb arrays, and these images might have paddings at the end of each row.

Alter the image rotation function to handle different bit depths and the paddings at the end of the rows.

Fixes https://github.com/nomacs/nomacs/issues/1232

### DIFF
--- a/ImageLounge/src/DkCore/DkImageStorage.cpp
+++ b/ImageLounge/src/DkCore/DkImageStorage.cpp
@@ -39,6 +39,7 @@
 
 #pragma warning(push, 0) // no warnings from includes - begin
 #include <QBitmap>
+#include <QColorSpace>
 #include <QDebug>
 #include <QPainter>
 #include <QPixmap>
@@ -294,6 +295,8 @@ template<typename T>
 QImage transposeImage(const QImage &imgIn)
 {
     QImage imgOut = QImage(imgIn.size().transposed(), imgIn.format());
+    imgOut.setColorTable(imgIn.colorTable());
+    imgOut.setColorSpace(imgIn.colorSpace());
 
     const int h = imgIn.height();
     const int w = imgIn.width();
@@ -320,6 +323,8 @@ QImage transposeImage24(const QImage &imgIn)
     // but somehow several times slower than the above transposeImage.
 
     QImage imgOut = QImage(imgIn.size().transposed(), imgIn.format());
+    imgOut.setColorTable(imgIn.colorTable());
+    imgOut.setColorSpace(imgIn.colorSpace());
 
     const int h = imgIn.height();
     const int w = imgIn.width();
@@ -342,6 +347,8 @@ QImage rotateImageCVMat(const QImage &imgIn, cv::RotateFlags rot, int type)
 {
     QSize size = rot == cv::ROTATE_180 ? imgIn.size() : imgIn.size().transposed();
     QImage imgOut = QImage(size, imgIn.format());
+    imgOut.setColorTable(imgIn.colorTable());
+    imgOut.setColorSpace(imgIn.colorSpace());
     const cv::Mat matIn = cv::Mat(imgIn.height(), imgIn.width(), type, (uchar *)imgIn.constBits(), imgIn.bytesPerLine());
     cv::Mat matOut = cv::Mat(imgOut.height(), imgOut.width(), type, imgOut.bits(), imgOut.bytesPerLine());
 
--- a/ImageLounge/src/DkCore/DkImageStorage.cpp
+++ b/ImageLounge/src/DkCore/DkImageStorage.cpp
@@ -290,29 +290,65 @@ QImage rotateImage(const QImage &img, double angle)
     return imgR;
 }
 
-QImage transposeImage(const QImage &img)
+template<typename T>
+QImage transposeImage(const QImage &imgIn)
 {
-    QImage imgIn = img;
-    if (img.depth() > 32) {
-        // Ensure the data is 32-bit aligned
-        imgIn = img.convertToFormat(QImage::Format_ARGB32);
-    }
-
     QImage imgOut = QImage(imgIn.size().transposed(), imgIn.format());
-
-    const QRgb *dataIn = reinterpret_cast<const QRgb *>(imgIn.constBits());
-    QRgb *dataOut = reinterpret_cast<QRgb *>(imgOut.bits());
 
     const int h = imgIn.height();
     const int w = imgIn.width();
+    const uchar *ptrIn = imgIn.constBits();
+    const int outBPL = imgOut.bytesPerLine();
+    const int inBPL = imgIn.bytesPerLine();
     for (int i = 0; i < h; i++) {
+        const T *lineIn = reinterpret_cast<const T *>(ptrIn);
+        uchar *ptrOut = const_cast<uchar *>(imgOut.constBits() + i * sizeof(T));
         for (int j = 0; j < w; j++) {
-            dataOut[j * h + i] = dataIn[i * w + j];
+            T *out = reinterpret_cast<T *>(ptrOut);
+            *out = lineIn[j];
+            ptrOut += outBPL;
+        }
+        ptrIn += inBPL;
+    }
+
+    return imgOut;
+}
+
+QImage transposeImage24(const QImage &imgIn)
+{
+    // NOTE: this implementation is more general (any depth that is multiple of 8 can be handled),
+    // but somehow several times slower than the above transposeImage.
+
+    QImage imgOut = QImage(imgIn.size().transposed(), imgIn.format());
+
+    const int h = imgIn.height();
+    const int w = imgIn.width();
+    const int outBPL = imgOut.bytesPerLine();
+    uchar *outBits = imgOut.bits();
+    const int charPerPixel = imgIn.depth() / CHAR_BIT;
+    for (int i = 0; i < h; i++) {
+        const uchar *ptrIn = imgIn.constScanLine(i);
+        uchar *ptrOut = outBits + i * charPerPixel;
+        for (int j = 0; j < w; j++) {
+            memcpy(ptrOut + outBPL * j, ptrIn + charPerPixel * j, charPerPixel);
         }
     }
 
     return imgOut;
 }
+
+#ifdef WITH_OPENCV
+QImage rotateImageCVMat(const QImage &imgIn, cv::RotateFlags rot, int type)
+{
+    QSize size = rot == cv::ROTATE_180 ? imgIn.size() : imgIn.size().transposed();
+    QImage imgOut = QImage(size, imgIn.format());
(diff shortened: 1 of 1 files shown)


ImageLounge/src/DkCore/DkImageStorage.cpp
```diff
@@ -327,41 +340,75 @@ QImage rotateImageFast(const QImage &img, double angle)
...
-        return imgOut;
+        // The pixel depth matters, but the color interpretation does not.
+        switch (img.depth()) {
+        case 1:
+            return rotateImageCVMat(img.convertToFormat(QImage::Format_Indexed8), rot, CV_8U);
+        case 8:
+            return rotateImageCVMat(img, rot, CV_8U);
+        case 16:
+            return rotateImageCVMat(img, rot, CV_16U);
+        case 32:
```
scrubbbbs, line 362: I think this is only valid if the channels are 8-bit, there are a few Qt image formats have <8 bits per channel (e.g. RGB555). ~As written, the 16U case is only valid for 16-bit grayscale images.~ I think this might be OK as 16U ccovers  the entire pixel

You are also missing rgb/bgr-888  (24-bit)  which is fairly common unless that can't work for some reason.

Likewise, not all 32-bit format are 8bits per channel), there are a 10-bit variants for HDR.

Ok now I am thinking if you rotate the entire pixel instead of per-component this would cover most cases (except rgb-888). So CV_8UC4 should be CV_32U

ImageLounge/src/DkCore/DkImageStorage.cpp
```diff
@@ -327,41 +340,75 @@ QImage rotateImageFast(const QImage &img, double angle)
...
 
+        QImage imgIn = img;
+
         if (angle == 90) {
-            return transposeImage(imgIn).mirrored(true, false);
+            switch (img.depth()) {
+            case 1:
+                imgIn = img.convertToFormat(QImage::Format_Indexed8);
+                return transposeImage<uint8_t>(imgIn).mirrored(true, false);
+            case 8:
```
scrubbbbs, line 381: A fallthrough case here is more clear since they both call the same function

ImageLounge/src/DkCore/DkImageStorage.cpp
```diff
@@ -327,41 +340,75 @@ QImage rotateImageFast(const QImage &img, double angle)
...
+            default:
+                imgIn = img.convertToFormat(QImage::Format_ARGB32);
+                return transposeImage<uint32_t>(imgIn).mirrored(true, false);
+            }
         }
 
         if (angle == 270) {
-            return transposeImage(imgIn.mirrored(true, false));
+            switch (img.depth()) {
+            case 1:
```
scrubbbbs, line 397: So this is the same as the 90 case except you mirror it first...why do you need this block, just do the one call to mirror() and use the rotate 90 block.

scrubbbbs (changes requested):

ImageLounge/src/DkCore/DkImageStorage.cpp
leejuyuu: Sorry, I would like to avoid using fallthrough. It could become a foot gun if we try to add or remove branches. I also think returning explicitly adds clearity.

ImageLounge/src/DkCore/DkImageStorage.cpp
leejuyuu: I guess you want to avoid repetition, but that would partly merge the 90- and 270-degree branches, while all the conditionals remain. It makes the code less obvious. 

It currently looks a bit ugly, but it's clear that one block only does 90-degree things, and the other is only related to 270-degree. It's more reading effort for me to parse {conditional mirror} -> {shared transpose} -> {conditional mirror} into {mirror->transpose} and {transpose->mirror}.

ImageLounge/src/DkCore/DkImageStorage.cpp
leejuyuu: As transposing and flipping move pixel as a whole, the internal structure of the pixel does not matter. So using CV_8UC4 should have the same effect as CV_32U, because they are all 32-bit unit.

Here's the possible depths of QImage. Formats like RGB555 would have padding, see https://doc.qt.io/qt-6/qimage.html#Format-enum.
image

I missed the 24-bit one. Might take some time to do since there are no native 24-bit type...

scrubbbbs: So long as opencv isn't rotating channels separately I guess you don't  need CV_32U, but that would be more transparent.  If the rotation angle was anything other than 90/180/270 then it wouldn't opencv have to subdivide and resample the pixels to get a nice result?

ImageLounge/src/DkCore/DkImageStorage.cpp
scrubbbbs: ok

ImageLounge/src/DkCore/DkImageStorage.cpp
scrubbbbs: ok

scrubbbbs (approved):

ImageLounge/src/DkCore/DkImageStorage.cpp
leejuyuu: There's no CV_32U, so I used CV_32F instead. The 24-bit implementation has been added.

We are not using opencv for angles other than 90, 180, and 270. Could try that since opencv could be more efficient than QPainter.

scrubbbbs:
We still have some crashes. Specifically, "monob" (1-bit) and "pal8" (color-indexed)  variants in  https://github.com/nomacs/formats_testset/tree/main/generated/png

Note, I did not test the non-opencv codepath, however the version before the original patch seems to crash on additional formats as well.

leejuyuu:
The crash happens in a strange place.

<details>
<summary>Backtrace</summary>

```
Thread 19 "Thread (pooled)" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffe620a6c0 (LWP 16941)]
Downloading 102.39 K source file /usr/src/debug/qt6-base/qtbase/src/gui/painting/qpixellayout.cpp
fetchIndexedToARGB32PM<(QPixelLayout::BPP)3> (    
    buffer=0x7fffe61fc8d0, 
    src=0x7fff9061a1fc "", index=0, count=900, 
    clut=0x0)
    at /usr/src/debug/qt6-base/qtbase/src/gui/painting/qpixellayout.cpp:759
759	       buffer[i] = qPremultiply(clut->at(s));
(gdb) bt
#0  fetchIndexedToARGB32PM<(QPixelLayout::BPP)3>
    (buffer=0x7fffe61fc8d0, 
    src=0x7fff9061a1fc "", index=0, count=900, 
    clut=0x0)
    at /usr/src/debug/qt6-base/qtbase/src/gui/painting/qpixellayout.cpp:759
#1  0x00007ffff371efcc in fetchTransformedBilinear_simple_scale_helper<(TextureBlendType)4>(unsigned int*, unsigned int*, QTextureData const&, int&, int&, int, int) [clone .constprop.0] [clone .isra.0] (b=<optimized out>, 
    end=end@entry=0x7fffe6208878, image=..., 
    fx=@0x7fffe62009f8: 668, 
    fdx=fdx@entry=66873, fy=<optimized out>)
    at /usr/src/debug/qt6-base/qtbase/src/gui/painting/qdrawhelper.cpp:2098
#2  0x00007ffff36967d6 in fetchTransformedBilinear<(TextureBlendType)4, (QPixelLayout::BPP)0> (
    buffer=0x7fffe6207ab0, data=0x55555584e6f8, 
    y=<optimized out>, x=<optimized out>, 
    length=882)
    at /usr/src/debug/qt6-base/qtbase/src/gui/painting/qdrawhelper.cpp:2363
#3  0x00007ffff36a511c in BlendSrcGeneric::fetch
    (this=0x7fffe6205a90, x=42, 
    y=<optimized out>, len=882, fetchDest=false)
    at /usr/src/debug/qt6-base/qtbase/src/gui/painting/qdrawhelper.cpp:4088
#4  handleSpans<BlendSrcGeneric>(int, QT_FT_Span_ const*, QSpanData const*, Operator const&)::{lambda(int, int)#1}::operator()(int, int) const (
    __closure=0x7fffffffa1c0, 
    cStart=<optimized out>, cEnd=<optimized out>)
    at /usr/src/debug/qt6-base/qtbase/src/gui/painting/qdrawhelper.cpp:4040
#5  0x00007ffff36ac4c9 in handleSpans<BlendSrcGeneric>(int, QT_FT_Span_ const*, QSpanData const*, Operator const&)::{lambda()#1}::operator()() const
    (__closure=0x555555c4b838)
--Type <RET> for more, q to quit, c to continue without paging--
ting/qdrawhelper.cpp:4062
#6  QRunnable::QGenericRunnable::Helper<handleSpans<BlendSrcGeneric>(int, QT_FT_Span_ const*, QSpanData const*, Operator const&)::{lambda()#1}>::impl(QRunnable::QGenericRunnable::HelperBase::Op, QRunnable::QGenericRunnable::HelperBase*, void*) (op=<optimized out>, that=0x555555c4b830, arg=<optimized out>)
    at /usr/src/debug/qt6-base/qtbase/src/corelib/thread/qrunnable.h:73
#7  0x00007ffff2cdf699 in QRunnable::QGenericRunnable::HelperBase::run (this=<optimized out>)
    at /usr/src/debug/qt6-base/qtbase/src/corelib/thread/qrunnable.h:61
#8  QRunnable::QGenericRunnable::run (this=0x55555572bb00)
    at /usr/src/debug/qt6-base/qtbase/src/corelib/thread/qrunnable.cpp:32
#9  QThreadPoolThread::run (this=0x55555589d500)
    at /usr/src/debug/qt6-base/qtbase/src/corelib/thread/qthreadpool.cpp:70
#10 0x00007ffff2cd90cd in operator() (__closure=<optimized out>)
    at /usr/src/debug/qt6-base/qtbase/src/corelib/thread/qthread_unix.cpp:354
#11 (anonymous namespace)::terminate_on_exception<QThreadPrivate::start(void*)::<lambda()> > (t=...)
    at /usr/src/debug/qt6-base/qtbase/src/corelib/thread/qthread_unix.cpp:290
#12 QThreadPrivate::start (arg=0x55555589d500)
    at /usr/src/debug/qt6-base/qtbase/src/corelib/thread/qthread_unix.cpp:318
#13 0x00007ffff28a339d in start_thread (arg=<optimized out>) at pthread_create.c:447
#14 0x00007ffff292849c in __GI___clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
```
</details>

leejuyuu:
The monob and pal8 crash should be fixed. It was due to not copying the indexed colors.

leejuyuu:
Sorry, I used a wrong Git email. Mark as draft for now.

leejuyuu:
I fixed the commit author. Thanks!

PR #1235, opened 2025-01-01
Fix bugs in Edit Image panel

### LINKED ISSUE #350: Thumbnail Preview of Image Adjustment - Rotation problem
**Describe the bug**

Moving the rotation slider, the thumbnail preview orientation eventually goes crazy, it doesn't represent the value indicated by the slider.

**To Reproduce**
1. Go to 'Right Click -> Edit -> Image Adjustment -> Rotate'
2. Click on 'Angle Slider. Keep the mouse button clicked and try different orientation values'
3. The thumbnail orientation goes crazy!

**Expected behavior**

Obviously, it would be nice if the thumbnail orientation would be the one indicated by the slider and applied to the actual image.

**Screenshots**

 - OS: Kubuntu 19.04
 - Version: 3.12

### LINKED ISSUE #681: Wrong growth of resize function
Type (Severity): S3 Major
Priority: P2 Medium
Assigned to: Author “https://nomacs.org”
Title: increasing the image resolution when switching resize
Version: 3.16 (current)
Environment: PC
Steps:
1.	open the program
2.	open the image
3.	open adjustments
4.	choose resize
5.	make 2 scale
6.	choose threshold
7.	choose resize

Actual result: resolution of image has been increased by 4 times
Expected Result: resolution of image has been increased by 2 times

### LINKED ISSUE #691: Panel for rotation layout jumps
**Describe the bug**
Preview image rotates and panel adapts to image angle in height, part of side bar sections are hidden  

**To Reproduce**
1.	Open program Nomacs
2.	Open any image
3.	Click Rotate button on side panel
4.	Drag slider and move from left to right

**Expected behavior**
Panel should be fixed in one position to avoid panel “jumping”

**Screenshots**
https://prnt.sc/14dwf55 
https://prnt.sc/14dwo1p 

 - OS: windows 7
 - Version Nomacs 3.16.1709

### LINKED ISSUE #723: Edit field. Fails to enter a value with one numeral
Число 1 раз
**Describe the bug**
the Edit field does not accept a value as a number. input only 1 numeral.

**To Reproduce**
1. Go to 'adjustment' - "Blur".
2. Click on 'edit field'
3. enter value

**Expected behavior**
Editing the previous value when entering a new number

**Desktop:**
 - OS: [Windows 10 x64]
 - Version [3.16.1709]

### LINKED ISSUE #725: Window jumps. The section window "resize" jumps if change position of the slider.
**Describe the bug**
When a slider changes positions for resizing image, the section window of the "resize" starts to jump

**To Reproduce**
1. Go to 'adjustment' - "resize".
2. Мove the slider to the minimum value

**Expected behavior**
The image smoothly resizes in accordance with the changed values

Screenshot_7

OS: [Windows 10 x64]
Version [3.16.1709]

### LINKED ISSUE #748: Image Rotates each time when the option is selected
**Steps to reproduce**
1. Open image
2. Go to Adjustments -> Image Adjustments 
3. In Rotate options set the angle (no 0)
4. Go to another option
5. Return to Rotate

**Actual result**
Image will be rotated each time we select the Rotation option

**Expected result**
Image should keep its rotation angle

**System**
Win 10 x64
nomacs ver 3.16.1709 [x64]

### LINKED ISSUE #751: Image Adjustments issue when typing with keyboard
**Describe the bug**
If the value of adjustments is typed in the Edit Image panel, it stops typing after the first digit 

**To Reproduce**
1. Open app
2. Open image
3. Go to Edit Image Panel (D) (at the right) -> Rotate
4. Type any 2 digit value in  the 'Angle' box
5. See the error

**Expected behavior**
There must be the opportunity to type several digits values

Win 10 x64
Nomacs ver 3.16.1709 [x64]

leejuyuu:
Fixes #350
Fixes #681
Fixes #691
Fixes #723
Fixes #725
Fixes #748
Fixes #751

### DIFF
--- a/ImageLounge/src/DkCore/DkImageContainer.cpp
+++ b/ImageLounge/src/DkCore/DkImageContainer.cpp
@@ -357,6 +357,7 @@ QImage DkImageContainer::imageScaledToWidth(int width)
 
 void DkImageContainer::setImage(const QImage &img, const QString &editName)
 {
+    scaledImages.clear();
     getLoader()->setEditImage(img, editName);
     mEdited = true;
 }
--- a/ImageLounge/src/DkCore/DkImageContainer.cpp
+++ b/ImageLounge/src/DkCore/DkImageContainer.cpp
@@ -336,25 +336,6 @@ QImage DkImageContainer::imageScaledToHeight(int height)
     return sImg;
 }
 
-QImage DkImageContainer::imageScaledToWidth(int width)
-{
-    // check cache first
-    for (const QImage &img : scaledImages) {
-        if (img.width() == width)
-            return img;
-    }
-
-    // cache it
-    QImage sImg = image().scaledToWidth(width, Qt::SmoothTransformation);
-    scaledImages << sImg;
-
-    // clean up
-    if (scaledImages.size() > 10)
-        scaledImages.pop_front();
-
-    return sImg;
-}
-
 void DkImageContainer::setImage(const QImage &img, const QString &editName)
 {
     scaledImages.clear();
--- a/ImageLounge/src/DkCore/DkImageContainer.h
+++ b/ImageLounge/src/DkCore/DkImageContainer.h
@@ -74,7 +74,6 @@ public:
     QImage image();
     QImage pixmap();
     QImage imageScaledToHeight(int height);
-    QImage imageScaledToWidth(int width);
 
     bool hasImage() const;
     bool hasSvg() const;
--- a/ImageLounge/src/DkGui/DkManipulatorWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkManipulatorWidgets.cpp
@@ -166,17 +166,6 @@ void DkManipulatorWidget::selectManipulator()
     QSharedPointer<DkBaseManipulator> mpl = am.manipulatorManager().manipulator(action);
     QSharedPointer<DkBaseManipulatorExt> mplExt = qSharedPointerDynamicCast<DkBaseManipulatorExt>(mpl);
 
-    // compute preview
-    if (mpl && mImgC) {
-        DkTimer dt;
-        QImage img = mpl->apply(mImgC->imageScaledToWidth(qMin(mPreview->width(), mMaxPreview)));
-        img = scaledPreview(img);
-
-        if (!img.isNull())
-            mPreview->setPixmap(QPixmap::fromImage(img));
-        qDebug() << "preview computed in " << dt;
-    }
-
     for (QWidget *w : mWidgets)
         w->hide();
 
--- a/ImageLounge/src/DkGui/DkManipulatorWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkManipulatorWidgets.cpp
@@ -172,17 +172,26 @@ void DkManipulatorWidget::selectManipulator()
 
 void DkManipulatorWidget::selectManipulatorInner(QSharedPointer<DkBaseManipulatorExt> mplExt)
 {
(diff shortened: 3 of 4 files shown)


scrubbbbs:
I get that the update of the preview and viewport are now serialized, but that would seem to defeat the purpose of the preview feature. I suppose we could debate if the preview is even needed anymore.

Other issue introduced in this patch,  when first opening the adjustments panel (if it was hidden) the preview is tiny:
image

Unrelated to this patch, I also noticed that non-adjacent operations of the same type are merged as well, which is really strange choice. I guess when you really scrutinize it the manipulators there are a quite a few problems ;-) You probably have this in your list, but the most glaring issues to me:
- the weird merging of change history (shouldn't it only merge at the top of the stack?)
- no way to reset a manipulator to default values
- no way to apply the current values without changing a value first, then changing it back

scrubbbbs:
On the small-image issue; that seems to come from the undo/redo patch, and will show if there is no manipulator selected.

leejuyuu:
> I also noticed that non-adjacent operations of the same type are merged as well

Which operations were you using? For example, rotate 45 -> resize -> rotate 45 should work as expected.

> I suppose we could debate if the preview is even needed anymore.

I also feel that the small preview is redundant. However, I left it there for now to buy some time :)).

> no way to reset a manipulator to default values
> no way to apply the current values without changing a value first, then changing it back

In order to solve these, I think we need to separate the preview and the underlying image object (currently it's inside some DkImageContainerT). When a user interacts with the knobs on the panel, only the preview image should be updated and no history should be pushed to stack. Then, we add an "apply" button to the panels, which would apply the manipulation to the image and create a history item. 

This way, it's easier to implement a reset. Also, there would be no need to merge operations because we still have the initial state inside the image container. 

However, the current preview is too small. I am also not sure whether it is intuitive to use viewport for preview.
Also, the viewport is currently quite complicated.
One way out might be create a dialog for the manipulators like the resize image dialog, 
it might be easier to implement.

scrubbbbs:
> > I also noticed that non-adjacent operations of the same type are merged as well
> 
> Which operations were you using? For example, rotate 45 -> resize -> rotate 45 should work as expected.

Rotate->Exposure->Rotate->Resize. Drag the resize slider around for a while and it will suddenly merge the two rotates in the history panel.

scrubbbbs:
> > I suppose we could debate if the preview is even needed anymore.
> 
> I also feel that the small preview is redundant. However, I left it there for now to buy some time :)).

I would like to revive it, the response of sliders is very chunky on normal photos. Until we can optimize the filters (threaded/GLSL etc) to get the chunk out I think we need it.

scrubbbbs:
> Then, we add an "apply" button to the panels, which would apply the manipulation to the image and create a history item.

I think apply after mouse-up on a slider, or key-up event on a combo box, would be good compromise so an apply button would not be needed necessarily.

scrubbbbs:
> However, the current preview is too small. I am also not sure whether it is intuitive to use viewport for preview.
> Also, the viewport is currently quite complicated.
> One way out might be create a dialog for the manipulators like the resize image dialog,
> it might be easier to implement.

I'm wondering if we can somehow temporarily replace the original image with downsampled one, modify the transform in the viewport to match, then the manipulators can work as they do now (modifying history etc). This would minimize the amount of messing around in the viewport. When the "apply" is clicked, we would have to restore the original image and viewport transform, then render the full image by replaying the history.

Main problem is,  we would have to able to replay the history. Instead of pushing an image to history we would push an action that renders a new image and/or modifies metadata.

Issue #1174, opened 2024-10-07
Image orientation not respected for RW2 files


cdauth:
I am using Nomacs 3.19.1 on Arch Linux installed through AUR.

RW2 files that have been taken in portrait mode are displayed in landscape mode. Other image viewers are correctly displaying them in portrait mode. Here is an example file (sorry, I had to gzip it because GitHub does not allow RW2 files): 
[P1270615.RW2.gz](https://github.com/user-attachments/files/17282967/P1270615.RW2.gz)

novomesk:
How it should look like?

This is development build on Windows:
![image](https://github.com/user-attachments/assets/736fc9dd-81b4-4e62-b7bb-46a5f891b759)

cdauth:
That's what it should look like, this is how it looks for me:
![image](https://github.com/user-attachments/assets/1784a3a1-d00b-4431-8a18-fcd226060a63)

Also in the preview it is not rotated.

leejuyuu:
This is on 3.19.0 (eca2377d), Arch Linux, Qt 6.7.2.

![image](https://github.com/user-attachments/assets/25f21503-3d8c-4765-8a2d-2d20d53d1ef9)

cdauth:
This is the version that I’m running:
![image](https://github.com/user-attachments/assets/ef4a6dbc-c74b-4d2e-a311-ebccc0ca80f6)

Could it be about the settings?
![image](https://github.com/user-attachments/assets/d6948136-83d4-4798-af63-3fded68479ab)

I also tested it with JPEG files, there it seems to work correctly.

leejuyuu:
updated to the same version and config as yours, but mine still looks correct.

cdauth:
I don't know how to proceed. Does someone know more details about how this feature is implemented and how it can be that it behaves differently on different systems?

novomesk:
Maybe the result is different depending on whether kimg_raw plug-in is installed or not.

cdauth:
If I go to the nomacs plugin manager, I do not see any plugins. Or do I need to check somewhere else?

novomesk:
Somewhere on disk.

For example:

/usr/lib/qt6/plugins/imageformats/kimg_raw.so

cdauth:
Yes, I do have that installed.

novomesk:
Make a backup, remove kimg_raw.so, close and reopen nomacs, test again, restore kimg_raw.so from backup.

leejuyuu:
[...]

PR #1238, opened 2025-01-06
Refactor thumbnail and general image loader, support mirrored  orientation

scrubbbbs:
Policy Changes
- request exif thumb  only  returns exif thumb (previously, returns if image is small)
- always transform exif thumb using exif orientation
- load TGA before Qt TGA, workaround bug in Qt 5.15

Other changes
- use the QIR transform if present, and it overrides exif orientation (equivalent to old policy)

Fixes
- ICNS: load largest image
- wrong orientation on images and/or embedded thumbs on TIFF,PNG,JXL-wrapped JPG
- metadata dock showing non-exif thumbnails
- thumbnails also honor "ignore exif orientation" setting

Features
- support for mirrored orientation
- add metadata for debugging image loader (in Data section)
- add thumb info to thumbnail tooltips
- more detailed logging

Refactoring
- use QImageReader instead of QImage::load() to have more  control over image loading

New metadata added by the image loader
image

Thumbnail info added to hover
image


leejuyuu:
I haven't finished reading the code. One thing I noticed is that the data stored in `QImage::setText` gets saved when the image is written to file. As the information is already in the file and is only used by nomacs internally, this should not happen. Also, doing everything in string loses compile-time type safety. I suggest storing a struct somewhere (maybe inside `DkBasicLoader`?). Something like 

```cpp
struct LoadeImageInfo {
    QString mime_type{};
    bool has_metadata{false};
    bool has_sidecar{false};
    QString loader{};
    bool is_rotated{false};
    int load_time_ms{0};
    int depth{0};
    bool has_alpha{false};
    QString pixel_format{};
    std::optional<int> color_count{};
    std::optional<QString> color_space{};
};

```

ImageLounge/src/DkCore/DkBasicLoader.cpp
```diff
@@ -992,23 +1062,23 @@ void DkBasicLoader::setHistoryIndex(int idx)
...
 
-    QFile file(filePath);
-    file.open(QIODevice::ReadOnly);
+//     QFile file(filePath);
+//     file.open(QIODevice::ReadOnly);
 
-    ba = file.readAll();
-}
+//     ba = file.readAll();
+// }
```
leejuyuu, line 1164: Should this be removed?

ImageLounge/src/DkCore/DkMetaData.cpp
```diff
@@ -946,18 +962,19 @@ QStringList DkMetaDataT::getIptcValues() const
...
 {
-    QStringList qtKeysInit = cImg.textKeys();
-
-    for (QString cKey : qtKeysInit) {
-        if (!cKey.isEmpty() && cKey != "Raw profile type exif") {
-            QString val = cImg.text(cKey).size() < 5000 ? cImg.text(cKey) : QObject::tr("<data too large to display>");
+    const QStringList keys = img.textKeys();
+    for (const auto &key : keys) {
+        if (key.isEmpty())
+            continue; // || key == "Raw profile type exif") {
```
leejuyuu, line 982: Is removing the check on "Raw profile type exif" intended?

ImageLounge/src/DkCore/DkThumbs.h
```diff
@@ -138,29 +152,33 @@ class DllCoreExport DkThumbNail
...
-        do_not_force,
-        force_exif_thumb,
-        force_full_thumb,
-        save_thumb,
-        force_save_thumb,
-    };
+    // void setImgExists(bool exists)
+    // {
+    //     mImgExists = exists;
+    // };
```
leejuyuu, line 164: Maybe remove this?

leejuyuu (commented):
I think 08c45e3a2f488892769472d064b67323017c7af6 should not be titled as refactor because it adds features like thumbnail tooltip.

ImageLounge/src/DkCore/DkMetaData.cpp
```diff
@@ -299,9 +299,9 @@ QString DkMetaDataT::getDescription() const
 int DkMetaDataT::getOrientationDegree() const
 {
     if (mExifState != loaded && mExifState != dirty)
-        return 0;
+        return -2;
 
-    int orientation = 0;
+    int orientation = -2;
```
leejuyuu: I don't like returning error code in the place of the actual value, but that would require a signature change. Maybe at least make `-2` a named constant?

ImageLounge/src/DkCore/DkBasicLoader.cpp
```diff
@@ -460,7 +514,20 @@ bool DkBasicLoader::loadQt(const QString &filePath, QImage &img, QSharedPointer<
...
+
+    bool supportsTransform = qir.supportsOption(QImageIOHandler::ImageTransformation);
+
+    if (supportsTransform)
+        img.setText("QImageReader.Transform", "0x" + QString::number(qir.transformation(), 16));
+    else
+        img.setText("QImageReader.Transform", "unsupported");
+
+    static const QList<QByteArray> brokenPlugins{"heic", "heif", "avif", "jxl"};
+    if (!supportsTransform && brokenPlugins.contains(qir.format())) {
```
leejuyuu: Consider `excludedPlugins` instead?

ImageLounge/src/DkCore/DkBasicLoader.cpp
```diff
@@ -258,23 +258,28 @@ bool DkBasicLoader::loadGeneral(const QString &filePath, QSharedPointer<QByteArr
...
-    // "jpe" fixes #435 - thumbnail gets loaded in the RAW loader
-    static const QList<QByteArray> qtFormats = QImageReader::supportedImageFormats() << "jpe";
+    static const QList<QByteArray> qtFormats = QImageReader::supportedImageFormats();
     static const QList<QByteArray> drifFormats{"drif", "yuv", "raw"};
     static const QList<QByteArray>
         rawFormats{"nef", "nrw", "crw", "cr2", "cr3", "arw", "dng", "raw", "rw2", "mrw", "srw", "orf", "3fr", "x3f", "mos", "pef", "iiq", "raf"};
     static const QList<QByteArray> tiffFormats{"tif", "tiff"};
     static const QList<QByteArray> psdFormats{"psb", "psd"};
-    static const QList<QByteArray> jpegFormats{"jpg", "jpeg", "jpe"};
+    static const QList<QByteArray> jpegFormats{"jpg", "jpeg"};
```
leejuyuu: Why remove jpe from the lists? I can still open one though.

ImageLounge/src/DkCore/DkBasicLoader.cpp
```diff
@@ -258,23 +258,28 @@ bool DkBasicLoader::loadGeneral(const QString &filePath, QSharedPointer<QByteArr
...
-    // Do we prefer our RAW loader rather than KImageFormats?
+    // - RAW must precede TIFF (most RAW formats use TIFF container)
+    // - Qt should get first attempt as it is more actively maintained (in case of future CVEs etc)
+    // - Qt 5.15 TGA plugin cannot read some TGAs correctly, and won't report an error, so try ours first
+    // - tiff after Qt as that also has support
+    // - psd after Qt as KIF has support
+    // - roh/vec go last since they are rarely used, I can't source a test file for either
+    //
+    // We prefer our RAW loader over KIF
+    // - KIF always loads the jpeg preview, we have an option to disable it
```
leejuyuu, line 291:
```suggestion
    // - KImageFormats always loads the jpeg preview, we have an option to disable it
```

ImageLounge/src/DkCore/DkBasicLoader.h
```diff
@@ -278,36 +296,47 @@ class DllCoreExport DkBasicLoader : public QObject
...
+    // };
 
-    bool getTraining()
-    {
-        return mTraining;
-    };
+    // bool getTraining()
+    // {
+    //     return mTraining;
+    // };
```
leejuyuu: Should these be removed?

ImageLounge/src/DkCore/DkBasicLoader.h
```diff
@@ -278,36 +296,47 @@ class DllCoreExport DkBasicLoader : public QObject
...
     bool isMetaDataEdited();
 
-    QString getFile() const
-    {
-        return mFile;
-    };
+    // QString getFile() const
+    // {
+    //     return mFile;
+    // };
```
leejuyuu, line 334: Should this be removed?

ImageLounge/src/DkCore/DkBasicLoader.h
```diff
@@ -374,17 +403,8 @@ class DllCoreExport DkBasicLoader : public QObject
     bool loadDRIF(const QString &filePath, QImage &img, QSharedPointer<QByteArray> ba = QSharedPointer<QByteArray>()) const;
 
 #ifdef Q_OS_WIN
-    bool saveWindowsIcon(const QString &filePath, const QImage &img) const;
+    // bool saveWindowsIcon(const QString &filePath, const QImage &img) const;
```
leejuyuu, line 406: Should these be removed?

leejuyuu:
Confirmed that #1174 is fixed. Is this a bug of KImageFormats? Should we report to upstream?

ImageLounge/src/DkCore/DkBasicLoader.cpp
```diff
@@ -194,219 +194,293 @@ DkBasicLoader::DkBasicLoader(int mode)
...
 bool DkBasicLoader::loadGeneral(const QString &filePath, QSharedPointer<QByteArray> ba, bool loadMetaData, bool fast)
 {
     DkTimer dt;
-    bool imgLoaded = false;
 
     mFile = DkUtils::resolveSymLink(filePath);
-    QFileInfo fInfo(mFile); // resolved lnk
-    QString newSuffix = fInfo.suffix();
+    const QFileInfo fileInfo(mFile); // resolved lnk
+    const QByteArray suffix = fileInfo.suffix().toLower().toLatin1();
```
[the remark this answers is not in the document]
scrubbbbs: The suffix lists stored in  QImageReader are QByteArray, so we'd have to convert our QString anyways for "==". So this just to makes it clear that non-latin chars are not supported in file extensions.

leejuyuu:
It's great to log thumbnail load time, but showing it on the tooltip is a bit strange.

ImageLounge/src/DkCore/DkBasicLoader.h
scrubbbbs: Yes, they have no effect

ImageLounge/src/DkCore/DkBasicLoader.h
scrubbbbs: This is never used, so yes.

ImageLounge/src/DkCore/DkMetaData.cpp
scrubbbbs: Yes, I left it commented out since I cannot find a reason why it exists. Maybe @novomesk knows

ImageLounge/src/DkCore/DkBasicLoader.h
leejuyuu: I mean remove them instead of commenting out.

ImageLounge/src/DkCore/DkBasicLoader.cpp
scrubbbbs: It's dead code, so yeah. Trivial change if we want to drop QSharedPointer.

ImageLounge/src/DkCore/DkMetaData.cpp
leejuyuu: It might be safer to keep the current behavior for now. It seems to be https://exiftool.org/TagNames/PNG.html

scrubbbbs:
> data stored in `QImage::setText` gets saved when the image is written to file

Well, I did not notice that! I like the having the the data tied to the image, and it "just works" with metadata browser. What if I strip these keys from the file when writing?

scrubbbbs:
> It's great to log thumbnail load time, but showing it on the tooltip is a bit strange.

It seemed like it might be useful at the time, but knowing that the thumb is embedded is probably enough

leejuyuu:
> > data stored in `QImage::setText` gets saved when the image is written to file
> 
> Well, I did not notice that! I like the having the the data tied to the image, and it "just works" with metadata browser. What if I strip these keys from the file when writing?

It seems most of them are for debugging/logging? Does displaying them on the metadata browser benefit the users?

Stripping the keys might work, but it could be harder to maintain in the future and more prone to errors (adding new key, key name changes, etc.).

scrubbbbs:
> Is this a bug of KImageFormats? Should we report to upstream?

I'm not sure if it is a bug. It's like the heif/avif plugins @novomesk has been working on.

Since there is no way to communicate that they return transformed output, the only thing they can do is implement the `QImageIOHandler::ImageTransformation` option which might not be so simple as it is modeled on the older EXIF orientation standard.

novomesk:
Check `QImageReader::supportsOption` with `QImageIOHandler::ImageTransformation` parameter.

When it returns `false`, image doesn't need further rotation.

scrubbbbs:
> It seems most of them are for debugging/logging? Does displaying them on the metadata browser benefit the users?

Yes, to debug the image loader, but also to triage bugs in the image loader. When a format doesn't work this will tell us a lot more than we had before. For example, it is easy to tell when kimageformats is being used instead of libraw. I wouldn't be opposed to adding a setting to hide this, much like the log to file option.

> Stripping the keys might work, but it could be harder to maintain in the future and more prone to errors (adding new key, key name changes, etc.).

We could have all keys under one prefix to mitigate this, instead of how it is now.

scrubbbbs:
> Check `QImageReader::supportsOption` with `QImageIOHandler::ImageTransformation` parameter.
> 
> When it returns `false`, image doesn't need further rotation.

That isn't enough because we have formats like PNG, TIFF and WEBP that can have EXIF orientation, but don't support that option.

leejuyuu:
> When a format doesn't work this will tell us a lot more than we had before. For example, it is easy to tell when kimageformats is being used instead of libraw. I wouldn't be opposed to adding a setting to hide this, much like the log to file option.

If that is the case, I suggest we write that info to the log but don't display it on the metadata panel UI. Part of the data is not about the image, but about how nomacs loads the image, I think they might lead to questions or confusion.

scrubbbbs:
> > When a format doesn't work this will tell us a lot more than we had before. For example, it is easy to tell when kimageformats is being used instead of libraw. I wouldn't be opposed to adding a setting to hide this, much like the log to file option.
> 
> If that is the case, I suggest we write that info to the log but don't display it on the metadata panel UI. Part of the data is not about the image, but about how nomacs loads the image, I think they might lead to questions or confusion.

The info loses its value if it is detached from the image object, but I'll remove it for now. I guess this has to wait until a larger refactor to make a wrapper for QImage. I'll also have to remove some stuff from the tooltips as it was  pulling info from there.

leejuyuu:
Looks good, thanks!

scrubbbbs:
Thanks, I expect to merge this week as-is.

novomesk:
I built nomacs from this PR.

During testing I observed that when I change images with arrows on keyboard, the picture in the nomacs stays the same.

When I move mouse inside, then the picture redraws.

leejuyuu:
@novomesk I cannot reproduce on Arch, which platform did you test on?

scrubbbbs:
@novomesk sounds like a qt or qt/wayland bug. You can try passing  `-platform <platform>`, for wayland there is "wayland" and "wayland-egl" or "help"  to get a list of available platform plugins.

I found this [qt bug](https://bugreports.qt.io/browse/QTBUG-97985) which sounds similar

novomesk:
OK, I'll try later. My problem doesn't seem to be related with this PR.

So test the images from https://github.com/nomacs/formats_testset

And you can merge afterward when you feel comfortable.

novomesk:
@scrubbbbs I am experiencing the redraw issue since https://github.com/nomacs/nomacs/commit/f052604c022d26ae28528d7e84d5df026a87b27c (from Dec 16, 2024)

Previous commit from Dec 10, 2024 doesn't have the issue.

scrubbbbs:
@novomesk Thanks for the update. Please open an issue and continue discussion there. Since that patch did not touch display logic (there was one small change to DkViewPort which doesn't explain this) we have some work to do.

Can you provide some more details about your configuration, I can try to reproduce it in a VM or maybe real hardware. Since I am not familiar with Gentoo I will need as much detail about building your configuration as possible. If gentoo has a config file for building the system I could copy that would be great.

novomesk:
OK, mystery solved. My redraw issues happened when I had new compiled binaries, but old CSS (Dark-Theme.css  Light-Theme.css  System.css) files installed.

scrubbbbs:
> OK, mystery solved. My redraw issues happened when I had new compiled binaries, but old CSS (Dark-Theme.css Light-Theme.css System.css) files installed.

I recall spending several hours wondering why css changes were not working; I needed to do "make install". I thought this was fixed now by copying styles to the right place in the build directory, so at least when running from there they would be  found.

novomesk:
I'd like that we build 3.19.2 version in the near future.

Issue #807, opened 2021-09-09
Alpha channel value on status bar under pixel stuck at 255 


JulienMaille:
**Describe the bug**

**To Reproduce**
1. Open a png with transparency
2. Hover on a transparent pixel
3. Read the alpha value in the status bar
4. value is 255

**Expected behavior**
Value should be 0

**Screenshots**
![image](https://user-images.githubusercontent.com/182520/132731681-dbb984fe-32cd-449d-8321-b76d8c6c6178.png)
![image](https://user-images.githubusercontent.com/182520/132731697-a4c94e8e-76e9-4eeb-ae5c-1cddbec31abf.png)

**Desktop:**
 - OS: Windows 10
 - Version: 3.16.1709 x64

**Additional context**

leejuyuu:
Reproduced nomacs 3.19.1 (ba263741), Arch Linux Qt 6.8.1.

PR #1249, opened 2025-01-11
Main View: fix pixel sampler alpha channel

scrubbbbs:
- Fix alpha value locked at 255 in status bar
- Add alpha value to hex string

Fixes: #807

### DIFF
--- a/ImageLounge/src/DkGui/DkViewPort.cpp
+++ b/ImageLounge/src/DkGui/DkViewPort.cpp
@@ -1405,19 +1405,24 @@ void DkViewPort::getPixelInfo(const QPoint &pos)
         return;
 
     QPoint xy = mapToImage(pos);
-
-    if (xy.x() == -1 || xy.y() == -1)
+    if (xy.x() < 0)
         return;
 
-    QColor col = getImage().pixel(xy);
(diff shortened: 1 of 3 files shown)


ImageLounge/src/DkGui/DkViewPort.cpp
```diff
@@ -1405,19 +1405,24 @@ void DkViewPort::getPixelInfo(const QPoint &pos)
...
 
     QPoint xy = mapToImage(pos);
-
-    if (xy.x() == -1 || xy.y() == -1)
+    if (xy.x() < 0)
         return;
 
-    QColor col = getImage().pixel(xy);
+    // TODO: This is rgb now, but we could display native pixel values too, even CMYK in Qt 6.8.0
+    const QImage &img = getImage();
```
leejuyuu, line 1412:
```suggestion
    const QImage img = getImage();
```
QImage implements `QImage::QImage(QImage&& other)`, so the returned object gets moved. There should be no need to use reference here.
https://doc.qt.io/qt-6/qimage.html#QImage-1

leejuyuu (commented):
Confirmed #807 is fixed.

PR #1250, opened 2025-01-12
Add a button to the file explorer that scrolls to the currently open file

spaghetti22:
Hi,

beginner in C++ and Qt here; I added a button to the file explorer that scrolls to the currently open file, which improves the usability of nomacs for me (a frequent user of it). I hope I am doing this right, please give feedback if I am not.

Paul

### DIFF
--- a/ImageLounge/src/DkGui/DkWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkWidgets.cpp
@@ -408,8 +408,6 @@ DkBrowseExplorer::DkBrowseExplorer(const QString &title, QWidget *parent, Qt::Wi
 {
     createLayout();
     readSettings();
-
-    connect(mRootPathBrowseButton, &QPushButton::clicked, this, &DkBrowseExplorer::browseClicked);
 }
 
 DkBrowseExplorer::~DkBrowseExplorer()
@@ -423,6 +421,11 @@ void DkBrowseExplorer::browseClicked()
 
     if (root != "")
         setRootPath(root);
+}
+
+void DkBrowseExplorer::scrollToCurrentClicked()
+{
+    mFileTree->scrollTo(mFileTree->currentIndex());
 }
 
 void DkBrowseExplorer::setRootPath(const QString &root)
@@ -439,10 +442,19 @@ void DkBrowseExplorer::createLayout()
     QWidget *rootPathWidget = new QWidget(this);
     QHBoxLayout *rpLayout = new QHBoxLayout(rootPathWidget);
     mRootPathLabel = new DkElidedLabel(rootPathWidget, "");
-    mRootPathBrowseButton = new QPushButton(tr("Browse"));
+
+    auto browseButton = new QPushButton(tr("Browse"));
+    connect(browseButton, &QPushButton::clicked, this, &DkBrowseExplorer::browseClicked);
+
+    auto scrollButton = new QToolButton();
+    scrollButton->setIcon(DkImage::loadIcon(":/nomacs/img/scroll-to-current.svg"));
+    scrollButton->setToolTip(tr("Scroll to current file"));
+    connect(scrollButton, &QPushButton::clicked, this, &DkBrowseExplorer::scrollToCurrentClicked);
+
     rpLayout->setContentsMargins(4, 2, 2, 2);
     rpLayout->addWidget(mRootPathLabel, 1);
-    rpLayout->addWidget(mRootPathBrowseButton);
+    rpLayout->addWidget(browseButton);
+    rpLayout->addWidget(scrollButton);
 
     mLayout->insertWidget(0, rootPathWidget);
 }
--- a/ImageLounge/src/DkGui/DkWidgets.h
+++ b/ImageLounge/src/DkGui/DkWidgets.h
@@ -452,6 +452,7 @@ public:
 
 public slots:
     void browseClicked();
+    void scrollToCurrentClicked();
     void setRootPath(const QString &root);
 
 protected:
@@ -460,7 +461,6 @@ protected:
     void writeSettings() override;
 
     QString mRootPath;
-    QPushButton *mRootPathBrowseButton;
     DkElidedLabel *mRootPathLabel;
 };
 
--- a/ImageLounge/src/img/scroll-to-current.svg
+++ b/ImageLounge/src/img/scroll-to-current.svg
@@ -1,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 160 160" style="enable-background:new 0 0 160 160;" xml:space="preserve">
+<style type="text/css">
+	.st0{fill:#231F20;}
+	.st0-stroke{stroke:#231F20; stroke-width: 10px; fill: none;}
+</style>
+<circle class="st0-stroke" cx="80" cy="80" r="70" />
+<polygon id="inward-pointer" class="st0" points="10,60 10,100 50,80" />
+<use href="#inward-pointer" transform="rotate(90 80 80)" />
+<use href="#inward-pointer" transform="rotate(180 80 80)" />
+<use href="#inward-pointer" transform="rotate(270 80 80)" />
+</svg>
--- a/ImageLounge/src/nomacs.qrc
+++ b/ImageLounge/src/nomacs.qrc
@@ -30,6 +30,7 @@
         <file>img/dir.svg</file>
         <file>img/rotating-cursor.svg</file>
         <file>img/save.svg</file>
+        <file>img/scroll-to-current.svg</file>
         <file>img/export.svg</file>
         <file>img/open.svg</file>
         <file>img/thumbs-move.svg</file>
--- a/ImageLounge/translations/nomacs_de.ts
+++ b/ImageLounge/translations/nomacs_de.ts
@@ -3395,6 +3395,11 @@ This option is only useful if plugins save sidecar files - so be careful!</sourc
       <source>Browse</source>
       <translation>Durchsuchen</translation>
     </message>
+    <message>
+      <location filename="../../src/DkGui/DkWidgets.cpp" line="452"/>
+      <source>Scroll file explorer to currently open element</source>
+      <translation>Dateimanager zu aktuell offenem Element scrollen</translation>
+    </message>
   </context>
   <context>
     <name>nmc::DkCentralWidget</name>


ImageLounge/src/DkGui/DkCentralWidget.h
```diff
@@ -145,6 +145,7 @@ class DllCoreExport DkCentralWidget : public DkWidget
 
 signals:
     void imageUpdatedSignal(QSharedPointer<DkImageContainerT>) const;
+    void directoryLoadedSignal(QString) const;
```
leejuyuu, line 148:
```suggestion
    void directoryLoadedSignal(const QString&) const;
```

When passing objects into functions, we usually pass them by reference to avoid copying.

ImageLounge/src/DkGui/DkWidgets.h
```diff
@@ -436,6 +436,7 @@ public slots:
     QTreeView *mFileTree = nullptr;
     DkFileSystemModel *mFileModel = nullptr;
     DkSortFileProxyModel *mSortModel = nullptr;
+    QString *mCurrentFilePath = nullptr;
```
leejuyuu, line 439:
```suggestion
    QString mCurrentFilePath {};

```

QString handles the dynamically allocated memory internally, so there is no need to use a pointer here.

ImageLounge/src/DkGui/DkWidgets.cpp
```diff
@@ -531,6 +541,7 @@ void DkExplorer::setCurrentPath(const QString &filePath)
     if (QFileInfo(filePath).isDir())
         mFileTree->expand(mSortModel->mapFromSource(mFileModel->index(filePath)));
 
+    mCurrentFilePath = new QString(filePath);
```
leejuyuu, line 544:
```suggestion
    mCurrentFilePath = filePath;
```
Did you notice that this does not have its corresponding `delete`? This is why using `QString` directly is less error-prone.

ImageLounge/src/DkGui/DkWidgets.cpp
```diff
@@ -440,9 +446,13 @@ void DkBrowseExplorer::createLayout()
     QHBoxLayout *rpLayout = new QHBoxLayout(rootPathWidget);
     mRootPathLabel = new DkElidedLabel(rootPathWidget, "");
     mRootPathBrowseButton = new QPushButton(tr("Browse"));
+    mScrollToCurrentButton = new QToolButton();
+    mScrollToCurrentButton->setIcon(DkImage::loadIcon(":/nomacs/img/scroll-to-current.svg"));
+    mScrollToCurrentButton->setToolTip(tr("Scroll file explorer to currently open element"));
```
leejuyuu, line 451:
```suggestion
    mScrollToCurrentButton->setToolTip(tr("Scroll to current file"));

```

ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -988,6 +988,7 @@ void DkNoMacs::showExplorer(bool show, bool saveSettings)
         });
         connect(mExplorer, &DkExplorer::openDir, getTabWidget(), &DkCentralWidget::loadDirToTab);
         connect(getTabWidget(), &DkCentralWidget::imageUpdatedSignal, mExplorer, &DkExplorer::setCurrentImage);
+        connect(getTabWidget(), &DkCentralWidget::directoryLoadedSignal, mExplorer, &DkExplorer::setCurrentPath);
```
leejuyuu, line 991: Why did you add this signal connection? The current path of the file explorer should already be handled somewhere else.

ImageLounge/src/DkGui/DkCentralWidget.cpp
```diff
@@ -842,6 +842,7 @@ void DkCentralWidget::showThumbView(bool show)
 
             connect(tw, &DkThumbScrollWidget::updateDirSignal, tabInfo->getImageLoader().data(), &DkImageLoader::loadDirRecursive, Qt::UniqueConnection);
             connect(tw, &DkThumbScrollWidget::filterChangedSignal, tabInfo->getImageLoader().data(), &DkImageLoader::setFolderFilter, Qt::UniqueConnection);
+            emit directoryLoadedSignal(tabInfo->getImageLoader().data()->getDirPath());
```
leejuyuu, line 845: Emitting a signal named `directoryLoadedSignal` when showing the thumbnail preview panel looks weird.

scrubbbbs:
> added a button to the file explorer that scrolls to the currently open file

If I understand correctly, you want to scroll back to the open file without changing the open file, sort of like the "Back" button, but keep the same image loaded?

Because changing the open file *should* always do this, but I thought I'd ask since I just found one case (history panel) where it doesn't -- that's a bug.

spaghetti22:
Hi,

thanks for your feedback, I will look into all of it in the next days.

As for your question what my goal was: yes, the idea is that it should scroll to the currently open file without changing the open file. (I got the idea from IntelliJ, which has a very similar button.) You are right that it is now possible to do this by changing to the next file and then back, but this is somewhat inconvenient and inefficient in my workflow. I often scroll very far away from the currently open file in the file explorer; the button gives the user a quick way to scroll the file explorer back to what is currently open.

scrubbbbs:
We merged some changes to your branch, you will need `git pull` before you can push additional changes

I don't know  if I like the placement of the button, makes the two buttons look off too me. Maybe if the browse button also had an icon?

I'd like to remove the dependency on other components. Can you use the file manager's signal to set the path? Something like:
```
    connect(mFileTree->selectionModel(), &QItemSelectionModel::selectionChanged ...) {
      mCurrentFilePath=...
    }
```

ImageLounge/src/DkGui/DkNoMacs.cpp
spaghetti22: I don't see how else I can access the path of the current directory in DkExplorer.

ImageLounge/src/DkGui/DkCentralWidget.cpp
spaghetti22: Hi, in my mind the "thumbnail preview panel" is a view of a directory? As far as I see (and a debug statement confirms this), this code is indeed executed when I switch to a directory view (by clicking a directory in the file explorer or by pressing shift+t). I think the terminology is clear enough, but if this is required, I can change it to "thumbViewLoaded" too.

ImageLounge/src/DkGui/DkNoMacs.cpp
leejuyuu: Have you tried opening another file with Ctrl+O when the explorer is shown? The file entry corresponding to the current file will be highlighted and changed accordingly. So, some signal connection in DkExplorer already calls `DkExplorer::setCurrentPath` when the current file changes. You don't need to add another one. 

If you are interested, it's here
https://github.com/nomacs/nomacs/blob/771ebe98e5a434d1ea958933fcec19787362f62d/ImageLounge/src/DkGui/DkNoMacs.cpp#L990

ImageLounge/src/DkGui/DkCentralWidget.cpp
leejuyuu: (This line can be removed)

`directoryLoaded` conveys that we finished creating the data structures that store the info about the directory, and this would mislead reader if emitted when a UI element is shown. The source of the `directoryLoaded` signal should be whatever that is actually reading the contents of the directory (which is DkImageLoader here).

leejuyuu:
> I'd like to remove the dependency on other components. Can you use the file manager's signal to set the path? Something like:
> 
> ```
>     connect(mFileTree->selectionModel(), &QItemSelectionModel::selectionChanged, this, [](...) {
>       mCurrentFilePath=...
>     });
> ```

There's no need to do this as well.

ImageLounge/src/DkGui/DkNoMacs.cpp
spaghetti22: I managed to simplify the code to make this signal unnecessary.

ImageLounge/src/DkGui/DkCentralWidget.cpp
spaghetti22: Signal no longer present in simplified code.

spaghetti22:
Thanks @scrubbbbs for the very helpful suggestion, I managed to get it to work using an even simpler method without dependencies on other components.

leejuyuu:
Looks great, thanks! As these commits are only for adding this feature, I suggest squashing them into one. Also, although it's not a strict rule, please try to limit commit messages below 72 columns wide.

spaghetti22:
Hi, I wasn't able to find out how to squash commits; a web search suggests that this is something you do when you merge the commits?

scrubbbbs:
We like to see one bug fix or new feature in a single commit so it is easier to maintain and review changes.

The "squash" is a git history editing command (from `git rebase  ...` command). It merges series of commits into one commit. So @leejuyuu is asking to take your 4 commits and squash them into one. When doing the squash you also have the opportunity to update the commit message to reflect the final work.

Since this is your first PR I'm happy to do that (squash) for you, just let me know.

leejuyuu:
@spaghetti22 Sorry that I wasn't being clear. I was saying to squash (combine) the 4 commits before and including [711f9ac](https://github.com/nomacs/nomacs/pull/1250/commits/711f9acfe4eb7b0867d036d685b08596b5b0bd43) into one.

Given the current state of this branch, I would do 

1. Rebase is a powerful tool but it's relatively easy to make serious mistakes. Before you get comfortable with it, I suggest always creating a temporary branch before you start, so you can go back easily if things go wrong.
1. Reset to 711f9ac to get rid of the merge commit.
2. Run `git rebase -i HEAD~4`, which allows us to edit the HEAD and its three ancestors. An editor (by default it's vim) will open and would look like this. You can read the comments for what this can do.
image

3. Change the `pick` of the bottom three commits to `s` to combine the changes into the first one. Save and close the editor to continue (`:wq` for vim).
image

4. The editor will open again to allow you to edit the commit message.

5. The whole process looks like this.
image

6. Do a force push to update the PR. Github might prevent you from doing this because you are using your master branch. Generally, people create new branches for contributing since it's more convenient. If this is causing you trouble, it's okay to simply push to another branch and create a separate PR.

Let me know if you have further questions :))

spaghetti22:
Hi, I hope I've now done it correctly according to your instructions. I am sorry for my cluelessness regarding git; I usually use git only from an IDE.

leejuyuu:
It's nothing to be sorry about. We are all still learning here :))

Thanks for your contribution!

Issue #716, opened 2021-07-06
[Minor UI] Renaming of file doesn't force update of file name in the header of APP. 


RatatataBoom:
Stepts to reproduce:
1. Open file (any type).
2. Open File > Rename > Give a new name to the file.
3. Click Save.
4. Check out application header.

Expected result: 
File renamed in file system and in Nomacs

Actual result:
File renamed in file system, but old name still present in Nomacs app header

Environment:

Win10 (19042.1052)
Nomacs ver: 3.16.1709[x64]

leejuyuu:
On 3.19.1 (ba263741) I cannot reproduce this. However, nomacs seems to load the next image on rename, which is a weird behavior.


https://github.com/user-attachments/assets/6bf02251-37d7-48a5-84e7-a09b540cd894

scrubbbbs:
I can't reproduce either, but I was inspired to comb over the function and I fixed a few things.

leejuyuu:
Closing as not reproducible. Feel free to reopen if needed.

PR #1253, opened 2025-01-14
Rename File: do not advance to the next file

scrubbbbs:
Inspired by #716

See commit message

### DIFF
--- a/ImageLounge/src/DkGui/DkCentralWidget.cpp
+++ b/ImageLounge/src/DkGui/DkCentralWidget.cpp
@@ -59,6 +59,12 @@
 #include <QStackedLayout>
 #include <QStandardPaths>
 #include <QTabBar>
+
+#ifdef Q_OS_WIN
+#include <QNtfsPermissionCheckGuard>
+#include <QStorageInfo>
+#endif
+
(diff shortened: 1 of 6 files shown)


leejuyuu:
Seems to work as intended.

 A recommendation for the future: viewing the diff is easier if the refactor step is separated into its own commit, especially for a scenario that moves a big block of text. The "bug fix" is probably just a few lines, but it's not visible in this commit. This is why I like [conventional commit](https://www.conventionalcommits.org/) a lot.

scrubbbbs:
> A recommendation for the future: viewing the diff is easier if the refactor step is separated into its own commit,

:100: agree and thanks for pointing that out. I realized when I made the commit this happened, but then I didn't have an easy way to go back. I will aim to commit early and more often to prevent getting caught in that situation.

leejuyuu:
I can share a trick for splitting a commit if you know it can be decomposed into two steps and the first step is easy (e.g. move a function -> fix bug): 

1. Copy the current file to an untracked file name.
1. Reset to the previous commit.
1. Do the first step and commit.
1. `mv` or `cp` the copied file to the path (it's still there after the reset because it's untracked). You can decide to commit or continue splitting here.

PR #1265, opened 2025-02-08
Directory view usability fixes

spaghetti22:
Various improvements to usability of the directory ("thumb") view, as follows:
- It is now possible to load directories without images from the explorer, this will now show an empty view instead of the images that were in the previous dir.
- Loading the directory view no longer displays information in the statusbar and metadata explorer that was about the previously loaded image. Instead it displays the file name and file size of the image currently selected or hovered over, as applicable.
- Loading the directory view now changes the titlebar to the directory that is loaded.

### DIFF
--- a/ImageLounge/src/DkGui/DkNoMacs.cpp
+++ b/ImageLounge/src/DkGui/DkNoMacs.cpp
@@ -211,6 +211,7 @@ void DkNoMacs::init()
 
     // connections to the image loader
     connect(getTabWidget(), &DkCentralWidget::imageUpdatedSignal, this, QOverload<QSharedPointer<DkImageContainerT>>::of(&DkNoMacs::setWindowTitle));
+    connect(getTabWidget(), &DkCentralWidget::thumbViewLoadedSignal, this, &DkNoMacs::setWindowTitleDirectory);
 
     DkActionManager::instance().enableMovieActions(false);
 
@@ -988,6 +989,7 @@ void DkNoMacs::showExplorer(bool show, bool saveSettings)
         });
         connect(mExplorer, &DkExplorer::openDir, getTabWidget(), &DkCentralWidget::loadDirToTab);
         connect(getTabWidget(), &DkCentralWidget::imageUpdatedSignal, mExplorer, &DkExplorer::setCurrentImage);
+        connect(getTabWidget(), &DkCentralWidget::thumbViewLoadedSignal, mExplorer, &DkExplorer::setCurrentPath);
     }
 
     mExplorer->setVisible(show, saveSettings);
@@ -1014,6 +1016,9 @@ void DkNoMacs::showMetaDataDock(bool show, bool saveSettings)
         addDockWidget(mMetaDataDock->getDockLocationSettings(Qt::RightDockWidgetArea), mMetaDataDock);
 
         connect(getTabWidget(), &DkCentralWidget::imageUpdatedSignal, mMetaDataDock, &DkMetaDataDock::setImage);
+        connect(getTabWidget(), &DkCentralWidget::thumbViewLoadedSignal, mMetaDataDock, [this]() {
+            mMetaDataDock->setImage(nullptr);
+        });
     }
 
     mMetaDataDock->setVisible(show, saveSettings);
@@ -1786,6 +1791,11 @@ void DkNoMacs::setWindowTitle(QSharedPointer<DkImageContainerT> imgC)
     setWindowTitle(imgC->filePath(), imgC->image().size(), imgC->isEdited(), imgC->getTitleAttribute());
 }
 
+void DkNoMacs::setWindowTitleDirectory(const QString &directory)
+{
+    setWindowTitle(directory, QSize(), false, QString());
+}
+
 void DkNoMacs::setWindowTitle(const QString &filePath, const QSize &size, bool edited, const QString &attr)
 {
     QString title;
@@ -1831,21 +1841,27 @@ void DkNoMacs::setWindowTitle(const QString &filePath, const QSize &size, bool e
     // TODO: move!
     DkStatusBar *bar = DkStatusBarManager::instance().statusbar();
 
-    if (((vp && !vp->getController()->getFileInfoLabel()->isVisible())
-         || !DkSettingsManager::param().slideShow().display.testBit(DkSettings::display_creation_date))
-        && getTabWidget()->getCurrentImage()) {
-        // create statusbar info
-        QSharedPointer<DkMetaDataT> metaData = getTabWidget()->getCurrentImage()->getMetaData();
-        QString dateString = metaData->getExifValue("DateTimeOriginal");
-        dateString = DkUtils::convertDateString(dateString, fInfo);
-        bar->setMessage(dateString, DkStatusBar::status_time_info);
-    } else
+    if (fInfo.isDir()) {
         bar->setMessage("", DkStatusBar::status_time_info); // hide label
-
-    if (fInfo.exists())
-        bar->setMessage(DkUtils::readableByte((float)fInfo.size()), DkStatusBar::status_filesize_info);
-    else
         bar->setMessage("", DkStatusBar::status_filesize_info);
+        bar->setMessage("", DkStatusBar::status_filenumber_info);
+    } else {
+        if (((vp && !vp->getController()->getFileInfoLabel()->isVisible())
+             || !DkSettingsManager::param().slideShow().display.testBit(DkSettings::display_creation_date))
+            && getTabWidget()->getCurrentImage()) {
+            // create statusbar info
+            QSharedPointer<DkMetaDataT> metaData = getTabWidget()->getCurrentImage()->getMetaData();
+            QString dateString = metaData->getExifValue("DateTimeOriginal");
+            dateString = DkUtils::convertDateString(dateString, fInfo);
+            bar->setMessage(dateString, DkStatusBar::status_time_info);
+        } else
+            bar->setMessage("", DkStatusBar::status_time_info); // hide label
+
+        if (fInfo.exists())
+            bar->setMessage(DkUtils::readableByte((float)fInfo.size()), DkStatusBar::status_filesize_info);
+        else
+            bar->setMessage("", DkStatusBar::status_filesize_info);
+    }
 }
 
 void DkNoMacs::settingsChanged()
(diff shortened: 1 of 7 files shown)


ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -27,6 +27,7 @@
 
 #include "DkThumbsWidgets.h"
 #include "DkActionManager.h"
+#include "DkBatch.h"
```
leejuyuu, line 30: Why was this included? I think this is not necessary.

ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -1831,21 +1839,27 @@ void DkNoMacs::setWindowTitle(const QString &filePath, const QSize &size, bool e
...
-    if (((vp && !vp->getController()->getFileInfoLabel()->isVisible())
-         || !DkSettingsManager::param().slideShow().display.testBit(DkSettings::display_creation_date))
-        && getTabWidget()->getCurrentImage()) {
-        // create statusbar info
-        QSharedPointer<DkMetaDataT> metaData = getTabWidget()->getCurrentImage()->getMetaData();
-        QString dateString = metaData->getExifValue("DateTimeOriginal");
-        dateString = DkUtils::convertDateString(dateString, fInfo);
-        bar->setMessage(dateString, DkStatusBar::status_time_info);
-    } else
+    if (directory) {
```
leejuyuu, line 1842: You can check whether the path is a directory with `finfo.isDir()`, then there's no need to pass in `directory`, and you won't need `setWindowTitleDirectory` either.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -1224,19 +1226,39 @@ void DkThumbScene::keyPressEvent(QKeyEvent *event)
...
+}
 
-        if (sf > 1)
-            info = QString::number(sf) + tr(" selected");
-        else
-            info = QString::number(mThumbLabels.size()) + tr(" images");
+void DkThumbScene::displayFileInfoInStatusbar(const QString &filePath)
+{
+    QFileInfo fInfo(filePath);
+    QString info = QFileInfo(filePath).fileName();
```
leejuyuu, line 1251:
```suggestion
```

This line is unused and can be removed.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -1224,19 +1226,39 @@ void DkThumbScene::keyPressEvent(QKeyEvent *event)
...
+    DkStatusBar *bar = DkStatusBarManager::instance().statusbar(); 
+    if (fInfo.exists()) {
+        bar->setMessage(DkUtils::readableByte((float)fInfo.size()), DkStatusBar::status_filesize_info);
```
leejuyuu, line 1254:
```suggestion
        bar->setMessage(DkUtils::readableByte(fInfo.size()), DkStatusBar::status_filesize_info);
```

Int can be converted to float, no need to cast.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -1224,19 +1226,39 @@ void DkThumbScene::keyPressEvent(QKeyEvent *event)
...
+        bar->setMessage(filePath);
+        QString dateString = DkUtils::convertDateString(dateString, fInfo);
```
leejuyuu, line 1256:
```suggestion
```

This doesn't seem to do anything as well.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -1224,19 +1226,39 @@ void DkThumbScene::keyPressEvent(QKeyEvent *event)
...
+    } else { // i.e. user STARTED hovering over a file
+        displayFileInfoInStatusbar(filePath);
+    }
+}
 
-        if (sf > 1)
-            info = QString::number(sf) + tr(" selected");
-        else
-            info = QString::number(mThumbLabels.size()) + tr(" images");
+void DkThumbScene::displayFileInfoInStatusbar(const QString &filePath)
```
leejuyuu, line 1248: This function does not depend on `DkThumbScene`, so I suggest making it a local function, i.e. simply `displayFileInfoInStatusbar`, and don't expose it in the header. You can move it before `DkThumbScene::showFile` to avoid forward declaration.

leejuyuu (changes requested):

leejuyuu:
Please don't include any previous merged commit in a PR. We try to maintain a linear commit history, so
1. Always branch off from the main/master branch when contributing.
2. If the main/master branch updates during the work on your branch, rebase onto the main branch instead of merging the main branch in.

In this case, please do a `git rebase <nomacs>/master`. Fill `<nomac>` with the remote name that points to this repo.

I mentioned last time that the commit message should generally be no longer than 80 columns wide. It's mostly due to historical reasons, but tools often assume people follow this. So you see below there's a weird wrapping of the title. See more in [this article](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
image

Therefore, I would wrap the commit messages like this
image

I think many editor can actually do this for you as you are typing.

ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -1014,6 +1016,7 @@ void DkNoMacs::showMetaDataDock(bool show, bool saveSettings)
         addDockWidget(mMetaDataDock->getDockLocationSettings(Qt::RightDockWidgetArea), mMetaDataDock);
 
         connect(getTabWidget(), &DkCentralWidget::imageUpdatedSignal, mMetaDataDock, &DkMetaDataDock::setImage);
+        connect(getTabWidget(), &DkCentralWidget::thumbViewLoadedSignal, mMetaDataDock, &DkMetaDataDock::clear);
```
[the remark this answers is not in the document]
spaghetti22: Done, thanks for the suggestion, I wasn't aware this was possible.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
spaghetti22: You are right, I think this was a relic from an older working version?

ImageLounge/src/DkGui/DkNoMacs.cpp
spaghetti22: I think I still need `setWindowTitleDirectory` unless I do it with a lambda as above (which is somewhat harder to read when we are creating QSize() and QString() here)? But I changed it to isDir.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
spaghetti22: You are right, copy and paste relic.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
spaghetti22: I changed it here, but the reason I did it like this originally was that I copied the code from `DkNoMacs::setWindowTitle` (currently line 1861 in my branch). Should it not be changed there too then?

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
spaghetti22: You are right, thanks.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
spaghetti22: OK, done, thanks. (I suspect you can tell I am mainly a Java developer and not used to this being possible.)

spaghetti22:
Hi, I am not sure how that previous commit ended up in there; I clearly need to get more practice with git especially on the command line.

I changed the commit message to have at most 80 characters per line and will pay attention to that in the future.

ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -211,6 +211,7 @@ void DkNoMacs::init()
 
     // connections to the image loader
     connect(getTabWidget(), &DkCentralWidget::imageUpdatedSignal, this, QOverload<QSharedPointer<DkImageContainerT>>::of(&DkNoMacs::setWindowTitle));
+    connect(getTabWidget(), &DkCentralWidget::thumbViewLoadedSignal, this, &DkNoMacs::setWindowTitleDirectory);
```
leejuyuu, line 214:
```suggestion
    connect(getTabWidget(), &DkCentralWidget::thumbViewLoadedSignal, this, [this](const QString &path) {
        setWindowTitle(path);
    });
```

You can do this because `DkNoMacs::setWindowTitleDirectory` has default values for the last three parameters.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
leejuyuu: Sure! You can do it in a separate commit. Thanks!

leejuyuu:
> I changed the commit message to have at most 80 characters per line and will pay attention to that in the future.

Please take a look at the commit title as well, they are generally kept shorter (~50 chars). Thanks!

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
leejuyuu: That's fine. It feels nice to share cool things :)).

ImageLounge/src/DkGui/DkNoMacs.cpp
spaghetti22: done, thanks

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
spaghetti22: done, as suggested, in a separate commit

leejuyuu:
This is good to go. Thanks!

scrubbbbs:
While this works well with explorer, it breaks a corner case; using Shift-T to toggle thumbnail view, but deciding not to change the current image, will not restore the previous state of the viewer (e.g. title bar etc will remain empty until switching images).

Rather than try to transform the app for the directory view, maybe we can make it more obvious that the dir view is functionally an overlay on top of the image? I think that would be easier than trying to account for all the state changes w/o unloading/reloading the image (which is also an option; but has issues as unsaved images etc have to be dealt with).

PR #1273, opened 2025-02-22
Dock usability improvements

spaghetti22:
- Closing a detached dock window will now reattach it instead of closing it. Previously (it seems) there was absolutely no way to reattach a dock without closing the app entirely. (If this causes other usability issues, I can attempt to do it differently too.)
- The metadata dock will no longer display a thumb of the previously loaded image when a directory (thumb) view is currently open. (Actually I do not see the usefulness of this thumb even when an image is open, but I am sure it was added for a reason.)

### DIFF
--- a/ImageLounge/src/DkGui/DkBaseWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkBaseWidgets.cpp
@@ -32,6 +32,7 @@
 
 #pragma warning(push, 0) // no warnings from includes - begin
 #include <QAction>
+#include <QCloseEvent>
 #include <QComboBox>
 #include <QDebug>
 #include <QEvent>
@@ -595,9 +596,14 @@ void DkDockWidget::setVisible(bool visible, bool saveSetting)
 
 void DkDockWidget::closeEvent(QCloseEvent *event)
 {
-    setVisible(false);
+    if (isFloating()) {
+        setFloating(false);
+        event->ignore();
+    } else {
+        setVisible(false);
 
-    QDockWidget::closeEvent(event);
+        QDockWidget::closeEvent(event);
+    }
 }
 
 Qt::DockWidgetArea DkDockWidget::getDockLocationSettings(const Qt::DockWidgetArea &defaultArea) const
--- a/ImageLounge/src/DkGui/DkMetaDataWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkMetaDataWidgets.cpp
@@ -440,8 +440,10 @@ void DkMetaDataDock::updateEntries()
     mModel = new DkMetaDataModel(this);
     if (!mImgC) {
         mProxyModel->setSourceModel(mModel);
+        mThumbNailLabel->hide();
         return;
     }
+    mThumbNailLabel->show();
     mModel->addMetaData(mImgC->getMetaData());
     mProxyModel->setSourceModel(mModel);
 


leejuyuu:
> * Closing a detached dock window will now reattach it instead of closing it. Previously (it seems) there was absolutely no way to reattach a dock without closing the app entirely. (If this causes other usability issues, I can attempt to do it differently too.)

This overwrites the intent of the close button, which might not be a good idea. Implementing redocking via mouse drag is probably the best, but also probably difficult. GIMP always starts a dockable dialog in the docked state, and I guess this might be easier to do.

scrubbbbs:
> * Previously (it seems) there was absolutely no way to reattach a dock without closing the app entirely.

Double-clicking the title bar of the dock widget undocks/redocks the widget.

> * The metadata dock will no longer display a thumb of the previously loaded image when a directory (thumb) view is currently open. (Actually I do not see the usefulness of this thumb even when an image is open, but I am sure it was added for a reason.)

The original intent is only to show if there is an embedded thumbnail, as this is part of the image metadata. This is corrected in #1238 which will be merged soon.

The thing about the thumbnail grid view is it that doesn't change the currently loaded image when it is shown. So it behaves like dock widgets in that it does not modify the state of the viewer until some action is taken.

So following that logic, the embedded thumbnail shouldn't be hidden, just like the other metadata (text data).

spaghetti22:
Hi, first of all thanks for the feedback, I will probably not have much time in the next few days to work on this further, but certainly will work on a better solution.

> Double-clicking the title bar of the dock widget undocks/redocks the widget.

As far as I can tell (on a KDE Plasma desktop), it doesn't. It does undock it (detach it from the main window), but when it is undocked, the only "title bar" that is still visible is that drawn by the window manager. Double clicking that does whatever the window manager is configured to do when double-clicking a titlebar, in my case that is maximizing it.

What I have now been trying to implement is a narrow button on the very top of each dock window that would redock it. I am still working on and thinking about the details; would this be an acceptable UI?

scrubbbbs:
> As far as I can tell (on a KDE Plasma desktop), it doesn't. It does undock it (detach it from the main window), but when it is undocked, the only "title bar" that is still visible is that drawn by the window manager. Double clicking that does whatever the window manager is configured to do when double-clicking a titlebar, in my case that is maximizing it.

I can appreciate the issue now. Even on Windows, the dock/undock widget (icon next to close box) is lost when undocking:

image
image

This is not the case on all window managers, gnome for example preserves it.

> What I have now been trying to implement is a narrow button on the very top of each dock window that would redock it. I am still working on and thinking about the details; would this be an acceptable UI?

Yes, but I think the solution you are looking for is [QDockWidget::setTitleBarWidget](https://doc.qt.io/qt-6/qdockwidget.html#setTitleBarWidget), which gives full control over the title bar area. It is not clear to me if this works when undocked but the documentation suggests it will. 

So there will be no need for a "narrow button" or some other compromise; it can be implemented similar to the default implementation with the float/unfloat button being next to the close button as in the Qt native version.

This will be a bit of extra work (e.g. needs close button, event handler maybe for double click action), but I think the result will be nicer.

spaghetti22:
I can see that the problems I was going to solve with this pull request have already been fixed in master by now. I am sorry for only getting back to you now. I am closing this pull request because there is nothing to do here anymore.

PR #1275, opened 2025-02-26
Update LibRaw submodule

novomesk:


scrubbbbs:
Thanks for looking into this. Unfortunately I have led you down the wrong path (I am getting reacquainted with Windows debugging...) 

It seems to definitely be a missing DLL. `vcomp14.dll` (looks like OpenMP support library). I don't know why windows did not just say this was missing, I had to pull out dependency walker:

image

scrubbbbs:
I forgot to mention this build of libraw also seems to be missing jpeg support for previews

novomesk:
This problematic file is not loaded via the preview?
image

Anyway, this update solves some security issues/bugs that were discovered since previous version.

scrubbbbs:
> This problematic file is not loaded via the preview?

Nope. If I enable the preview I get this in the debugger log:

```
DebugString: "[CRITICAL] "JPEG compression support is not configured""
DebugString: "[CRITICAL] "JPEG compression support is not configured""
DebugString: "[CRITICAL] "Sorry, requested compression method is not configured""
DebugString: "[CRITICAL] "Sorry, requested compression method is not configured""
Skipping unsupported debug type IMAGE_DEBUG_TYPE_POGO in module raw.dll...
Did not find any supported debug types in module raw.dll!
DLL Loaded: [00007FFC72430000](x64dbg://localhost/address64#00007FFC72430000) C:\Users\John\Downloads\nomacs-portable-win\nomacs-3.19.1\raw.dll
Thread 3924 created, Entry: ntdll.[00007FFCCAA659B0](x64dbg://localhost/address64#00007FFCCAA659B0), Parameter: [000001EDDFF0E590](x64dbg://localhost/address64#000001EDDFF0E590)
DLL Unloaded: [00007FFC72430000](x64dbg://localhost/address64#00007FFC72430000) raw.dll
EXCEPTION_DEBUG_INFO:
           dwFirstChance: 1
           ExceptionCode: C06D007E (VCPP_EXCEPTION_ERROR_MOD_NOT_FOUND)
          ExceptionFlags: 00000080
        ExceptionAddress: kernelbase.[00007FFCC840BB0A](x64dbg://localhost/address64#00007FFCC840BB0A)
        NumberParameters: 1
ExceptionInformation[00]: [000000DF8A143E30](x64dbg://localhost/address64#000000DF8A143E30)
First chance exception on [00007FFCC840BB0A](x64dbg://localhost/address64#00007FFCC840BB0A) (C06D007E, VCPP_EXCEPTION_ERROR_MOD_NOT_FOUND)!
```
However the debug string does not appear anywhere in raw.dll or portable directory files (windows component?) and it comes before the raw.dll finishes loading which is quite strange. The debug prints do go away if I disable the preview image however.

novomesk:
Sometimes I have impression that it would be good to build under msys2 or to use KDE craft build system.

scrubbbbs:
> Sometimes I have impression that it would be good to build under msys2 or to use KDE craft build system.

I'm not familiar with KDE Craft, but apparently it uses windeployqt which will not find dependencies for delay-load or runtime-loaded DLLs (e.g. nomacs plugins, raw.dll). But maybe it has other tricks up its sleeve. Otherwise it would not catch this particular issue (openmp DLL missing). Since delay-load DLLs are not in the import table, the only way to find them is to execute the code where they are loaded. 

And then there are other DLLs that are loaded explicitly by name like plugins which could also have more dependencies.

It is possible to do runtime dll checking [in a script](https://github.com/nomacs/nomacs/blob/master/ImageLounge/cmake/MingwPackager.sh), but that requires executing a codepath that will load all DLLs that might be used by the application. One way would be an argument to nomacs that touches all the subsystems known to use delay-loaded or LoadLibrary() libraries.

Issue #1283, opened 2025-03-16
Visual glitches and inconsistencies since updating


namaenonaimumei:
Pretty generic title, as I couldn't confirm what exactly is the root cause of issues mentioned below (but I felt like they are all part of the same bug).

**Describe the bug**
Since updating to latest version of nomacs, rendering seems to behave very different from before - has renderer been reworked/changed in few past months, during switch to qt6?

More on the issue in depth below:
*  When using nomacs to directly launch in fullscreen mode (not slideshow), slideshow navigation panel always shows up now, regardless of whether you pressed space/hovered over it and so, fading out after 3 seconds as per relevant option in settings - disabling Display>Show Player makes player no longer show up at all, not sure if that's a workaround or intended way panel functions now - but I decided to mention this, as it also causes visual glitch mentioned in the next part.

* ~~When previewing image with player panel enabled, startup feels sluggish?/takes longer?, as it seems to draw in delay separating content into 3 parts - first being bottom line roughly covering player panel layer with part of image below it, second being rest of the image vertically in line with panel and third being the remainder only drawn on input event (see next point), this behaviour differs a bit between filetypes/filesize possibly implying loadtime difference, but is never "right" and overall doesn't seem intentional to me (but may be caused entirely by issues described below, which become more obvious once player panel is disabled).~~ **Fixed by removing old files in `/usr/local/share/nomacs/Image Lounge/themes/`, see #1284**

* ~~This is the biggest one and seems to be caused by renderer buffering or similar (can be qt related, although I mostly run qt6 apps and this never happened elsewhere). To describe it, nomacs seems to always refresh 1 frame behind, that is, example, when switching preview to the next image with **keyboard keys**, only previous image displays, no image being displayed at all when launching direct fullscreen at first, updating only once you hover over navigational buttons/use those instead of keyboard/call any UI panel through a keybind/press Tab/do anything else that nomacs itself considers an "input" event in general. The easiest way to see this behaviour is to preview a gif, it will only draw new frames as long as Tab is being pressed.~~ **Fixed by removing old files in `/usr/local/share/nomacs/Image Lounge/themes/`, see #1284**

* You can also see UI elements (that should be hidden) blinking for a brief moment, when launching directly into fullscreen mode, not sure whether that's a separate issue.

[...]

PR #1285, opened 2025-03-17
Fix startup glitches related to fullscreen mode

scrubbbbs:
See commits

Related: #1283

### DIFF
--- a/ImageLounge/src/DkCore/DkBaseViewPort.cpp
+++ b/ImageLounge/src/DkCore/DkBaseViewPort.cpp
@@ -507,17 +507,7 @@ void DkBaseViewPort::contextMenuEvent(QContextMenuEvent *event)
 // protected functions --------------------------------------------------------------------
 void DkBaseViewPort::draw(QPainter &painter, double opacity)
 {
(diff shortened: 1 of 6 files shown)


novomesk:
Do you plan more work on this?

scrubbbbs:
> Do you plan more work on this?

With that last commit I think this is all for now. I've filed [this Qt bug](https://bugreports.qt.io/browse/QTBUG-134893) which is a source of some glitches.

There are a couple of things on my list but nothing that needs to hold up this PR right now if you want to move forward.
- Make toolbar/statusbar visibility bound to display mode for consistency
- Disable widget fading when switching modes and maybe with "hide all panels" as the former causes HUD widgets to jump around after switching.

novomesk:
Please rebase to make it easier to test all recent changes together.

scrubbbbs:
@novomesk, @leejuyuu  are trying to get away from practice of squashing PRs, I'd like to keep it that way as it is easier to follow the change history (mainly) and would be much easier to back things out if we need to. If it feels like some squashing is needed I am happy to rebase and squash the related stuff, and come to some sort of compromise.

PR #1290, opened 2025-03-22
fixes compilation when LIBHEIF is enabled

souravdas142:
When enables 
`-DENABLE_HEIF=ON`

still its showing 
`nomacs will be compiled with HEIF support .................... NO`

### DIFF
--- a/ImageLounge/CMakeLists.txt
+++ b/ImageLounge/CMakeLists.txt
@@ -360,7 +360,7 @@ ELSE()
     MESSAGE(STATUS " nomacs will be compiled with plugin support .................. NO")
 ENDIF()
 
-IF(libheif_FOUND)
+IF(LIBHEIF_FOUND)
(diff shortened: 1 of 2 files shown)


scrubbbbs:
It seems ENABLE_LIBHEIF is not mentioned anywhere in the source code or build files, so the proper fix would be to remove it. It may have been retained to avoid breaking package maintainers' scripts.

HEIF support is now provided by a qt imageformat plugin such as kimageformats.

leejuyuu:
I remember @novomesk mentioned that ``ENABLE_HEIF`  is used only for MSVC builds.

novomesk:
`IF(libheif_FOUND)` will be removed. It is most probably some old remnant from the times, when nomacs built `qt-heif-image-plugin`

On Linux we recommend to use heif plugin from `kimageformats` package.

PR #1295, opened 2025-03-27
cmake: add precompiled header (default disabled)

scrubbbbs:
I am a bit frustrated with nomacs compile times, this helps a little bit without much effort. The list of headers included is fairly conservative right now, and would be faster if tuned for the working set of files for a particular feature.

The biggest speedup is using a debug build with plugins disabled...this is a must for me right now.

```
config: clang-19, debug, plugins disabled

make clean:              37s / 27s
touch dksettings.h:      27s / 20s
touch dkthumbs.h:        23s / 19s
touch dkthumbswidgets.h: 13s / 11s
touch dkwidgets.h:       14s / 12s
touch dkcentralwidget.h  16s / 14s

config: gcc-14, debug, plugins disabled

make clean:              35s / 31s
touch dksettings.h:      25s / 20s
touch dkthumbs.h:        23s / 21s
touch dkthumbswidgets.h: 13s / 12s
touch dkwidgets.h:       13s / 13s
touch dkcentralwidget.h  16s / 15s
```


leejuyuu:
Have you tried using ccache? I generally don't need to compile the whole thing every time so this is helpful. It eats disk space though.

scrubbbbs:
I haven't tried ccache in nomacs, AFAIK it won't help if a header is modified that is included by many files. DkSettings.h for example affects 44 files (with plugins disabled). It would help when only changing build configurations but that isn't something I do very often.

PR #1305, opened 2025-04-02
Various Mac fixes

scrubbbbs:
Goals for this PR:

- [x] Update icon
- [x] File types registration
- [x] Fix RAW loader segfault
- [x] Include kimageformats (no homebrew package)
- [x] MacOS github workflow

image


ImageLounge/macosx/build-kif.sh
```diff
@@ -0,0 +1,59 @@
...
+# highly annoying auto-update behavior potentially breaking things
+export HOMEBREW_NO_AUTO_UPDATE=1
+
+# downgrade for extra-cmake-modules 6.13, unavailable in homebrew
+KIF_VERSION=752b18a
+
+SRC_DIR="kimageformats"
+BUILD_DIR="kimageformats/build"
+
+brew install --quiet libde265 openexr libavif jpeg-xl openjpeg jxrlib ninja \
```
novomesk, line 30: `libheif` is needed too.

hmueller01:
I just tested this PR on a real Intel MacBook Pro (2019):
```
nomacs 3.21.0, x86_64
6a2e85b8, 2025-04-03


macOS Sequoia (15.4), 0GB
Qt 6.8.2, cocoa, scale=2.00
Exiv2 0.28.5
LibRAW 0.21.3-Release
OpenCV 4.11.0
TIFF 4.7.0
QuaZip v1
```
without building the kimageformats. It can load jp2 and heic formats without crash ... 

Here the cmake output:
```
> cmake -D QT_VERSION_MAJOR=6 -D ENABLE_QUAZIP=ON --install-prefix /Applications ../ImageLounge/.
-- The C compiler identification is AppleClang 17.0.0.17000013
-- The CXX compiler identification is AppleClang 17.0.0.17000013
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Performing Test HAVE_STDATOMIC
-- Performing Test HAVE_STDATOMIC - Success
-- Found WrapAtomic: TRUE
-- Found OpenGL: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/OpenGL.framework
-- Found WrapOpenGL: TRUE
-- Could NOT find WrapVulkanHeaders (missing: Vulkan_INCLUDE_DIR) 
-- Found Cups: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/lib/libcups.tbd (found version "2.3.4")
-- A release build (non-debug). Debugging outputs are silently ignored.
-- Found PkgConfig: /usr/local/bin/pkg-config (found version "2.4.3")
-- Checking for module 'exiv2>=0.27'
--   Found exiv2, version 0.28.5
-- exiv2 version by pkg-config: 0.28.5
-- exiv2 contains exiv2.hpp, using it
-- Found OpenCV: /usr/local/Cellar/opencv/4.11.0_1 (found version "4.11.0") found components: core imgproc
-- 
-- APPLE: some hacks with OpenCV libraries linking will be performed (only limited set of libs will be used)
-- APPLE:    originals: opencv_core;opencv_imgproc
-- APPLE:          new: opencv_core;opencv_imgproc
-- 
-- Checking for module 'libraw>=0.12.0'
--   Found libraw, version 0.21.3
-- Found TIFF: /usr/local/lib/libtiff.dylib (found version "4.0.9")
-- QUAZIP: QuaZip-1 for Qt6 found
-- Could NOT find GTest (missing: GTEST_LIBRARY GTEST_INCLUDE_DIR GTEST_MAIN_LIBRARY) 
CMake Warning at CMakeLists.txt:129 (message):
  GTest not found, disabling testing.


-- revision is 6a2e85b8, 2025-04-03
CMake Warning at cmake/MacBuildTarget.cmake:132 (message):
  Qt jpeg2000 plugin installed, this is known to crash nomacs (#1307)

  We recommend disabling this in homebrew.  For jpeg2000 support you may
  install kimageformats package or run 'make kimageformats' to build from
  source.

  To disable you may delete:
  /usr/local/share/qt/plugins/imageformats/libqjp2.dylib and
  /usr/local/lib/cmake/Qt6Gui/Qt6QJp2Plugin*.cmake

Call Stack (most recent call first):
  CMakeLists.txt:239 (include)


-- Found Qt plugins path: /usr/local/share/qt/plugins
-- nomacs version is 3.21.0
-- using plugin directory: /Users/hmueller/src/opensource/nomacs.hmueller01/ImageLounge/plugins
-- 
-- ----------------------------------------------------------------------------------
--  nomacs - Image Lounge 3.21  <https://nomacs.org>
--  Qt version: 6.8.2
--  nomacs will be compiled with OPENCV support .................. YES
--  nomacs will be compiled with LIBRAW support .................. YES
--  nomacs will be compiled with extended TIFF support ........... YES
--  nomacs will be compiled with plugin support .................. YES
--  nomacs will be compiled with Translations .................... YES
--  nomacs will be compiled with QuaZip support .................. YES
--  nomacs tests ................................................. NO
--  nomacs benchmarks ............................................ NO
-- ----------------------------------------------------------------------------------
-- Configuring done (10.8s)
-- Generating done (0.4s)
-- Build files have been written to: /Users/hmueller/src/opensource/nomacs.hmueller01/build
```

ImageLounge/macosx/build-kif.sh
```diff
@@ -0,0 +1,61 @@
...
+# downgrade for extra-cmake-modules 6.13, unavailable in homebrew
+KIF_VERSION=752b18a
+
+SRC_DIR="kimageformats"
+BUILD_DIR="kimageformats/build"
+
+brew install --quiet libde265 openexr libavif libheif jpeg-xl openjpeg jxrlib ninja \
+  extra-cmake-modules karchive vulkan-headers || exit 1
+
+if [ ! -d $SRC_DIR ]; then git clone https://github.com/KDE/kimageformats || exit 2; fi
```
novomesk, line 33: The original source is `https://invent.kde.org/frameworks/kimageformats.git`

But the git snapshot usually require very latest `extra-cmake-modules` and `karchive`

I am afraid that the build may break at any time in the future.

leejuyuu: I suggest specifying the latest tag instead of using the main branch.

scrubbbbs: The kimageformats part is optional, so technically it cannot break the build, e.g. "make" will always produce a working nomacs, albeit with less format support. This would be an annoyance for maintaining release builds of nomacs at worst, but I think depending on homebrew is going to be the bigger headache (they are always changing, updating and dropping older package versions etc).

So I would propose we mitigate the issue with a series of steps (I think this is a reasonable order):
- the changes mentioned here (origin repo and release tag)
- keep kimageformats as an optional, manual build step
- get this into github CI workflow
- submit a homebrew formula for kimageformats
- submit a formula for nomacs so homebrew is less likely to break our build

scrubbbbs:
> without building the kimageformats. It can load jp2 and heic formats without crash ...

@hmueller01 thanks for testing. The crash happened every time I loaded the thumbnails panel (shift+T), specifically with the files mentioned in #1307 (I did not check anything else), but not as frequent when browsing.

The macos version of QtGui has support heic built-in unlike other platforms, so kimageformats is not needed for that. However libheif should have more capabilities than the Qt plugin so we should include it. I don't expect it would cause a conflict, the image loader will probably choose the kimageformats one first (alphabetically), and potentially fall back to the Qt plugin if it can't open the file.

hmueller01:
I also answered on #1307. Thumbnails panel does not crash on my side ...

ImageLounge/macosx/build-kif.sh
scrubbbbs: I added a workflow for macos which includes building kimageformats. For now its just the minimal configuration of nomacs as we don't have the minutes to build the full release binary (which takes 15-20 minutes mostly due to OpenCV)

The workflow is disabled for now as I burned about 500 minutes getting it to work (2000 monthly allocation). We may want to drop some builds from the linux.yml that don't have much value to save some minutes.

ImageLounge/tests/CMakeLists.txt
```diff
@@ -13,7 +13,10 @@ target_link_libraries(
 
 add_custom_target(
     check
-    COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure
+    COMMAND
+        LD_LIBRARY_PATH=../:$ENV{LD_LIBRARY_PATH}
+        DYLD_LIBRARY_PATH=../:$ENV{DYLD_LIBRARY_PATH}
+        ${CMAKE_CTEST_COMMAND} --output-on-failure
```
leejuyuu, line 19: I think this should not be necessary. The `make check` works both on my machine and on CI. What does your `ldd <build_dir>/tests/core_tests` output?

ImageLounge/tests/DkUtils_test.cpp
```diff
@@ -1,7 +1,16 @@
...
 #include <string>
 
+TEST(LinkedVersionTest, Test)
+{
+    // if these don't match, we definitely have an invalid test environment
+    // if they do, we may or may not...a build serial number would be required,
+    // but that is not something package maintainers like.
+    EXPECT_EQ(std::string(NOMACS_REVISION_STR), std::string(nmc::revisionString));
+}
+
```
leejuyuu, line 13: This test doesn't make sense to me.

README.md
```diff
@@ -218,44 +218,60 @@ Install required dependencies:
...
+cd nomacs; mkdir build; cd build
+
+# Qt5 / Intel
+$ Qt5_DIR=/usr/local/opt/qt5/ cmake -D QT_VERSION_MAJOR=5 ../ImageLounge/.
+
+# Qt6 / Intel or Apple Silicon 
+$ cmake -D QT_VERSION_MAJOR=6 -D ENABLE_QUAZIP=ON ../ImageLounge/.
+
+# Qt5 / Apple Silicon
+$ Qt5_DIR=/opt/homebrew/opt/qt5/ cmake -DQT_QMAKE_EXECUTABLE=/opt/homebrew/opt/qt5/bin/qmake ../ImageLounge/.
```
leejuyuu, line 233: Please avoid placing the shell prompt in code blocks. The content of the code block should be a valid shell script.

README.md
```diff
@@ -218,44 +218,60 @@ Install required dependencies:
...
 ```
 
-For Homebrew on Intel models:
+Run make:
 
 ```console
-# qt5
-$ export CPLUS_INCLUDE_PATH=/usr/local/include
-$ Qt5_DIR=/usr/local/opt/qt5/ cmake -D QT_VERSION_MAJOR=5 --install-prefix /Applications ../ImageLounge/.
+$ make
```
leejuyuu, line 239: Shell prompt.

README.md
```diff
@@ -218,44 +218,60 @@ Install required dependencies:
...
 
-Run make:
+Nomacs registers supported file types via the Info.plist file in the app bundle. This is essential for open-with, drag-and-drop, etc features of the Finder. Supported types vary depending on what options to cmake, Qt and OS version, homebrew configuration, and even nomacs user-specified custom file types (via `Tools/Add Image Format`). 
+
+Note that nomacs does not automatically make itself the default application for any supported types at this time, you will need to you use the open-with function in "Get Info" etc.
 
+To ensure it is correct for the current build, run
 ```console
+$ make filetypes
 $ make
```
leejuyuu, line 265: Shell prompt

README.md
```diff
@@ -265,7 +281,7 @@ If macdeployqt complains about `ERROR: Cannot resolve rpath "@rpath/QtGui.framew
 ```console
 $ cd /usr/local/lib/QtGui.framework/Versions/A
 $ install_name_tool -id '@rpath/QtGui.framework/Versions/A/QtGui' QtGui
-% otool -L QtGui| head -2
+$ otool -L QtGui | head -2
```
leejuyuu, line 284: Shell prompt.

ImageLounge/macosx/build-kif.sh
leejuyuu: I took a look at the stats. Last month we used ~1300 minutes, so I think it will probably be okay. Let's take note that when testing CI stuff (frequent push), we should probably skip the likely unaffected part of the build, and add them back right before merging.

scrubbbbs: The macos.yml *minimal* build for both Intel/Arm combined is about 3x slower (6-7  minutes vs 2-3) than the current linux.yml. So we will probably go over and dropping stuff from linux.yml isn't going to save too much. I think this means we will most likely go over our allocation if we run it on every pull request.

We might want to keep it manual and we'll just have to remember to run it before merging pull requests, unless there is a way to enforce that.

> we should probably skip the likely unaffected part of the build

Yes, I realized that a bit late but did eventually have appveyor and linux.yml disabled while working out the issues.

leejuyuu: > 3x slower (6-7 minutes vs 2-3)

Does the time mostly spent on compiling nomacs? Or other things like installing dependencies are also slow?

ImageLounge/tests/CMakeLists.txt
scrubbbbs: My current build "make check" segfaults because I did a "sudo make install" on another branch (I usually have the master branch installed to /usr/local). I think you are lucky or you have LD_LIBRARY_PATH set to the build directory or in ld.so.conf

```
ldd tests/core_tests | grep nomacsCore
	libnomacsCore.so.3 => /usr/local/lib/libnomacsCore.so.3 (0x00007e18de200000)
```

ImageLounge/tests/DkUtils_test.cpp
scrubbbbs: It checks if the runtime version of nomacsCore matches the compile-time version in the test executable. The typical cause comes from the previous change discussed.

ImageLounge/macosx/build-kif.sh
scrubbbbs: It's all slow. But the dependencies are extremely slow compared to apt, it takes around 5-7 minutes to pull everything including OpenCV (which adds 4-5 minutes)

ImageLounge/tests/CMakeLists.txt
leejuyuu: I don't set `LD_LIBRARY_PATH` and I install nomacs only to /usr/lib and /usr/bin. This is what I get.

image

I think this is an environment issue, and probably should be fixed in the environment instead of the build config.

leejuyuu: I am not a fan of this, but if you think this is necessary, please use `CMAKE_BINARY_DIR` instead of a relative path.

ImageLounge/tests/DkUtils_test.cpp
leejuyuu: The comments and commit message are misleading. Both strings were set at compile time. The intention of the test seems to be "check whether the test executable and the libnomacsCored.so were compiled with the same `NOMACS_REVISION_STR`."

I'm starting to think these are solving the wrong problem. I used the dynamic library when I added the tests because it is already available. However, to ensure there is no mismatch, perhaps it is better to statically link the test executable.

scrubbbbs: The intention is to to keep from loading the wrong library. And it does this. As long as the git repository is clean, the revision string is unique to the build.

The "wrong problem" is a misconfigured environment or stale build of nomacscore, and there is nothing we can do to solve that directly.

novomesk:
Do you expect more more on this? If you feel it is fine, merge.

PR #1262, opened 2025-02-02
Thumbnails

leejuyuu:

PR #1314, opened 2025-04-29
Improve Zipfile Support

scrubbbbs:
Goals for this PR:
- [x]  Remove file path member from DkImageContainer and use QFileInfo instead
- [x] Add DkFileInfo for uniform access to files regardless of type (ordinary, zip, etc)
- [x] Replace QFileInfo with DkFileInfo in DkImageContainer
- [x] Fix now (intentionally) broken zip file support up to where it is currently
- [x] Fix sorting of zip files by metadata (file size,created date etc)
- [x] Fix display of zip files metadata in panels, dock
- [x] Fix opening of zip files from history
- [x] Fix opening of zip files from explorer (view loses selection)
- [x] Fix animations in zip files
- [x] Fix incorrect sorting of zip files by name
- [x] Check if we broke compatibility with older quazip versions
- [x] Fix folder filter for zips
- [x] Fix explorer selected item for zips
- [x] Fix directory modtime check for zips 
- [x] Fix Drop/paste zip files / zip file urls
- [x] Merge with master after PR #1262 is merged
- [x] Fix display of zip file metadata in thumbnail view
- [x] Fix zip file thumbs in history
- [x] Make zip usable for batch input


leejuyuu:
Did someone enable Copilot on this repo?

ImageLounge/src/DkCore/DkFileInfo.h
```diff
@@ -0,0 +1,215 @@
...
+    {
+        friend class DkFileInfo;
+
+    private:
+        SharedData(const QFileInfo &info);
+        SharedData(const QString &zipPath, const QuaZipFileInfo64 &info);
+
+        QFileInfo mFileInfo; // zip: constructed from encoded path in the zipfile
+        QFileInfo mContainerInfo; // zip: info of the .zip file itself
+        ZipData mZipData;
```
leejuyuu, line 208: This seems to make every DkFileInfo construction pay for the allocation of the ZipData (80 bytes) even if QuaZIP is not enabled or the file is not a zip file.

ImageLounge/src/DkCore/DkFileInfo.cpp
```diff
@@ -0,0 +1,530 @@
...
+#endif
+#endif
+
+#include <QDir>
+
+namespace nmc
+{
+
+// this is the same marker as KIO
+const QString DkFileInfo::ZipData::mZipMarker = "#/";
```
leejuyuu, line 46: This could be `constexpr` and initialized in the header.

leejuyuu (commented):
I only scanned through the two new files quickly. Sorry.

ImageLounge/src/DkCore/DkFileInfo.h
scrubbbbs: It does. I was trying to remove as much conditional code as possible and settled on this. I suppose it could be a pointer but that also complicates things. On the one hand, it uses more memory, on the other, it gives DkFileInfo a fixed size which should help with heap fragmentation and requires fewer allocations.

It could be a factory pattern so each type of DkFileInfo was a different subclass etc, but I don't think the complication is worth it right now.

ImageLounge/src/DkCore/DkFileInfo.cpp
```diff
@@ -0,0 +1,530 @@
...
+        io.reset(new QFile(path()));
+    }
+
+    if (!io->open(QIODevice::ReadOnly)) {
+        qWarning() << "[FileInfo] failed to open i/o" << path() << io->errorString();
+        io.clear();
+    }
+
+    return io;
+}
```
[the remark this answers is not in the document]
scrubbbbs: Makes sense. Thanks for suggesting.

ImageLounge/src/DkCore/DkFileInfo.cpp
leejuyuu: You can do 

```cpp
        static constexpr QStringView sZipMarker = u"#/";
```

ImageLounge/src/DkCore/DkFileInfo.cpp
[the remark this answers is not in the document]
leejuyuu: > The null pointer is not possible since `isFromZip` must be false if WITH_QUAZIP is not defined

This isn't obvious without checking the `isFromZip` function. Maybe add a `Q_ASSERT(io)` right below the `#endif`?

leejuyuu:
> > Did someone enable Copilot on this repo?
> 
> I didn't enable it, it showed up so I thought I'd try it out

Please check if it has any associated costs. I'm not sure whether the service is free or not.

ImageLounge/src/DkCore/DkFileInfo.h
```diff
@@ -0,0 +1,215 @@
...
+     *
+     * @note checks the header of files with no suffix, which may get slow
+     *       for large directories.
+     **/
+    static DkFileInfoList readDirectory(const QString &dirPath, const QString &nameFilter = {});
+
+    // todo: clean path before giving fileInfo
+    // remove trailing '/' : breaks fileName()
+    //
+    DkFileInfo(const QString &path = "");
```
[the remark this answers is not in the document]
scrubbbbs: It would be nice to remove it but we do need a default constructor to initialize the shared data pointer.

ImageLounge/src/DkCore/DkFileInfo.h
scrubbbbs: Yes, an interface to better separate the implementations should be added at some point (factory pattern, e.g. the path would determine what type to construct, with virtual methods in place of conditional code).

I don't think the public interface is final so I'm holding off on refactoring until this is all figured out.

ImageLounge/src/DkCore/DkBasicLoader.cpp
```diff
@@ -1129,40 +1129,16 @@ void DkBasicLoader::setHistoryIndex(int idx)
...
-
-#ifdef WITH_QUAZIP
-    if (fi.dir().path().contains(DkZipContainer::zipMarker()))
-        return DkZipContainer::extractImage(DkZipContainer::decodeZipFile(filePath), DkZipContainer::decodeImageFile(filePath));
-#endif
-
-    QFile file(filePath);
-    file.open(QIODevice::ReadOnly);
+    DkFileInfo file(filePath);
+    Q_ASSERT(file.isFile());
```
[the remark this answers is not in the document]
scrubbbbs: It's an assertion because the error is already caught somewhere else

ImageLounge/src/DkCore/DkImageContainer.cpp
```diff
@@ -413,35 +369,46 @@ bool DkImageContainer::saveImage(const QString &filePath, int compression /* = -
...
+    if (fInfo.size() > 100 * 1024 * 1024 && !DkUtils::isLoadable(fInfo)) {
+        qWarning() << "refusing to cache large file:" << fInfo.fileName();
+        return {};
     }
 
-    QFile file(fInfo.absoluteFilePath());
-    file.open(QIODevice::ReadOnly);
+    std::unique_ptr<QIODevice> io = fInfo.getIODevice();
+    if (!io)
+        return {};
```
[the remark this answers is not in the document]
scrubbbbs: Error already reported in getIoDevice

ImageLounge/src/DkCore/DkThumbs.cpp
```diff
@@ -91,36 +92,41 @@ std::optional<LoadThumbnailResult> loadThumbnail(const QString &filePath, LoadTh
...
-#ifdef WITH_QUAZIP
-    if (QFileInfo(filePath).dir().path().contains(DkZipContainer::zipMarker()))
-        ba = DkZipContainer::extractImage(DkZipContainer::decodeZipFile(filePath), DkZipContainer::decodeImageFile(filePath));
-#endif
+    DkFileInfo fileInfo(filePath);
+
+    if (fileInfo.isShortcut() && !fileInfo.resolveShortcut()) {
+        qWarning() << "[Thumbnail] broken shortcut:" << filePath;
+        return std::nullopt;
+    }
```
leejuyuu: The original code resolves path when `isSymLink`, however, this seems to be resolving when `isShortcut()`. Will this fail to catch some of the cases?

From the `QFileInfo` documentation, it says 

```
bool QFileInfo::isSymLink() const
Returns true if this object points to a symbolic link, shortcut, or alias; otherwise returns false.
```

Is seems we should use `isSymLink` to catch all platforms? Is there any reason that we need the more specific `isShortcut`?

ImageLounge/src/DkGui/DkThumbsWidgets.h
```diff
@@ -176,7 +176,7 @@ public slots:
     QMenu *contextMenu;
     QVector<QAction *> contextMenuActions;
 
-    std::vector<QString> mFilePaths{};
+    std::vector<DkFileInfo> mFiles{};
```
leejuyuu, line 179: Is it necessary to store the full `DkFileInfo`? I remember it has heap allocation more than the path string.

ImageLounge/src/DkCore/DkThumbs.cpp
scrubbbbs: The idea was to only resolve things that are not automatically resolved when opening files for reading/writing. This seems to be limited to windows shortcut files (.lnk) and macOS alias files.

ImageLounge/src/DkGui/DkThumbsWidgets.h
scrubbbbs: Yes there is definitely the potential for more memory per item, but only if the shared data detaches. This probably won't happen most of the time.

ImageLounge/src/DkCore/DkThumbs.cpp
scrubbbbs: BTW sorry for the delay I seem to have missed the notification on your latest comments.

ImageLounge/src/DkGui/DkThumbsWidgets.h
leejuyuu: I mean, there's at least one `QFileInfo` and `ZipData` associated with a `DkFileInfo` instance. It is going to add up if someone is browsing through a directory containing tens of thousands of images.

ImageLounge/src/DkCore/DkThumbs.cpp
leejuyuu: > This seems to be limited to windows shortcut files (.lnk) and macOS alias files.

We need to be sure about this. Otherwise, it would be safer to use the original `isSymLink()` call since we know it currently works than risking a regression.

ImageLounge/src/DkGui/DkThumbsWidgets.h
scrubbbbs: It might add  up to some few megabytes at worst, by my estimation.

Everything inside `DkFileInfo` is wrapped in `QSharedData` so it is copy on write. The instance stored to thumbnailer originates from `DkFileInfo::readDirectory()`, and references the same heap memory. 

Beside that, `sizeof(DkFileInfo)` is only 8 bytes, the same as `QString`. When a copy is required, the empty structure is still only 104 bytes (`sizeof(DkFileInfo::SharedData)`).

Compare that to a single thumbnail which at 400px uses at 640,000 bytes...it is a drop in the bucket when you start talking about 1000's of images.

leejuyuu: COW only matter when cloning the same instance, and we are talking about an array of instances originates from different paths.

The numbers you are listing does not include heap allocations that could be referenced inside QFileInfo. The problem is, I could not see the necessity to take more memory, even if it's 100 B for each image, to store the file infos.

ImageLounge/src/DkCore/DkThumbs.cpp
scrubbbbs: `QFileInfo::isSymLink()` only recognizes .lnk and alias besides normal symlinks, confirmed in [Qt sources](https://codebrowser.dev/qt6/qtbase/src/corelib/io/qfileinfo.cpp.html#1143).

There are other types of link files but they were never supported (.url, .desktop, .webloc etc)

ImageLounge/src/DkGui/DkThumbsWidgets.h
scrubbbbs: > COW only matter when cloning the same instance, and we are talking about an array of instances originates from different paths.

The instances come from the image containers, these were allocated when the directory was read. So there isn't any copying into the thumbnails, because of COW feature of QSharedData.

DkFileInfo has to live on the heap somewhere or else the performance will generally be terrible for zip files. 
DkImageContainer seemed like the logical place for this. There is also a small benefit for non-zip files as QFileInfo can cache file metadata.

Constructing a single DkFileInfo from a path is extremely slow for large zip files (sometimes seconds to decompress data etc).

leejuyuu: Ok, I wasn't aware that the DkImageContainer already stores the file info object.

ImageLounge/src/DkCore/DkThumbs.cpp
leejuyuu: However, `isShortCut` does not return true for normal symlink, if I understand the API docs correctly. I am saying that by reducing the set of files for which we call `symLinkTarget()`, we are risking a possible regression.

scrubbbbs: After reviewing usages of `DkFileInfo::isShortcut()`, the only regression would be when a symbolic link and target had a different file extension, with the former being either missing, incorrect, or unsupported, and the latter being supported. Considering that case is technically invalid input anyways, do we have to handle it?

leejuyuu: Yes, because this already exists, I think we should.

scrubbbbs: I have a fix for this, will be pushed soon. I also found this is needed to prevent broken links from appearing in readDirectory().

scrubbbbs:
I reviewed all the commits again, I think we are good.

The most visible change (for non-zip users) is that open directory will now open to thumbnails instead of the first image; this can be changed in general settings to get the old behavior.

@leejuyuu do you have any further concerns?

leejuyuu:
> The most visible change (for non-zip users) is that open directory will now open to thumbnails instead of the first image; this can be changed in general settings to get the old behavior.

What does this mean? Open directory already go to the "Thumbnail Preview" view before this PR. 
Also, I would prefer configurable new behaviors to be opt-in instead of opt-out, unless it is clear that the original behavior is to be discouraged.

scrubbbbs:
> > The most visible change (for non-zip users) is that open directory will now open to thumbnails instead of the first image; this can be changed in general settings to get the old behavior.
> 
> What does this mean? Open directory already go to the "Thumbnail Preview" view before this PR.

Sometimes it did, sometimes it didn't. It depends on how it was opened (explorer panel, drag-drop, file menu, command line, recents, etc). 

> Also, I would prefer configurable new behaviors to be opt-in instead of opt-out, unless it is clear that the original behavior is to be discouraged.

Not really an option since behavior was inconsistent before. I think that thumbnails is the expected behavior for image viewers and so should be the default.

leejuyuu:
I agree that it should be consistent. This is ready for merge. Thanks!

scrubbbbs:
@novomesk, this branch seems to require a merge commit, I guess because it started before plugins were merged. Can you enable them again?

PR #1343, opened 2025-06-14
Drop Qt5

leejuyuu:


ImageLounge/CMakeLists.txt
```diff
@@ -33,8 +33,6 @@ find_package(Python QUIET)
 # include macros needed
 include("cmake/Utils.cmake")
 
-set(QT5_MIN_VERSION 5.15.0)
-
 # different compile options
```
scrubbbbs, line 36: We should set this to 6.0.x? Currently will not produce a configuration error if old build recipe is used: `-D QT_VERSION_MAJOR=5`

ImageLounge/cmake/Unix.cmake
```diff
@@ -118,19 +118,7 @@ if(ENABLE_QUAZIP)
...
-                # ater the final migration to 1.x in Linux distros
-                SET(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
-
-                find_package(QuaZip5)
-                if(NOT QUAZIP_FOUND)
-                    message(FATAL_ERROR "QUAZIP: even 0.x not found. It's mandatory when used with ENABLE_QUAZIP enabled, you can also disable USE_SYSTEM_QUAZIP")
-                endif()
-
-                add_definitions(-DWITH_QUAZIP)
-            endif (QuaZip-Qt${QT_VERSION_MAJOR}_FOUND)
```
scrubbbbs, line 133: We should produce an error if quazip is not found, as-is will show quazip is enabled in the cmake summary but it is not actually enabled.

ImageLounge/CMakeLists.txt
leejuyuu: I've made the configuration fail if `QT_VERSION_MAJOR` is set to a value other than 6.

ImageLounge/cmake/Unix.cmake
leejuyuu: A fatal error has been added here.

scrubbbbs (approved):
Looks good now, thanks!

PR #1347, opened 2025-06-20
Fix deprecations since Qt 6.0

### LINKED ISSUE #1342: Drop support for Qt5
I'd suggest to finish Qt5 support in nomacs.

That would allow to focus resolving compile warnings about obsolete code when Qt6 is used.

I think nomacs 3.20 could be Qt6 only.

As a first step, I suggest the Qt5 workflows to be removed. @leejuyuu What do you think?

leejuyuu:
Closes #1342


scrubbbbs:
Are we targeting a minimum Qt6 version? There remain ~10 deprecations against 6.9.0.

leejuyuu:
Minimum 6.0. We probably should not narrow valid version range too much in one step. Those you see (e.g. `QImage::mirrored()`) requires some newer features, and I am not planning to add conditionals for them at this moment.

scrubbbbs (approved):
Looks good, thanks!

PR #1350, opened 2025-06-28
Improve plugins loading on Unix targets

scrubbbbs:
Fixes a problem I have loading plugins. Plugins were not loading after I did a "make install" of nomacs.

I have an unusual setup:
- nomacs is installed to /usr/local
- qt is installed to /opt/qt/... (usually the latest Qt version I compiled myself)
- the QLibrary::path(QtLibraries) gives /opt/qt/... but plugins were installed to /usr/local, so it doesn't find them

This patch allows this to work and also should be more reliable since it first looks in cmake install path.

Note the flatpak is ~now probably broken~ currently already broken:
- plugins installed to `/app/lib/nomacs-plugins` but nomacs doesn't look there

### DIFF
--- a/ImageLounge/src/DkCore/DkPluginManager.cpp
+++ b/ImageLounge/src/DkCore/DkPluginManager.cpp
@@ -1152,8 +1152,9 @@ void DkPluginManager::loadPlugins()
 
         for (const QString &fileName : pluginsDir.entryList(QDir::Files)) {
             DkTimer dtt;
-#ifdef Q_OS_LINUX
-            // needed because of symbolic links of sonames
+#ifdef Q_OS_UNIX
+            // symlinks to libraries will create duplicates
+            // FIXME: still possible to get duplicates if nomacs version changes
             QFileInfo file(pluginsDir.absoluteFilePath(fileName));
             if (file.isSymLink())
                 continue;


leejuyuu:
Renaming the directory might break existing installations. I think we should avoid that.

scrubbbbs:
> Renaming the directory might break existing installations. I think we should avoid that.

The install path has not changed, only the build path. The install path remains the same, `<cmake_prefix>/lib<lib-suffix>/nomacs-plugins`.

scrubbbbs:
Tested on ubuntu, arch, freebsd, and haiku. ~MacOS should be fine, I cannot test at the moment as my VM is broken again.~ MacOS is working as well.

scrubbbbs:
Confirmed this also fixes flatpak plugin loading. I had to make some change to the flatpak manifest but none related to this PR.

Update: [diff for flatpak manifest](https://github.com/flathub/org.nomacs.ImageLounge/commit/870b09b1c0c3c4501b007b259e59a0d618e8757c)

novomesk:
Regarding flatpak build, we will wait till the next release.

ImageLounge/src/DkCore/DkPluginManager.cpp
```diff
@@ -1307,7 +1309,36 @@ void DkPluginManager::createPluginsPath()
...
+    // Also include app dir path since it is useful for portable builds or testing.
+    // Note that qt.conf could be manipulated for appImage, snap, flatpak
+    // in case this is insufficient.
+    //
+    QStringList libPaths;
+    libPaths += qApp->applicationDirPath();
+    libPaths += NOMACS_INSTALL_LIBPATH;
+    libPaths += QLibraryInfo::path(QLibraryInfo::LibrariesPath);
+
+    qInfo() << "plugins search path:" << libPaths;
```
leejuyuu: This should probably be a debug log.

PR #1352, opened 2025-07-06
Suport multiple files in tabs

maurolacy:
Fixes handling of multiple files in the `--tab` / `-t` option.


scrubbbbs:
Thanks for contributing! This is somewhat fixed in #1314, `--tab` and `--dir` have been removed and there is no special argument needed to open multiple files or directories. You can start from that branch to do further work on tabs.

scrubbbbs:
Can this be closed now?

PR #1357, opened 2025-07-20
Show edited status when changing ratings

leejuyuu:


leejuyuu:
Thanks for testing!


> 2. Rating change is shown multiple times in edit history

This is expected when setting to different values. I'll fix the case when setting to the same value repeatedly. Do you think changing the text to "Set rating to 5" makes them clearer?

I think others are separate bugs.

scrubbbbs:
> > 2. Rating change is shown multiple times in edit history
> 
> This is expected when setting to different values. I'll fix the case when setting to the same value repeatedly. Do you think changing the text to "Set rating to 5" makes them clearer?

Yes. Also, the history label doesn't need to start with a verb, it could be shortened to "Rating 5" with "Remove rating" or "Clear rating" for the 0 case.

leejuyuu:
I fixed the duplicated history entry. However, this does not work well for non-existing EXIF data for now.

I feel that starting the history item with a verb is more natural, as it corresponds to an action.

PR #1364, opened 2025-07-24
Fix a few pseudocolor bugs

scrubbbbs:
Fixed pseudocolor mode was not being setup correctly (effectively always disabled) there were a few issues with the toolbar state.


leejuyuu:
This fixed the pseudocolor bug. Thanks!

PR #1369, opened 2025-07-28
Formatting

scrubbbbs:
- [x] dos2unix (crlf->lf)
- [x] clang-format-18 (w/current .clang-format)
- [x] format cmake files with gersemi
- [x] workflow check for cmake files
- [x] .clang-format modifications (column limit => 120, etc)
- [x] clang-format-18
- [x] add scripts used for formatting steps


scrubbbbs:
I think this is ready

ImageLounge/.clang-format
```diff
@@ -1,13 +1,8 @@
 ---
-# SPDX-FileCopyrightText: 2019 Christoph Cullmann <cullmann@kde.org>
-# SPDX-FileCopyrightText: 2019 Gernot Gebhard <gebhard@absint.com>
+# C/C++ style based on KDE ECM configuration:
+# https://invent.kde.org/frameworks/extra-cmake-modules/-/blob/master/kde-modules/clang-format.cmake
 #
-# SPDX-License-Identifier: MIT
```
leejuyuu: Once the copyright text and the license are there, I think we cannot remove them without violating the terms.

When referencing some external code, it is best to use the permalink instead of pointing to the master branch.

scripts/format-cmake.sh
```diff
@@ -0,0 +1,31 @@
...
+
+args="--in-place"
+[ -z "$1" ] || args="$@"
+
+abort() { echo -e "\nformat-cmake: $1\n" >&2; exit 1; }
+
+# don't allow version change without intervention
+version=0.21.0
+~/.gersemi/bin/gersemi --version | head -n1 | grep -v $version
+[ $? -eq 1 ] || abort "gersemi version invalid, required version $version"
```
leejuyuu: The version is already pinned in the CI environment. This is unnecessary.

scripts/format-cmake.sh
```diff
@@ -0,0 +1,31 @@
...
+
+cd ImageLounge || abort "you must run this script from the repository root"
+
+~/.gersemi/bin/gersemi \
```
leejuyuu:
```suggestion
gersemi \
```

The script should only assume that the gersemi command is present and not force a specific path.

.github/workflows/linux.yml
```diff
@@ -79,3 +79,14 @@ jobs:
         run: |
           cd ${GITHUB_WORKSPACE}
           git clang-format --diff ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }} 2>&1
+      
+      - name: Setup gersemi
+        run: |
+          python -m venv ~/.gersemi
+          ~/.gersemi/bin/pip install gersemi==0.21.0
```
leejuyuu:
```suggestion
          python -m venv ~/.gersemi
          source ~/.gersemi/bin/activate
          pip install gersemi==0.21.0
```

scrubbbbs:
@leejuyuu thanks for taking a look, if you have nothing else I want to do a few low-risk/low-effort [clang-tidy fixes](https://clang.llvm.org/extra/clang-tidy/checks/list.html).

I will do each check/fix as a separate commit so they are easier to understand. I will add a `.clang-tidy` so we could add it to CI if we want, but this is a bit different workflow (more like compiling nomacs, but much slower).

Anyways these are the changes I was thinking about
- 0 to nullptr
- use override/virtual properly
- use "=default" "=delete" on constructors
- modernize make_shared/make_unique

leejuyuu:
Sure! I've been using clang-tidy in my editor as well, so it's great to apply some of the suggested fixes. 

If possible, please merge this branch first. I plan to fix some rating things during the weekend.

Issue #1359, opened 2025-07-21
Edited image does not reload properly after go to next in directories containing single file


leejuyuu:
1) if there is one image in the directory:
- change rating
- go to next image (which is the same image)
- click "Yes" on the changes dialog
- modified state is not cleared, however the file is saved

_Originally posted by @scrubbbbs in https://github.com/nomacs/nomacs/pull/1357#pullrequestreview-3038149284_

We also need to test whether other edited state are also affected.

leejuyuu:
For the "No" case, the wrong behavior appears when you do this twice. Other editing, such as cropping, is also affected.

PR #1374, opened 2025-08-02
Comment and ratings

leejuyuu:


leejuyuu:
@scrubbbbs I could not reproduce https://github.com/nomacs/nomacs/issues/1359. Would you mind testing again on this branch?

scrubbbbs:
#1359 seems fixed. The clicking "Yes" condition works, but the "No" condition does not (did it work before?). 

If there is one image in the directory, after any modification, clicking "No" leaves you stuck on that image (it should reload the image, effectively, but does not).

Another maybe new bug, when selecting the crop tool, the rating panel hides (expected). But when leaving crop tool, it does not reappear. It seems to try to, I can see it animate in partially and then it hides. Comes from `DkControlWidget.cpp:555`, maybe can be removed now with the other fixes you've made.

I also having issue with the clear rating, it doesn't seem to work correctly with history:
- clear rating + control-Z does not revert to previous rating
- it seems the initial rating can never be restored by pressing control-Z, it stops at the first rating assigned

 Another test with odd behavior:
- set rating 0 and save, reload image
- set rating in sequence 1..5 works as expected
- set rating 0, rating displays 2 stars, saving shows 5 stars
- if I keep pressing 0 then it cycles through different stars until I eventually get to 0

leejuyuu:
> Another maybe new bug, when selecting the crop tool, the rating panel hides (expected). But when leaving crop tool, it does not reappear. It seems to try to, I can see it animate in partially and then it hides. Comes from DkControlWidget.cpp:555, maybe can be removed now with the other fixes you've made.

This seems to be introduced by the new fading mixin. The `setVisible()` method is no longer synchronous, so the `showFileInfo(true)` call is a no-op because `mFileInfoLabel` is not visible yet. It would work if I change the code to the following.

```cpp
        mFileInfoLabel->setFadeEnabled(false);
        mFileInfoLabel->setVisible(false);
        mFileInfoLabel->setFadeEnabled(true);
        showFileInfo(true);
```

I am going to remove this block since the animation is not really necessary because the crop widget is fading away as well. However, I think we might want to reconsider overriding the synchronous `setVisible()` method. The current behavior is surprising. The asynchronous one that associates with the fading animation should probably be indicated in the method name.

scrubbbbs:
`setVisible()` on fade widgets was always async, e.g. `QWidget::isVisible()` would give the wrong answer WRT what the animation was doing. There was no override of `isVisible()` in the old implementation to try to prevent this.

Perhaps `isVisible()` `isHidden()` etc should be deleted functions in `DkFadeMixin` subclasses, this would catch most of these usages, and we could go fix them with `isHiding()` or `isShowing()` etc. Remaining usages would have to be found some other way (e.g. a vector of qwidget would bypass the deleted function).

I know of at least one other glitch due to this, when hiding a widget, pressing the hotkey twice rapidly does not re-show the widget, and the third keypress is ignored.

leejuyuu:
> setVisible() on fade widgets was always async

I think this might be a bad design. Currently, the caller needs to know that it is async to use it properly, so it would be better to make a separate function and indicate this in the name (e.g. setVisibleWithFade).

leejuyuu:
> If there is one image in the directory, after any modification, clicking "No" leaves you stuck on that image (it should reload the image, effectively, but does not).

This is more universal than I thought. Other editing, such as crop, is also affected. I am going to skip it for now so that this branch can be merged first.

scrubbbbs:
> > setVisible() on fade widgets was always async
> 
> I think this might be a bad design. Currently, the caller needs to know that it is async to use it properly, so it would be better to make a separate function and indicate this in the name (e.g. setVisibleWithFade).

It seems that way but maybe there was a reason for it. If you want to make a universal fade in/out for widgets, overriding `setVisible()` will allow this to happen even when another widget or view is doing this, for example `QStackedWidget` (IIRC there is a case where this is used).

PR #1375, opened 2025-08-02
Clang-Tidy

scrubbbbs:
Setup clang-tidy and do a few easy/low-risk fixes (e.g. no chance of behavior/semantic change)

### DIFF
--- a/ImageLounge/plugins/AffineTransformations/src/DkSkewEstimator.h
+++ b/ImageLounge/plugins/AffineTransformations/src/DkSkewEstimator.h
@@ -64,7 +64,7 @@ public:
         dir_end,
     };
 
-    DkSkewEstimator(QWidget *mainWin = 0);
+    DkSkewEstimator(QWidget *mainWin = nullptr);
     ~DkSkewEstimator();
 
     double getSkewAngle();
--- a/ImageLounge/src/DkCore/DkMetaData.cpp
+++ b/ImageLounge/src/DkCore/DkMetaData.cpp
@@ -65,7 +65,7 @@ QSharedPointer<DkMetaDataT> DkMetaDataT::copy() const
     metaDataN->mFileInfo = mFileInfo;
     metaDataN->mExifState = mExifState;
 
-    if (mExifImg.get() != 0) {
+    if (mExifImg.get() != nullptr) {
         // ImageFactory::create(type) may crash even if old Image object has that type
         try {
             // Load new Exiv2::Image object
@@ -129,7 +129,7 @@ void DkMetaDataT::readMetaData(const DkFileInfo &file, QSharedPointer<QByteArray
         return;
     }
 
-    if (mExifImg.get() == 0) {
+    if (mExifImg.get() == nullptr) {
         qDebug() << "[Exiv2] image could not be opened for exif data extraction";
         return;
     }
@@ -249,7 +249,7 @@ bool DkMetaDataT::saveMetaData(QSharedPointer<QByteArray> &ba, bool force)
         return false;
     }
 
-    if (exifImgN.get() == 0) {
+    if (exifImgN.get() == nullptr) {
         qDebug() << "image could not be opened for exif data extraction";
         return false;
     }
@@ -1031,7 +1031,7 @@ void DkMetaDataT::setThumbnail(QImage thumb)
             // whipe all exif data of the thumbnail
             auto exifImgThumb = Exiv2::ImageFactory::open(reinterpret_cast<const byte *>(ba.constData()), ba.size());
 
-            if (exifImgThumb.get() != 0 && exifImgThumb->good())
+            if (exifImgThumb.get() != nullptr && exifImgThumb->good())
                 exifImgThumb->clearExifData();
         } catch (...) {
             qDebug() << "could not clear the thumbnail exif info";
--- a/ImageLounge/src/DkCore/DkProcess.h
+++ b/ImageLounge/src/DkCore/DkProcess.h
@@ -69,6 +69,15 @@ class DllCoreExport DkAbstractBatch
 public:
     DkAbstractBatch() = default;
 
+    // ok, this is important:
+    // we are using the abstract class to process specialized items
+    // if we allow copy operations - we get slicing issues
+    // so we just allow pointers to the batch processing functions.
+    // but pointers result in threading issues - so we just use
+    // QSharedPointers -> problem solved : )
+    DkAbstractBatch(const DkAbstractBatch &o) = delete;
+    DkAbstractBatch &operator=(const DkAbstractBatch &o) = delete;
+
     virtual void setProperties(...) {};
     virtual void saveSettings(QSettings &) const {};
     virtual void loadSettings(QSettings &) {};
@@ -94,16 +103,6 @@ public:
     QString settingsName() const;
 
     static QSharedPointer<DkAbstractBatch> createFromName(const QString &settingsName);
-
-private:
-    // ok, this is important:
-    // we are using the abstract class to process specialized items
-    // if we allow copy operations - we get slicing issues
-    // so we just allow pointers to the batch processing functions.
-    // but pointers result in threading issues - so we just use
-    // QSharedPointers -> problem solved : )
-    DkAbstractBatch(const DkAbstractBatch &o);
-    DkAbstractBatch &operator=(const DkAbstractBatch &o);
 };
 
 #ifdef WITH_PLUGINS
--- a/ImageLounge/src/DkCore/DkProcess.h
+++ b/ImageLounge/src/DkCore/DkProcess.h
@@ -340,7 +340,7 @@ public:
         batch_item_end
     };
 
-    DkBatchProcessing(const DkBatchConfig &config = DkBatchConfig(), QWidget *parent = 0);
+    DkBatchProcessing(const DkBatchConfig &config = DkBatchConfig(), QWidget *parent = nullptr);
 
     void compute();
     static bool computeItem(DkBatchProcess &item);
--- a/ImageLounge/src/DkCore/DkProcess.h
+++ b/ImageLounge/src/DkCore/DkProcess.h
@@ -67,7 +67,7 @@ class DkMetaDataT;
 class DllCoreExport DkAbstractBatch
 {
 public:
-    DkAbstractBatch(){};
+    DkAbstractBatch() = default;
 
     virtual void setProperties(...) {};
     virtual void saveSettings(QSettings &) const {};
@@ -267,7 +267,7 @@ protected:
 class DllCoreExport DkBatchConfig
 {
 public:
-    DkBatchConfig(){};
+    DkBatchConfig() = default;
     DkBatchConfig(const QStringList &fileList, const QString &outputDir, const QString &fileNamePattern);
 
     virtual void saveSettings(QSettings &settings) const;
--- a/ImageLounge/src/DkCore/DkProcess.h
+++ b/ImageLounge/src/DkCore/DkProcess.h
@@ -111,18 +111,18 @@ class DllCoreExport DkPluginBatch : public DkAbstractBatch
 public:
     DkPluginBatch();
 
-    virtual void saveSettings(QSettings &settings) const override;
-    virtual void loadSettings(QSettings &settings) override;
+    void saveSettings(QSettings &settings) const override;
+    void loadSettings(QSettings &settings) override;
 
     virtual void preLoad();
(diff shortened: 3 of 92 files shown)


ImageLounge/src/DkCore/DkProcess.h
```diff
@@ -398,7 +398,7 @@ class DllCoreExport DkBatchProfile
     static QString defaultProfilePath();
     static QString profileNameToPath(const QString &profileName);
     static QString makeUserFriendly(const QString &profilePath);
-    static QString extension();
+    static QString fileSuffix();
```
leejuyuu: Instead of a method, I suggest making this a static string view

```cpp
    static constexpr std::string_view sFileSuffix = "pnm";
```

ImageLounge/src/DkGui/DkLogWidget.cpp
```diff
@@ -40,7 +40,12 @@ related links:
 namespace nmc
 {
 
-QSharedPointer<DkMessageQueuer> msgQueuer = QSharedPointer<DkMessageQueuer>();
+static DkMessageQueuer *getMsgQueuer()
+{
+    // avoid static destruction, this is globally referenced via qDebug() etc
+    static auto *querer = new DkMessageQueuer;
+    return querer;
+}
```
leejuyuu: Can we make this be owned by DkLogWidget? It seems to be the only thing that uses this.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -311,7 +311,7 @@ void DkFilePreview::drawThumbs(QPainter *painter)
     // mouse over effect
     QPoint p = worldMatrix.inverted().map(mapFromGlobal(QCursor::pos()));
 
-    for (int idx = 0; idx < mFiles.size(); idx++) {
+    for (int idx = 0; (unsigned int)idx < mFiles.size(); idx++) {
```
leejuyuu: Avoid C-style casting. Also, we should use size type instead of unsigned int.

```suggestion
    for (int idx = 0; static_cast<qsizetype>(idx) < mFiles.size(); idx++) {
```

I think the cast is not strictly necessary. Size types always cover the range of positive signed integers.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -602,7 +602,7 @@ void DkFilePreview::mouseMoveEvent(QMouseEvent *event)
             if (worldMatrix.mapRect(thumbRects.at(idx)).contains(event->pos())) {
                 selected = idx;
 
-                if (selected < mFiles.size() && selected >= 0) {
+                if (selected >= 0 && (unsigned int)selected < mFiles.size()) {
```
leejuyuu: See previous comment.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -839,7 +839,7 @@ void DkFilePreview::setFileInfo(QSharedPointer<DkImageContainerT> cImage)
 
     int tIdx = -1;
 
-    for (int idx = 0; idx < mFiles.size(); idx++) {
+    for (int idx = 0; (unsigned int)idx < mFiles.size(); idx++) {
```
leejuyuu: See previous comment.

ImageLounge/src/DkGui/DkWidgets.cpp
```diff
@@ -238,7 +238,7 @@ void DkThumbsSaver::thumbLoaded()
     mNumSaved++;
     emit numFilesSignal(mNumSaved);
 
-    if (mNumSaved == mWatchers.size() || mStop) {
+    if ((unsigned int)mNumSaved == mWatchers.size() || mStop) {
```
leejuyuu: See previous comment.

ImageLounge/src/DkCore/DkMetaData.cpp
```diff
@@ -778,7 +778,7 @@ QImage DkMetaDataT::getPreviewImage(int minPreviewWidth) const
         for (size_t idx = 0; idx < pList.size(); idx++) {
             if (pList[idx].width_ > (uint32_t)maxWidth && pList[idx].width_ > (uint32_t)minPreviewWidth) {
                 mIdx = (int)idx;
-                maxWidth = pList[idx].width_;
+                maxWidth = (int)pList[idx].width_;
```
leejuyuu: Avoid C-style casting.

ImageLounge/plugins/AffineTransformations/src/DkSkewEstimator.h
```diff
@@ -70,22 +70,23 @@ class DkSkewEstimator
...
-    QVector<int> selectedLineTypes;
-    cv::Mat matImg;
-    int rotationFactor;
-    QProgressDialog *progress;
-    QWidget *mainWin;
+    static constexpr int nIter = 200;
+    static constexpr double mSigma = 0.3;
+    static constexpr double mSepThr = 0.1;
+    static constexpr int mEpsilon = 2;
+    static constexpr int mKMax = 7;
```
leejuyuu: Suggest prefixing static member with "s" instead of "m".

CONTRIBUTING.md
```diff
@@ -1,36 +1,80 @@
...
+  - If you have fixed an issue, add `Fixes: #<num>` in the body, on one line.
 
-## Style
+## Code Style
 
 nomacs might have a weird coding style at the first glance, however we try to follow at least some rules:
 
-- `using namespace` is prohibited (even for `std`)
-- Naming
+- **Naming**
```
leejuyuu: Suggest making these <h3>. Also applies to "Formatting", etc. You'll likely need to decrease the indent of the bullets by two spaces.

```suggestion
### Naming
```

CONTRIBUTING.md
```diff
@@ -1,36 +1,80 @@
...
+
+- **Issue Triage**
+  - Confirm if latest version fixes the issue
+  - Request or provide additional information, test files, screenshots etc
+  - Provide workarounds
+  - Test on different platforms
+  - Find duplicates of other issues
+    - => may be tagged with `Duplicate: #<issue>` in and closed.
+  - Tag related issues
+    - => may be tagged with `Related: #<issue>`
```
leejuyuu:
```suggestion
  - Find duplicates of other issues
     You can comment `Duplicate: #<issue>` in the newer issue. We would then close them.
  - Tag related issues
      You can comment `Related: #<issue>` to help us group them.

File: CONTRIBUTING.md [hunk 11]
```diff
@@ -1,36 +1,80 @@
...
+  
+- **Testing**
+  - Try to break things
+  - Compile nomacs and test the latest changes in-between releases
+  
+- **Documentation**
+    - Add or update FAQ (search issues for ideas)
+    - Document features (sync, batch, etc)
+    - Describe workflows, use cases
```
leejuyuu: Extra indentation.

CONTRIBUTING.md
```diff
@@ -1,36 +1,80 @@
...
+- Changed code should compile without warnings.
 - Push to your fork and submit a pull request.
 
-We will comment on pull requests within one week. Here are some things that increase the chances of your PR being accepted:
+We will usually comment on pull requests within one week. Here are some things that increase the chances of your PR being accepted:
 
-- Write good commit messages.
-- Follow our style guide.
+- Do not break another feature related to/dependent on your changes.
+- Do not break another platform, aim to avoid platform-specific features.
```
leejuyuu: I am trying a softer tone.

```suggestion
- Please try to test areas that the PR is changing or that depend on your changes.
- We support Windows, Linux, and macOS. If possible, please try to test on different platforms.
- We tend to avoid platform-specific features. Please discuss with us beforehand.
```

CONTRIBUTING.md
```diff
@@ -1,36 +1,80 @@
...
+- **Naming**
   - Class names start with a capital letter (i.e. `DkBaseManipulator`).
   - Methods start with lowercase (i.e. `name()`).
   - Prefix members with an m (i.e. `int mAction = 0`).
   - Prefix classes with `Dk`.
-- format C++ sources with `clang-format -i --style=file your_contribution.cpp`
-- Encapsulate everything with the namespace `nmc`
+- **Formatting**
+  - Format C++ sources with `scripts/format-cpp.sh` (or configure editor to use clang-format)
+  - Format cmake sources with `scripts/format-cmake.sh`
```
leejuyuu:
```suggestion
  - We format C++ sources `clang-format`. You can run `scripts/format-cpp.sh` or configure your editor to use `clang-format`.
  - We format cmake sources with `gersemi`. You can run `scripts/format-cmake.sh`.
  ```

File: CONTRIBUTING.md [hunk 14]
```diff
@@ -1,36 +1,80 @@
...
+- **C++ guidelines** 
+  - Standard: C++17
+  - Encapsulate everything with the namespace `nmc`
+  - `using namespace` is prohibited (even for `std`) for portability
+  - Avoid using `SIGNAL` and `SLOT` macros in favor of function pointers.
+  They enable compile time check of the signal and slot names and signature.
+  - Avoid `auto` unless the type name appears on the same line. This helps with readability.
+    - Exception: loop variables: `for(const auto& x : container)`
+    - Exception: iterators (but prefer the new loop syntax)
+    - Always use cv/pointer/reference specifiers: `auto *var`, `auto &var` etc
```
leejuyuu: Not sure what this means. (L76)

ImageLounge/src/DkGui/DkLogWidget.cpp
scrubbbbs: Sure, so long as it is a pointer to prevent static destruction.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
scrubbbbs: The c-cast is an old habit, thanks for noticing. We can enforce it in clang-tidy but IIRC there are 800-900 instances to change.

It is necessary to silence a warning - compare between signed/unsigned values. If `idx` was ever set to negative in the loop body then the loop would terminate early. The cast formally states the writer doesn't think this is possible and silences the warning.

CONTRIBUTING.md
scrubbbbs: In otherwords, never drop the `&`, `*`, `const`, or `volatile` when using auto. There is a clang-tidy check for this.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
leejuyuu: > We can enforce it in clang-tidy but IIRC there are 800-900 instances to change.

There's no need to do this right away. We can avoid it from increasing for now.

> The cast formally states the writer doesn't think this is possible and silences the warning.

Agreed.

CONTRIBUTING.md
leejuyuu: Can you point me to the specific rule? I would like to know the rationale behind this.

I suggesting changing the text to your comment.

```suggestion
    - Use `auto` only to replace the type name. This means the `&`, `*`, or qualifiers (e.g. `const` and `volatile`) should always be included. For  example,
    
       ```cpp
       // Do write this
       auto *foo = new Foo(); 
       // instead of 
       auto foo = new Foo(); 
       ```
```

scrubbbbs: I must have remembered this wrong, c++ guidelines don't say anything about it, but  `moderize-use-auto` has a `RemoveStars` option (default disabled). But it does not warn if it missing, only keeps it in the suggested fix. So not technically an enforced rule but I think it is a good idea for readability.

leejuyuu: Ok, we can still keep it in the CONTRIBUTING.md.

ImageLounge/plugins/AffineTransformations/src/DkSkewEstimator.h
scrubbbbs: Maybe we can we decide on a naming rule for constants, and use the "s" prefix for true statics. There is a mix throughout nomacs. I think either `ALL_CAPS` or `kCamelCase` would be best. Qt uses `TitleCase` but that mixes up with classes/typenames somewhat. `all_lower_case` is another pattern (also often used for enums)

leejuyuu: I like kCamelCase for members. Looks more consistent.

PR #1380, opened 2025-08-10
Compute thumb dialog text

leejuyuu:


scrubbbbs:
Thanks. This has merge conflicts with #1375, I'd like to merge that one first since it will be easier

PR #1382, opened 2025-08-19
Support HiDPI Displays

scrubbbbs:
Retrying #1102. The approach is the same, it seems that dividing the image size used to form the viewport transformation by `devicePixelRatioF()` gives the simplest solution. This appears it will work on all window managers.

- [x] Fix device scaling in the main viewport.
- [x] Fix cursor position / color under cursor
- [x] Fix compression and resize dialogs
- [x] Fix overview
- [x] Fix crop
- [x] Fix icons
- [x] Fix paint plugin
- [x] Fix transform plugin
- [x] Fix dragging window between monitors with different scales
- [x] Fix dynamically changing device scale w/o restarting nomacs

### DIFF
--- a/ImageLounge/src/DkCore/DkUtils.cpp
+++ b/ImageLounge/src/DkCore/DkUtils.cpp
@@ -574,8 +574,9 @@ QString DkUtils::getBuildInfo()
     info += QSysInfo::prettyProductName() + ", " + memory + "\n";
 
     // library versions (dynamic version where possible)
+    QWidget *window = qApp->activeWindow();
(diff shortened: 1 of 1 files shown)


scrubbbbs:
The parts completed (see checklist) seem to be stable and are ready for review.

samyak-jain:
This overall looks great on my system. One thing I noticed is that the rotations under the Edit menu no longer work in this build.

scrubbbbs:
> This overall looks great on my system. One thing I noticed is that the rotations under the Edit menu no longer work in this build.

Thanks, I think we have a regression prior to this PR. The rotation currently won't work unless the image has metadata.

PR #1384, opened 2025-08-21
feat: File based viewer

maurolacy:
Changes the central widget to work with a list of files instead of a directory or directories.

This is handy for accepting a list of files from the command line. Likely, through a shell script wrapper like

```
$ cat /usr/local/bin/nomacs
#!/bin/bash

QT_LOGGING_RULES="*.info=false" /Applications/nomacs.app/Contents/MacOS/nomacs "$@"
$ 
```

### DIFF
--- a/ImageLounge/src/DkGui/DkControlWidget.cpp
+++ b/ImageLounge/src/DkGui/DkControlWidget.cpp
@@ -290,9 +290,7 @@ void DkControlWidget::connectWidgets()
     // playing
     connect(mPlayer, &DkPlayer::previousSignal, mViewport, &DkViewPort::loadPrevFileFast);
     connect(mPlayer, &DkPlayer::nextSignal, mViewport, &DkViewPort::loadNextFileFast);
-    connect(mPlayer, &DkPlayer::loadFileSignal, mViewport, [this](const QString &filePath) {
-        mViewport->loadFile(filePath);
-    });
+    connect(mPlayer, &DkPlayer::loadFileSignal, mViewport, qOverload<const QString &>(&DkViewPort::loadFile));
 
     // cropping
     connect(mCropWidget, &DkCropWidget::cropImageSignal, mViewport, &DkViewPort::cropImage);


ImageLounge/src/DkGui/DkWidgets.cpp
```diff
@@ -1365,6 +1400,46 @@ void DkPlayer::showTemporarily(bool autoHide)
     DkFadeWidget::show(false);
 }
 
+void DkPlayer::setFileList(const QStringList& files)
+{
+    mFileList = files;
+    mCurrentIndex = 0;
+}
+
+void DkPlayer::setCurrentFile(const QString& filePath)
```
scrubbbbs: These accessors are never used, you may omit them.  Edit: wrong line: the ones below getFileList etc. You may add simple functions to the header file.

scrubbbbs:
Thanks for this contribution. I just had the one comment regarding unused code. 

Since you have scoped it to slideshow only I guess we don't have to be concerned with breaking other features. I will try it out tomorrow to see how it works.

maurolacy:
This is working fine now AFAIK. It still has some side effects, but since they are present in the original version as well, I think it's fine.

At some point it would be good to switch entirely to file-based mode / behaviour. But, not in this PR / at this stage IMO.

ImageLounge/src/main.cpp
```diff
@@ -334,11 +335,33 @@ int main(int argc, char *argv[])
...
+            // Expand directories into file lists and collect all image files
+            QStringList allImageFiles;
+            for (const QString &path : slideshowFiles) {
+                nmc::DkFileInfo fileInfo(path);
+                if (fileInfo.isDir()) {
+                    // Directory: expand to image files
+                    QDir dir(path);
+                    QStringList filters = nmc::DkSettingsManager::param().app().fileFilters;
+                    QStringList imageFiles = dir.entryList(filters, QDir::Files, QDir::Name);
+                    for (const QString &imageFile : imageFiles) {
```
scrubbbbs: You can support zip files and future file system wrappers by using DkFileInfo::readDirectory() instead of QDir()::entryList(), it already filters for supported files.

ImageLounge/src/main.cpp
```diff
@@ -329,30 +330,66 @@ int main(int argc, char *argv[])
...
+                        allImageFiles.append(dir.filePath(imageFile));
+                    }
+                } else {
+                    // Single file: add directly
+                    allImageFiles.append(path);
```
scrubbbbs: Build the list with `fileInfo.path()` here to support filesystem wrapper (could be different from path passed on command line).

maurolacy:
A little bit overkill IMO, but OK with me. I would prefer to fully switch to file-based, for all modes.

OK, I'll add this flag, and implement the full switch based / depending on it.

ImageLounge/src/main.cpp
```diff
@@ -341,11 +342,9 @@ int main(int argc, char *argv[])
                 nmc::DkFileInfo fileInfo(path);
                 if (fileInfo.isDir()) {
                     // Directory: expand to image files
-                    QDir dir(path);
-                    QStringList filters = nmc::DkSettingsManager::param().app().fileFilters;
-                    QStringList imageFiles = dir.entryList(filters, QDir::Files, QDir::Name);
-                    for (const QString &imageFile : imageFiles) {
-                        allImageFiles.append(dir.filePath(imageFile));
+                    nmc::DkFileInfoList imageFiles = nmc::DkFileInfo::readDirectory(fileInfo.dirPath(), "");
```
scrubbbbs: You need `fileInfo.path()` here as you already have a directory, `dirPath()` will get the parent directory.

ImageLounge/src/main.cpp
```diff
@@ -331,54 +338,104 @@ int main(int argc, char *argv[])
...
+            } else {
+                // Single file: add directly
+                allImageFiles.append(path);
+            }
+        }
+        return allImageFiles;
+    };
+
     // Handle slideshow with file list support
     if (parser.isSet(slideshowOpt)) {
```
scrubbbbs: There is a bit of duplicated code here, I think most of it goes away if you have this structure:

```c++
QStringList fileList = parser.positionalArguments();
if (parser.isSet(fileListOpt))
    fileList = expandFileList(fileList);

if (slideshowOpt && filelistOpt)
{
// new slideshow file loading 
}
else
{
// default file loading

if (slideShowOpt)
   startSlideShow();
}
```

ImageLounge/src/DkGui/DkCentralWidget.cpp
```diff
@@ -1148,32 +1148,97 @@ void DkCentralWidget::load(const QString &path)
...
     if (fileInfo.isDir()) {
-        if (!loader->loadDir(fileInfo.path())) {
-            setInfo(tr("I could not load \"%1\"").arg(path));
-            return;
-        }
+        // Check if we're in file-list mode (has an active file list)
+        bool hasActiveFileList = hasViewPort() &&
+                                 getViewPort()->getController()->getPlayer() &&
+                                 getViewPort()->getController()->getPlayer()->hasFileList();
+
```
scrubbbbs: If you want to take this beyond slideshow from command line, I think we need an abstraction to make it maintainable. It is already quite difficult to debug the directory mode without adding file-lists to the mix.

Directory mode could be viewed as a special case of file-list mode, and we can figure out what a common interface needs to look like to avoid all the special cases sprinkled throughout nomacs. For example there could be a class called `DkFileList` that encapsulates the differences between these modes. Some things it would need to handle to take care of both modes:

- update the list when something changes (file deletion, renaming file, file changed on disk, directory contents changed).
- jumping to the begin/end
- jumping to some arbitrary offset (e.g. page up/page down function)
- looping
- recursive-scan behavior -- if prev/next/jump hits the end of a directory in dir mode, it looks for the prev/next sibling in the parent/starting directory. This is one of the weird things nomacs does that won't apply to filelist mode.
- sorting function (maybe file-list mode is excluded from sorting if the order is important)

maurolacy: Agreed, this is too complicated already. Nevertheless, it's working fine AFAICT.

Will think about a proper abstraction / clean-up around all this functionality.

maurolacy: In the mean time, if you think this has merit, we can perhaps reset this PR to commit a7563f0e. Which deals only with a list of files for slideshow mode.

And then work on a file list container, and generalising file list behaviour / mode, in another PR.

scrubbbbs: I like that idea, reduces churn and that should be ready to merge.

maurolacy: Done.

scrubbbbs:
I think this is ready to merge unless @leejuyuu  has any comments.

ImageLounge/src/DkGui/DkControlWidget.cpp
```diff
@@ -290,6 +290,9 @@ void DkControlWidget::connectWidgets()
     // playing
     connect(mPlayer, &DkPlayer::previousSignal, mViewport, &DkViewPort::loadPrevFileFast);
     connect(mPlayer, &DkPlayer::nextSignal, mViewport, &DkViewPort::loadNextFileFast);
+    connect(mPlayer, &DkPlayer::loadFileSignal, mViewport, [this](const QString &filePath) {
+        mViewport->loadFile(filePath);
+    });
```
leejuyuu:
```suggestion
    connect(mPlayer, &DkPlayer::loadFileSignal, mViewport, qOverload<const QString &>(&DkViewPort::loadFile));
 ```

File: ImageLounge/src/DkGui/DkWidgets.h [hunk 8]
```diff
@@ -182,6 +188,10 @@ public slots:
     QPushButton *nextButton;
     QPushButton *playButton;
     QWidget *container;
+
+    // File list support
+    QStringList mFileList;
+    int mCurrentIndex;
```
leejuyuu: It's safer to initialize int here.
```suggestion
    int mCurrentIndex = 0;
```

ImageLounge/src/main.cpp
```diff
@@ -114,6 +119,10 @@ int main(int argc, char *argv[])
     QCommandLineOption slideshowOpt(QStringList() << "slideshow", QObject::tr("Start slideshow playback"));
     parser.addOption(slideshowOpt);
 
+    QCommandLineOption fileListOpt(QStringList() << "file-list",
+                                   QObject::tr("Treat directories as file lists (expand to individual files)"));
+    parser.addOption(fileListOpt);
```
leejuyuu, line 124: I think we should spend some more time figuring out the flag name, as it would become something public, so we cannot easily change it. At first glance, `--file-list` seems vague to me. Does this only expand directories? If so, would `--expand-dirs` be more direct? Or does it also do other things?

ImageLounge/src/main.cpp
```diff
@@ -329,30 +338,67 @@ int main(int argc, char *argv[])
 
     nmc::DkCentralWidget *cw = w->getTabWidget();
 
-    bool loading = false;
+    QStringList filePaths = parser.positionalArguments();
+    if (parser.isSet("file-list")) {
```
leejuyuu, line 342: It would seem we can place `QCommandLineOption` in `isSet()`? If that is the case, the string key should be replaced by `fileListOpt`.

ImageLounge/src/main.cpp
```diff
@@ -329,30 +338,67 @@ int main(int argc, char *argv[])
...
+            }
+        }
+        filePaths = allImageFiles;
+    }
 
-    for (auto &filePath : parser.positionalArguments()) {
-        if (filePath.isEmpty())
-            continue;
+    // Handle slideshow with file list support
+    if (parser.isSet("file-list") && parser.isSet(slideshowOpt)) {
```
leejuyuu, line 362: It would seem we can place `QCommandLineOption` in `isSet()`? If that is the case, the string key should be replaced by `fileListOpt`.

ImageLounge/src/main.cpp
leejuyuu: Also, the expansion is not recursive? This probably should also be mentioned in the description.

maurolacy: We can add the option of recurse dirs in a follow-up. Will improve the description.

I like `--file-list` more, but am OK with both names. @scrubbbbs what do you think?

maurolacy:
> One thing feels very strange to me is that `DkPlayer` is a UI element. It should not be aware of what it is playing. It only need to send the next or previous signals. The file list and its logic should probably live in `DkViewPort` instead.

Hmm, I think you're right. Will refactor.

ImageLounge/src/main.cpp
scrubbbbs: `--file-list` is more forward-looking as the intent seems to be a mode change on the viewer

Directory expansion feels like a separate concept that doesn't need to be included in nomacs. For example, it is possible to pass "*" on the command line to get the same effect, and it does not limit to more advanced use cases, e.g. 

```sh
nomacs --file-list --slideshow ~/Pictures/*.jpg
find ~/Pictures -name '*.CRW' -print0 | xargs -0 nomacs --file-list --slideshow
```

This feels like a better solution; use the shell expansion for the file list and filter the list when `--file-list` is in effect.

maurolacy: Though the difference is only seen when passing files from the command line, and in slideshow mode (i.e. it's in the end a slight detail), I think the version with `--file-list` is more correct, and should eventually be the default.

scrubbbbs: I would like to keep `--file-list` as well but drop the expansion feature and change the description to what it currently does. Something like "Use file list passed on command-line instead of current directory contents."

maurolacy: I would have switched to file list based behaviour entirely, and without a flag. That would also have made the changes less complex, by example when integrating with Open Dir and Open File.

If I have a number of tabs opened, and then press space in one of them, to enter slideshow mode, the logical thing to do is, to slideshow over the list of currently opened images. Not over the files in the directory in which the current image happens to be.

If I use Open File a number of times, to open files in tabs, and then go to slideshow, again the logical thing is to slideshow over those files. And so on and so forth (i.e. in any selection of images and dirs).

The current behaviour is wrong; or at least, counterintuitive. That is what my PR attempted to change.

scrubbbbs: > The current behaviour is wrong; or at least, counterintuitive.

Most users will have one file open at a time, the natural thing to do is show the contents of that directory in the slideshow. The filters and sorting are there to help get the images one wants to actually see in the order they want to see them.

With tabs, users expect each tab to be in its own sandbox, e.g. they do not expect to rotate the current image and have all other open images also rotate. Likewise they do not expect the slideshow to pull files from all open tabs as the default behavior.

I think in general, any action that applies to all tabs should not be the default behavior, e.g. "CTRL+W" to close the current tab and "CTRL+SHIFT+W" to close all tabs. 

I think an action like "Slideshow All Tabs"  would be the best place to put this feature into the UI side, maybe in the Tabs menu. There should be no need to change OpenDir/OpenFile then, it is just a mode change on the slideshow components. For those who want this to be the default they can rebind the shortcut key for play/pause to this action.

maurolacy: > Most users will have one file open at a time, the natural thing to do is show the contents of that directory in the slideshow.

Fair enough. This is likely from where this behaviour comes from.

> With tabs, users expect each tab to be in its own sandbox, e.g. they do not expect to rotate the current image and have all other open images also rotate.

That is not valid reasoning, as slideshow's domain is different than rotate's domain. Slideshow is by definition a group operation. What we're discussing here is, precisely, what the contents of that group should be.

> Likewise they do not expect the slideshow to pull files from all open tabs as the default behavior.

Likewise, I (a user) do not expect the slideshow to pull files from the current's file directory as the default. Particularly, when having multiple tabs opened.

> I think an action like "Slideshow All Tabs" would be the best place to put this feature into the UI side, maybe in the Tabs menu. There should be no need to change OpenDir/OpenFile then, it is just a mode change on the slideshow components. For those who want this to be the default they can rebind the shortcut key for play/pause to this action.

I don't know. In any case, the `--file-list` flag could serve to switch behaviours. But, it should apply consistently for all operations. Not only when slideshowing over arguments from the command line. That would only introduce more inconsistency and confusion.

maurolacy:
Closing as we don't seem to agree on the expected behaviour when multiple tabs are opened.

Will perhaps revisit this in the future, and propose a new variant based on your comments. I'm pretty happy with my local version of nomacs, btw.

PR #1385, opened 2025-08-22
F/quiet flag

maurolacy:
Add a `--quiet|-q` flag to suppress informational messages.


leejuyuu:
This should be done by setting the log level to warning, instead of adding conditionals to all qInfo calls.

leejuyuu:
For example, 

```bash
QT_LOGGING_RULES="*.info=false" nomacs
```

maurolacy:
All right, will revert the changes. You can still merge this for the couple of (minor) fixes related to the Mac OSX installer, and spurious warning if you want.

ImageLounge/macosx/make-portable.sh
scrubbbbs, line 1: Curious as to how two versions of "cut" are on the system, is it commonplace?

ImageLounge/src/DkCore/DkImageContainer.cpp
scrubbbbs, line 1: I'd like to understand the source of the null images. Maybe we can prevent them in a more global manner.

ImageLounge/src/DkGui/DkWidgets.cpp
scrubbbbs, line 1: Dkoverview is rewritten in #1382 which would cancel out these checks.

ImageLounge/macosx/make-portable.sh
maurolacy: I would say so. GNU coreutils `cut` doesn't support the `-w` flag.

ImageLounge/src/DkGui/DkWidgets.cpp
maurolacy: Can revert those if you want; and leave the others.

leejuyuu:
I suggest doing a `git rebase -i` to drop the reverted commits. This helps maintain a clean history.

ImageLounge/src/DkCore/DkImageContainer.cpp
scrubbbbs: It occurred to me the null images might be from the slideshow PR work? If so you can filter the inputs with `DkUtils::isLoadable()` to do a fast check on known file extensions.

maurolacy: The `QImage::scaleHeight: Image is a null image` warning appears in the current master as well.

PR #1386, opened 2025-08-23
Fix single file directory issue

leejuyuu:


ImageLounge/src/DkGui/DkViewPort.cpp
```diff
@@ -259,6 +259,7 @@ void DkViewPort::onImageLoaded(QSharedPointer<DkImageContainerT> image, bool loa
     // things todo if a file was not loaded...
     if (!loaded) {
         mController->getPlayer()->startTimer();
+        mController->updateImage(nullptr);
         return;
```
scrubbbbs, line 263: Should mController should have a `loadingFailed() or removeImage()` method instead of specific flag for metadata updating? I would think other components would need a similar logic besides metadata.

scrubbbbs:
As far as loading/saving it seems to work a lot better. I found a few things:

With modified metadata, I don't get the save prompt when opening another file/folder in explorer or `File/Open`

When clicking off all stars I don't get the clear rating history entry and saving does not remove the rating.

When image has no rating and exif but no xmp, and I set the rating, xmp entries are added. When I then revert to original in history, the xmp entries are deleted in metadata panel, and also from the image when saved.

There is a desync with the metadata panel display however, only for the xmp fields:
image

ImageLounge/src/DkGui/DkViewPort.cpp
leejuyuu: Probably. There are multiple code paths for loading a new file, reloading the same file, not actually reloading but updating everything, etc. This part is already messy, so I feel it's better not to add more functions until we make the data flow clear.

leejuyuu:
@scrubbbbs Thanks for testing! I probably won't have enough time to work on this anytime soon. As these issues are not directly due to the PR, maybe we can mark them in a separate issue and merge this branch first?

scrubbbbs:
> @scrubbbbs Thanks for testing! I probably won't have enough time to work on this anytime soon. As these issues are not directly due to the PR, maybe we can mark them in a separate issue and merge this branch first?

That's fine. I think the problem with not being able to remove the rating should be top priority however. I'll break these out into separate issues.

PR #1404, opened 2025-09-18
Misc Easy Fixes

scrubbbbs:
Various trivial fixes I could find in issue tracker.

### DIFF
--- a/ImageLounge/src/DkGui/DkBatch.cpp
+++ b/ImageLounge/src/DkGui/DkBatch.cpp
@@ -465,8 +465,12 @@ void DkBatchInput::setDir(const QString &dirPath)
     mExplorer->setCurrentPath(dirPath);
 
     mCDirPath = dirPath;
-    qDebug() << "setting directory to:" << dirPath;
+
+    // block signals to prevent recursion through updateInputDir->setInputDir->setDir
+    mDirectoryEdit->blockSignals(true);
     mDirectoryEdit->setText(mCDirPath);
+    mDirectoryEdit->blockSignals(false);
+
     emit newHeaderText(mCDirPath);
     emit updateInputDir(mCDirPath);
     mLoader->loadDir(mCDirPath, false);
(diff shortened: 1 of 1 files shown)


ImageLounge/src/DkGui/DkToolbars.cpp
```diff
@@ -1158,7 +1158,7 @@ void DkToolBarManager::createDefaultToolBar()
     auto nomacs = dynamic_cast<QMainWindow *>(DkUtils::getMainWindow());
     Q_ASSERT(nomacs);
 
-    mToolBar = new DkMainToolBar(QObject::tr("Edit ToolBar"), nomacs);
+    mToolBar = new DkMainToolBar(QObject::tr("Toolbar"), nomacs);
     mToolBar->setObjectName("EditToolBar");
```
leejuyuu, line 1170: Should the following line match the new name?

ImageLounge/src/DkCore/DkActionManager.cpp
```diff
@@ -1456,7 +1456,7 @@ void DkActionManager::createActions(QWidget *parent)
     mToolsActions[menu_tools_batch]->setStatusTip(QObject::tr("Apply actions to multiple images"));
     mToolsActions[menu_tools_batch]->setShortcut(QKeySequence(shortcut_batch_processing));
 
-    mToolsActions[menu_tools_wallpaper] = new QAction(QObject::tr("&Wallpaper"), parent);
+    mToolsActions[menu_tools_wallpaper] = new QAction(QObject::tr("&Set Desktop Wallpaper"), parent);
```
leejuyuu: The '&' seems to be the underline in the menu. Since "Settings" already uses the letter "S", this "&" should probably be removed.

ImageLounge/src/DkGui/DkToolbars.cpp
scrubbbbs: No, the object name is more like a unique identifier for a widget or family of widgets. Primarily, nomacs uses the object name for settings groups and stylesheet selectors, but there are also some odd corner cases. So it could be renamed but that means also changing all the references to it and potentially breaking things.

ImageLounge/src/DkCore/DkActionManager.cpp
scrubbbbs: In this instance, the key combination "ALT+T, S" works correctly.

"ALT+" accelerators are scoped to the active/focused menu so this won't be a problem unless there is another "S" accel in the same menu -- and even then it is a minor problem as the "S" can be repeated to jump to the desired menu (though it would be annoying).

Nevertheless, I changed the accelerator back to "W", I think it's a better choice.

PR #1405, opened 2025-09-18
Fix Warnings

scrubbbbs:
Fix remaning deprecation warnings and other warnings

### DIFF
--- a/ImageLounge/src/DkCore/DkImageStorage.cpp
+++ b/ImageLounge/src/DkCore/DkImageStorage.cpp
@@ -414,8 +414,19 @@ QImage rotateImageFast(const QImage &img, double angle)
             return rotateImageCVMat(img.convertToFormat(QImage::Format_ARGB32), rot, CV_32F);
         }
 #else
+
+#if QT_VERSION >= QT_VERSION_CHECK(6, 9, 0)
+        constexpr auto flipImage = [](QImage image, Qt::Orientations flags) {
+            return image.flipped(flags);
+        };
+#else
+        constexpr auto flipImage = [](QImage image, Qt::Orientations flags) {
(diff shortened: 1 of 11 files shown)


ImageLounge/src/DkCore/DkImageStorage.cpp
```diff
@@ -414,51 +414,72 @@ QImage rotateImageFast(const QImage &img, double angle)
...
+
+#if QT_VERSION >= QT_VERSION_CHECK(6, 9, 0)
+        constexpr auto flipImage = [](QImage image, Qt::Orientations flags) {
+            return image.flipped(flags);
+        };
+#else
+        constexpr auto flipImage = [](QImage image, Qt::Orientations flags) {
+            return image.mirrored(flags & Qt::Horizontal, flags & Qt::Vertical);
+        };
+#endif
```
leejuyuu: This should probably be a function instead of lambda. The first param should be `const QImage& image`.

ImageLounge/src/DkGui/DkDialog.cpp
```diff
@@ -1862,7 +1862,7 @@ void DkShortcutsDialog::createLayout()
 
     QItemEditorCreatorBase *shortcutListCreator = new QStandardItemEditorCreator<QKeySequenceEdit>();
 
-    factory->registerEditor(QVariant::KeySequence, shortcutListCreator);
+    factory->registerEditor(QMetaType::QKeySequence, shortcutListCreator);
```
leejuyuu, line 1865: This doesn't need version guard?

scrubbbbs: Not needed, since added in Qt 6.0

PR #1409, opened 2025-09-30
OpenCV 4.12.0 in Windows build

novomesk:


scrubbbbs:
In appveyor it seems the cache is not working, opencv seems to be recompiled on every run.

novomesk:
The cache was not created because of the previous timeouts. Now it is OK.

scrubbbbs:
Did you have to restart the workflow run from appveyor or did it resolve itself?

novomesk:
I restarted it manually but it would auto-fix itself after the first future successful pass in master.

PR #1412, opened 2025-10-05
Theme Fixes

scrubbbbs:
Print dialog before/after
print-dialog-before
After:
print-dialog-after

Progress bar before/after

https://github.com/user-attachments/assets/afc9b3ea-1fa3-4dfe-9e3d-b515a1ac90d5

https://github.com/user-attachments/assets/9f4e874a-ab90-4461-9807-cc079be38b2e

Directory edit text before/after
dir-edit-before
dir-edit-after

Messagebox before/after
msgbox-before
msgbox-after


scrubbbbs:
@novomesk @leejuyuu I think this is good. The main issue you may have is that we need to create temporary files for the colorized svgs. The current code will clean up this svgs on exit of nomacs. But if nomacs crashes it will leave temporary files around until something else cleans them up (in my case, a reboot will clean them)

PR #1413, opened 2025-10-07
High Quality Thumbnails

scrubbbbs:
Updating main comment as a lot has changed.

This PR adds a user option for "high quality" thumbnails. By "high quality" what I am going for is:
- No "fuzzy" thumbnails, at least for big images
- If the EXIF thumbnail is too small, use the full image (or RAW preview)
- Use anti-aliasing (CV_AREA filter) and correct sizing for HiDPI screens.

Because we will be loading the full-size image when we didn't before (e.g. we only loaded EXIF usually) I have added some features to hide or speed that up:
- Thumbnail preloading; when scrolling get the the next visible page of thumbs queued up
- XDG-compatible thumbnail cache -  share thumbs with Linux file managers
- non-XDG mode for Windows/macOS, much smaller/faster cache by using JPEG.
- More settings to tweak behavior

There is also some notable new behavior:
- With hq thumbs mode a lot less memory may be used. We previously held the 400px image in the memory cache, now the "right-sized" image is held which is usually smaller.
- The memory cache isn't as crucial when disk cache is enabled, as cached items are fast to retrieve from SSD (on the order of 5-20ms).

Proposed settings with the default values.

image

Effect of the anti-aliasing. It is subtle but I think worth it.

backyard-normal

backyard-hq

### DIFF
--- a/ImageLounge/src/DkCore/DkBasicWidgets.cpp
+++ b/ImageLounge/src/DkCore/DkBasicWidgets.cpp
@@ -130,6 +130,12 @@ void DkSlider::setMaximum(int maxValue)
     maxValLabel->setText(QString::number(maxValue));
 }
 
+void DkSlider::setRange(int min, int max)
+{
+    setMinimum(min);
+    setMaximum(max);
+}
+
 void DkSlider::setTickInterval(int ticValue)
 {
     slider->setTickInterval(ticValue);
@@ -138,6 +144,11 @@ void DkSlider::setTickInterval(int ticValue)
 int DkSlider::value() const
 {
     return slider->value();
+}
+
+void DkSlider::setValueSuffix(const QString &format)
+{
+    sliderBox->setSuffix(format);
 }
 
 void DkSlider::setFocus(Qt::FocusReason reason)
--- a/ImageLounge/src/DkCore/DkBasicWidgets.h
+++ b/ImageLounge/src/DkCore/DkBasicWidgets.h
@@ -56,10 +56,11 @@ public:
     QSlider *getSlider() const;
     void setMinimum(int minValue);
     void setMaximum(int maxValue);
+    void setRange(int min, int max);
     void setTickInterval(int ticValue);
     int value() const;
+    void setValueSuffix(const QString &format);
     void setFocus(Qt::FocusReason reason);
-
 public slots:
     void setValue(int value);
 
--- a/ImageLounge/src/DkCore/DkCachedThumb.cpp
+++ b/ImageLounge/src/DkCore/DkCachedThumb.cpp
@@ -4,6 +4,7 @@
 #include "DkTimer.h"
 #include "DkUtils.h"
 
+#include <QBuffer>
 #include <QCryptographicHash>
 #include <QDir>
 #include <QSaveFile>
@@ -357,26 +358,29 @@ void DkCachedThumb::save(const QImage &img, int loadedBinSize)
     Q_ASSERT(!thumb.colorSpace().isValid() || thumb.colorSpace() == QColorSpace{QColorSpace::SRgb});
     thumb.setColorSpace({}); // prevent libpng "warning: iCCP: known incorrect sRGB profile"
 
-    // Write to a temporary and atomic swap to prevent thumb corruption (other processes or cleanup function)
-    QSaveFile f(cacheFilePath);
-    if (f.open(QFile::WriteOnly | QFile::Truncate)) {
-        const char *format = "png";
-        int quality = 50;
-        if (!sXdgCompliant && !DkImage::alphaChannelUsed(thumb)) {
-            // On Windows/macOS we don't need XDG compliance and can use JPEG thumbs
-            // Linux users also have this option if they don't care about sharing thumbs
-            format = "jpg";
-            quality = 90;
-        }
-        if (thumb.save(&f, format, quality)) {
-            if (f.commit()) {
+    const char *format = "png";
+    int quality = 50;
+    if (!sXdgCompliant && !DkImage::alphaChannelUsed(thumb)) {
+        // On Windows/macOS we don't need XDG compliance and can use JPEG thumbs
+        // Linux users also have this option if they don't care about sharing thumbs
+        format = "jpg";
+        quality = 90;
+    }
+
+    // We have many threads writing; compress to buffer for less disk fragmentation/syscalls
+    QByteArray data;
+    QBuffer buffer(&data);
+    if (thumb.save(&buffer, format, quality)) {
+        // Write to a temporary and atomic swap to prevent thumb corruption (other processes or cleanup function)
+        QSaveFile f(cacheFilePath);
+        if (f.open(QFile::WriteOnly | QFile::Truncate)) {
+            if (f.write(data) == data.length() && f.commit()) {
                 // qInfo() << "[CachedThumb] SAVED" << mFileInfo.fileName() << img.size() << "to" << thumb.size() << dt;
             }
         }
-    }
-
-    if (f.error() != QFile::NoError) {
-        qWarning() << "[CacheThumb] write failed:" << f.error() << f.errorString() << cacheFilePath;
+        if (f.error() != QFile::NoError) {
+            qWarning() << "[CacheThumb] write failed:" << f.error() << f.errorString() << cacheFilePath;
+        }
     }
 }
 
--- a/ImageLounge/src/DkCore/DkCachedThumb.cpp
+++ b/ImageLounge/src/DkCore/DkCachedThumb.cpp
@@ -5,6 +5,7 @@
 #include "DkUtils.h"
 
 #include <QBuffer>
+#include <QByteArrayView>
 #include <QCryptographicHash>
 #include <QDir>
 #include <QSaveFile>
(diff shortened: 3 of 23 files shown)


scrubbbbs:
The difference between default mode and high quality mode:
- default mode only uses exif thumb, otherwise scales full image to 400px
- hq mode uses exif thumb if bigger than requested size, scales full image or thumb to the requested size (depends on widget)

Note that in hq mode, the thumb cache will usually miss between grid/ribbon view because the thumb sizes are different.

scrubbbbs:
Rebased and added a XDG-compatible thumbnail cache. Now depends on #1521, though not much.

scrubbbbs:
@leejuyuu  this is pretty much complete if you want to take a first look, I'd be interested on any feedback on the decisions I've made here as you did the previous refactoring/feature work on thumbnails.

I know you are concerned about resource usage and I think we are still doing pretty good in that area and somewhat improved given that there are more tunable settings for those who want minimal resource usage.

I don't think there is much left to do besides a few minor details and I plan to finish it once #1521 is merged.

leejuyuu:
Hi, do you plan to rebase this branch? Github not supporting fast-forward seems to mess up the diff.

Haven't looked into this too much, but here's some questions:

- The XDG thumbnail cache seems to be in some standard directory, which means other applications could (and might) read and write from that directory. Is there a locking mechanism to avoid race conditions between nomacs and others?
- If another application already populates the cache directory, what will happen? Will they get overwritten by nomacs?
- I feel that the disk cache should be an opt-in feature instead of opt-out. I remember seeing some issue reporting "nomacs changes my file manager thumbnails." It might be better for users to be aware that this might cause changes outside of nomacs and apply explicitly.

scrubbbbs:
> Is there a locking mechanism to avoid race conditions between nomacs and others?

Yes and no. The potential race happens in `QSaveFile::commit()` which swaps a temporary with the target file, which relies on the rename system call being atomic. So we rely on the kernel, file system and mount options to support this. It should be pretty rare for this to fail, as atomic rename is crucial to prevent file corruption in modern OS.

>     * If another application already populates the cache directory, what will happen? Will they get overwritten by nomacs?

If the size we want is there and not stale, we use it. Otherwise we write a new one, overwriting what was there before.
 
>     * I feel that the disk cache should be an opt-in feature instead of opt-out. I remember seeing some issue reporting "nomacs changes my file manager thumbnails." It might be better for users to be aware that this might cause changes outside of nomacs and apply explicitly.

I was also thinking it should be opt-in but for different reasons - since it is a new feature and consumes extra resources. At the moment, it is disabled by setting storage to 0MB. Memory cache can be disabled similarly.

I believe the issue you are referring to is with file type registration affecting icons on Windows Explorer, I'm not sure where we stand on that at the moment but it is unrelated.

Another "changes outside of nomacs" issue is that we trim the thumbnail cache when starting nomacs, which will drop thumbnails created by other applications as well. This probably needs to be optional as we could be fighting with the desktop environment's cleanup method/settings.

leejuyuu:
>  At the moment, it is disabled by setting storage to 0MB. Memory cache can be disabled similarly.

While setting the value to 0 has the same effect, it is usually clearer to have a dedicated boolean enable/disable toggle. I prefer setting it to disabled by default. 

Looking at the settings, some suggestions:

- "Shared thumbnail cache" and "Thumbnail disk cache" should use the same terminology and ideally be grouped together.
- If cache is disabled, disabling (graying out) the size control should make it clearer that they are in the same group.
- The "disk cache" would be more accurate as "disk cache limit" or "max disk cache size".
- What is the unit of "Maximum thumbnail size" (pixel, count)?
- The positions of some of the tooltips look slightly low on my end.


> we trim the thumbnail cache when starting nomacs, which will drop thumbnails created by other applications as well. 

This seems different from "If the size we want is there and not stale, we use it. Otherwise we write a new one, overwriting what was there before.", so it probably needs more explanation.

What is the reason to drop thumbnails? This must be disabled by default. It is risky to silently delete files created by other applications.

This is overall a great feature. As nomacs now interacts with other applications through this shared directory, I think it is important to ensure that whichever app or user who already uses it does not get disrupted.

leejuyuu:
Also, does the high-quality thumbnail require OpenCV? Can we disable the setting for non-OpenCV build if that is the case?

scrubbbbs:
> > we trim the thumbnail cache when starting nomacs, which will drop thumbnails created by other applications as well.
> 
> This seems different from "If the size we want is there and not stale, we use it. Otherwise we write a new one, overwriting what was there before.", so it probably needs more explanation.
> 
> What is the reason to drop thumbnails? This must be disabled by default. It is risky to silently delete files created by other applications.

It is to prevent using too much storage space for thumbnails. With PNG thumbnails at 1024px this can easily grow to 10s of GB (one thumbnail is about 1MB). The risk is minimal since this is a shared cache; XDG spec doesn't say anything about how thumbnail cache should be cleaned up so the expectation is that cache entries can disappear at any time while the application is running.

And so applications using the cache cannot make any assumptions about cached thumb lifetime, they shouldn't have any problems with what we are doing.

The issue to consider is if there is any other thumbnail cleanup routine on the system that has different retention policy than nomacs. For example, it could have a smaller or larger maximum cache size setting or it might delete thumbs based on age or if the original no longer exists. I haven't studied these on the major desktop environments yet, but since this is an unknown variable we have to have an option to disable the cleanup when the cache is shared.

scrubbbbs:
> Also, does the high-quality thumbnail require OpenCV? Can we disable the setting for non-OpenCV build if that is the case?

I discovered recently that `Qt::TransformationMode::SmoothTransformation` is effectively the same as `CV_AREA` for downsampling so we don't have to disable it for non-CV builds. But I will need to check if this works.

On the topic of the non-OpenCV build maybe that should be deprecated. Is anyone using this that we know of? To mitigate the large dependency we can have an option to build & link a minimal OpenCV for nomacs, e.g. `USE_SYSTEM_OPENCV=OFF`

leejuyuu:
> XDG spec doesn't say anything about how thumbnail cache should be cleaned up so the expectation is that cache entries can disappear at any time while the application is running.
> And so applications using the cache cannot make any assumptions about cached thumb lifetime, they shouldn't have any problems with what we are doing.

The problem is that we do not own the files. Other applications may or may not respect the spec. When they do not, generally, the user won't blame the spec-breaker—they'll blame nomacs for "breaking" an app that previously worked. I would like the implementation to be more defensive so that we do not end up in that situation.


> The issue to consider is if there is any other thumbnail cleanup routine on the system that has different retention policy than nomacs. For example, it could have a smaller or larger maximum cache size setting or it might delete thumbs based on age or if the original no longer exists. I haven't studied these on the major desktop environments yet, but since this is an unknown variable we have to have an option to disable the cleanup when the cache is shared.

If such a tool exists, maybe we can omit the housekeeping and point the users to these tools. This lowers the responsibility of nomacs, and probably adheres closer to Unix philosophy.

> On the topic of the non-OpenCV build maybe that should be deprecated. 

Might be a good idea. The Ubuntu and AUR packages all require OpenCV. A minimum version of OpenCV is not necessary in my opinion, as the core and improc modules are not that large.

scrubbbbs:
> The problem is that we do not own the files. Other applications may or may not respect the spec. When they do not, generally, the user won't blame the spec-breaker—they'll blame nomacs for "breaking" an app that previously worked. I would like the implementation to be more defensive so that we do not end up in that situation.

I don't think we will be breaking any apps, it has been standard practice to `rm -rf ~/.cache/thumbnails` for a long time.

In the worst case, thumbnails would regenerated unexpectedly, which isn't that big of a problem:
- It is unlikely to show up at all due to LRU deletion.
- There is low expectation that files accessed a long time ago would have thumbs cached
- If it is an annoyance to users, they have the option to disable cleanup (which was disabled by default) or increase the maximum cache size. Perhaps we can show the current cache usage in the settings so it is more clear if the size setting will cause a trim to occur.

> > The issue to consider is if there is any other thumbnail cleanup routine on the system that has different retention policy than nomacs [...]
> 
> If such a tool exists, maybe we can omit the housekeeping and point the users to these tools. This lowers the responsibility of nomacs, and probably adheres closer to Unix philosophy.

It seems there is no DE that has this feature and they are happy to let cache grow indefinitely by default rather than the off-chance a thumbnail has to regenerate. There also seems to be shortage of reliable programs that can do it, for example even today KDE "Sweeper" program has [major flaws in this regard](https://bugs.kde.org/show_bug.cgi?id=487878).

I think it makes sense that a program that generates files can clean up after itself, when it cannot that is a major annoyance. Also users won't have to search for and install another utility which may or may not work properly, or hacking something themselves, e.g. `rm -rf ~/.cache/thumbnails`

Additionally, it makes sense to add a feature to nomacs to delete all history which would include thumbnails as well as recent files.

leejuyuu:
> it has been standard practice to rm -rf ~/.cache/thumbnails for a long time.

This is exactly my point. "The user" executes `rm`, therefore the result is intended. In contrast, opening a directory in nomacs resulting in the deletion of files is unexpected.

> I think it makes sense that a program that generates files can clean up after itself

I agree, given that the program owns the file, e.g., by storing it somewhere like `~/.cache/nomacs`.

scrubbbbs:
> > it has been standard practice to rm -rf ~/.cache/thumbnails for a long time.
> 
> This is exactly my point. "The user" executes `rm`, therefore the result is intended. In contrast, opening a directory in nomacs resulting in the deletion of files is unexpected.
> 

The default behavior will be not to do this, since that is what everyone else is doing. So it should be expected because the setting had to be enabled.

> > I think it makes sense that a program that generates files can clean up after itself
> I agree, given that the program owns the file, e.g., by storing it somewhere like `~/.cache/nomacs`.

There is no owner of files in a shared cache. So by that logic, they cannot be deleted by any program. Since users do not know such things exist that they need to take care of then who should be cleaning them up?

leejuyuu:
> There is no owner of files in a shared cache. So by that logic, they cannot be deleted by any program. Since users do not know such things exist that they need to take care of then who should be cleaning them up?

The user should know the cache exists (at some point) and make the decision to either `rm -r` or set up some periodic job for that. Given that most DE do not remove thumbnails by default, I'd guess this can be expected.

scrubbbbs:
> > There is no owner of files in a shared cache. So by that logic, they cannot be deleted by any program. Since users do not know such things exist that they need to take care of then who should be cleaning them up?
> 
> The user should know the cache exists (at some point) and make the decision to either `rm -r` or set up some periodic job for that. 

The problem is "should know", people are not aware that a program they are using is slowly consuming all of their free disk space until they run out. So what is worse, nomacs deleting temp files that are easily re-created, or getting low disk space warning in the middle of trying to get work done, having to find a tool to fix it, then finding out nomacs is holding 20GB of thumbnails from months ago. I would prefer to have trivial temp files deleted than be forced to fix the problem myself.

If thumbnail files took forever and a day to re-create then I can see your point, but usually they do not, and we have the checkbox to disable the feature.

> Given that most DE do not remove thumbnails by default, I'd guess this can be expected.

We have an opportunity to provide something valuable nobody else is bold enough to do, or perhaps too lazy. As a compromise, if shared thumbs are enabled, we could disable automatic cleanup and add a button to explicitly cleanup thumbnails with a a warning message e.g. "You are about to delete old thumbnails from ALL programs" etc.

leejuyuu:
> nomacs deleting temp files that are easily re-created

The problem is not the deletion. The problem is that the files are shared, and we cannot control the effects of deleting them unless we know who else might be using them.

> disable automatic cleanup and add a button to explicitly cleanup thumbnails with a a warning message e.g. "You are about to delete old thumbnails from ALL programs" etc.

This sounds reasonable to me. The deletion should be triggered explicitly by the user.

scrubbbbs:
How about no button but we add this dialog when enabling the feature?

image

leejuyuu:
The setting is not clear. Can the thumbnails be cached to disk and not shared? If the 'Delete old thumbnails' is an independent control, the statement 'When thumbnails are shared, nomacs will also delete thumbnails' is not accurate. It isn't clear when the deletion will happen.

scrubbbbs:
> The setting is not clear. Can the thumbnails be cached to disk and not shared?

Yes. This is the only mode available on Windows and macOS, until someone wants to hook into the system thumbnail APIs. The share checkbox will be missing or greyed out on Windows/macOS.

> If the 'Delete old thumbnails' is an independent control, the statement 'When thumbnails are shared, nomacs will also delete thumbnails' is not accurate. It isn't clear when the deletion will happen.

Yes, it is independent. Deletion is described in the tooltip of delete thumbnails checkbox, which is when nomacs starts.

leejuyuu:
> > The setting is not clear. Can the thumbnails be cached to disk and not shared?
> 
> Yes. This is the only mode available on Windows and macOS, until someone wants to hook into the system thumbnail APIs. The share checkbox will be missing or greyed out on Windows/macOS.

On Linux, is it possible to select whether thumbnails are shared? Or it is always shared?

scrubbbbs:
> > > The setting is not clear. Can the thumbnails be cached to disk and not shared?
>>
> > Yes. This is the only mode available on Windows and macOS, until someone wants to hook into the system thumbnail APIs. The share checkbox will be missing or greyed out on Windows/macOS.
> 
> On Linux, is it possible to select whether thumbnails are shared? Or it is always shared?

Sharing is optional on Linux. Sharing is slower and uses more space for each thumb as it must use PNG. Non-shared mode uses JPG.

leejuyuu:
> Sharing is optional on Linux. Sharing is slower and uses more space for each thumb as it must use PNG. Non-shared mode uses JPG.

I thought that they are always shared on Linux. This changes the conclusions, though. What if we don't support the shared thumbnails? Now that the thumbnails are under our control (e.g. in `~/.cache/nomacs/thumbs`), we no longer need to expose the housekeeping logic and timing. The user only needs to set the max size. This seems to simplify things a lot, and I do not think we have much to lose due to the mentioned performance. Also, Nomacs UI generally requests specific sizes of thumbnails; I'd guess reading from the shared directory will miss more often, and Nomacs is going to replace the files anyway.

scrubbbbs:
> Also, Nomacs UI generally requests specific sizes of thumbnails; I'd guess reading from the shared directory will miss more often, and Nomacs is going to replace the files anyway.

The requested size is rounded up to the nearest power of two (not exceeding max-thumb-size setting) to get the cached thumb. So an exact size match is not required, cache hits are more likely.

I have another idea, we can delete thumbs only if nomacs created them or overwrote them at some point. This is a bit slower to do the deletion but addresses the ownership issue.

scrubbbbs:
> I have another idea, we can delete thumbs only if nomacs created them or overwrote them at some point. This is a bit slower to do the deletion but addresses the ownership issue.

@leejuyuu  I implemented this in the current version. I think that resolves your primary concern and the PR seems ready to merge as-is - unless you want to do a more detailed review.

leejuyuu:
Some suggestions for the settings. I would like the meaning settings to be clear enough at a glance.

- Maximum thumbnail size has no unit. It isn't clear whether it is file size or image dimension.
- "Thumbnail memory cache" -> "Thumbnail memory cache limit" (or size)
- "Save thumbnails" -> "Cache thumbnails to disk". Using "save" might be confused with saving the thumbnail when saving an edited file.
- "Delete old thumbnails" -> "Delete old thumbnails on start" or "Clear old cache on start"
- "Maximum disk space" -> "Maximum disk space for cache" or "Maximum disk cache limit"

leejuyuu:
Otherwise, I think this is good to go. Thanks!

PR #1425, opened 2025-10-28
Support AppImage

scrubbbbs:
I think this will help get more people on board testing the Linux version, as many can't be bothered to compile it or they don't want to mess up their current install.

There isn't much to it, just `make appimage` after compiling. It seems to do a good job pulling any dependencies we need, and will include kimageformat-plugins when present.

For CI I'm building AppImages for gcc configurations only, I think that is enough. I have also added quazip+kimageformats images built against neon user repository for 22.04/jammy and 24.04/noble which are suitable for releases.

I've elected to use docker images in CI. These are all built within github workflows and derived from official images (`ubuntu:24.04` and `ubuntu:22.04`). So there is no part that ever touched my machine to make the docker images or AppImages (exception is the linuxdeploy binary I had to add to the repo).

The use of docker currently is not strictly required, I could have apt-installed everything into the `ubuntu-latest` runner to get just about the same result. However, I think it might be worthwhile for a few reasons:
- It does speed up the builds as the dependency fetch/install is no longer involved. Caveat is I cannot seem to run more than two jobs at a time - still unable to find an answer on this but it is some kind of resource limit.
- Opens the door to build against anything we can get into a docker image, latest versions of Qt libraries not in repos yet for example.
- The builds are reproducible if we get into a tricky issue, just clone the image and run in docker locally.

For maintenance we will periodically need to rebuild the docker images from the Actions control panel, I think it can be put on a schedule to automate it. If we forget, worst case is maybe builds take longer (for the moment) since apt packages are still checked in `linux.yml`.

Note that to update docker images (github calls them "Packages"), the recipes in `nomacs/scripts/docker` and workflow file `.github/workflows/docker.yml` must be pushed to master first - this is a limitation/security feature of packages. As a result I did have to push quite a few times to get things right despite pre-testing in another account.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,5 +1,6 @@
 name: Linux workflow
 on: 
+  workflow_dispatch:
   push:
     branches:
       - master


scrubbbbs:
I'm working on an issue with libheif in the neon/24.04 AppImage, please don't use these binaries yet.

ImageLounge/cmake/UnixBuildTarget.cmake
```diff
@@ -124,6 +124,66 @@ include(CPack)
...
+    DEPENDS ${LINUXDEPLOY_FILE} ${LINUXDEPLOY_QT_FILE}
+    COMMAND rm -rf squashfs-root linuxdeploy
+    COMMAND ${LINUXDEPLOY_QT_FILE} --appimage-extract
+    COMMAND mv -f squashfs-root linuxdeploy
+    COMMAND ${LINUXDEPLOY_FILE} --appimage-extract
+    COMMAND rsync -a squashfs-root/ linuxdeploy/
+    COMMAND tar -acvf ${LINUXDEPLOY_ARCHIVE} linuxdeploy
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+    COMMENT "Updating ${LINUXDEPLOY_ARCHIVE}"
+)
```
leejuyuu: I think these should not be in the build configuration. The cmake things should only check that the required binary exists or is functional. How and where to install depends on the specific distro, and the user should be aware that they are downloading binary from somewhere and will execute them later.

As these seems to be used in the Dockerfile, it can be extracted into a separate script, or written in the Dockerfile directly. It might be beneficial to leverage the caching of docker build.

Also, the URL seems to point to the latest release. Depend on how stable it is, you might want to pin to a specific version.

ImageLounge/3rdparty/tools/linuxdeploy.tgz
leejuyuu, line 1: What does this contain? It's weird to see tgz in the source tree.

ImageLounge/cmake/UnixBuildTarget.cmake
scrubbbbs: Yes this is not great solution. It also has the problem that the file in the repo is not architecture-independent and I want to be able to offer builds for the arm64 etc as well.

I do like having utility targets like this for maintenance purposes. But I suppose they could also be stored in `./scripts/` instead, that could be cleaner. I will look into compiling linuxdeploy from source and including in the docker images.

Keep in mind linuxdeploy also downloads things on its own (at least the bootstrap binary blob part), so we can't completely avoid downloading binaries.

ImageLounge/3rdparty/tools/linuxdeploy.tgz
scrubbbbs: Inside is just AppDir structure, the result of unpacking the linuxdeploy AppImage

leejuyuu: It seems to be made by linuxdeploy? It's probably better to pull it when needed in the linuxdeploy install scripts.

scrubbbbs: It's the output file from `make update-linuxdeploy`. But yeah I am going to try another direction on this and drop it as well as the other stuff mentioned. I should be able to write build script for linuxdeploy in `./nomacs/scripts/` and reference that from the Dockerfiles.

scrubbbbs: I have figured out how to compile linuxdeploy, will be pushing those changes soon

PR #1448, opened 2025-11-08
implement keyboard scrolling (pgup, pgdn, etc) in thumb view

spaghetti22:
This PR implements PgUp, PgDn, Home, End, Space and Shift+Space keyboard scrolling in the thumbnail view.

### DIFF
--- a/ImageLounge/src/DkGui/DkThumbsWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkThumbsWidgets.cpp
@@ -1699,6 +1699,45 @@ DkThumbsView::DkThumbsView(DkThumbScene *scene, QWidget *parent /* = 0 */)
 
     connect(horizontalScrollBar(), &QScrollBar::valueChanged, this, &DkThumbsView::onScroll);
     connect(verticalScrollBar(), &QScrollBar::valueChanged, this, &DkThumbsView::onScroll);
+
+    auto *scrollOnePageUpAction = new QAction(this);
+    scrollOnePageUpAction->setShortcuts({QKeySequence::MoveToPreviousPage, QKeySequence(Qt::SHIFT | Qt::Key_Space)});
+    connect(scrollOnePageUpAction, &QAction::triggered, this, [this] {
+        if (verticalScrollBar()->isVisible()) {
+            verticalScrollBar()->setValue(verticalScrollBar()->value() - verticalScrollBar()->pageStep());
+        }
+    });
+
+    auto *scrollOnePageDownAction = new QAction(this);
+    scrollOnePageDownAction->setShortcuts({QKeySequence::MoveToNextPage, QKeySequence(Qt::Key_Space)});
+    connect(scrollOnePageDownAction, &QAction::triggered, this, [this] {
+        if (verticalScrollBar()->isVisible()) {
+            verticalScrollBar()->setValue(verticalScrollBar()->value() + verticalScrollBar()->pageStep());
+        }
+    });
+
+    auto *scrollToTopAction = new QAction(this);
+    scrollToTopAction->setShortcuts({Qt::Key_Home, QKeySequence::MoveToStartOfDocument});
+    connect(scrollToTopAction, &QAction::triggered, this, [this] {
+        if (verticalScrollBar()->isVisible()) {
+            verticalScrollBar()->setValue(verticalScrollBar()->minimum());
+        }
+    });
+
+    auto *scrollToEndAction = new QAction(this);
+    scrollToEndAction->setShortcuts({Qt::Key_End, QKeySequence::MoveToEndOfDocument});
+    connect(scrollToEndAction, &QAction::triggered, this, [this] {
+        if (verticalScrollBar()->isVisible()) {
+            verticalScrollBar()->setValue(verticalScrollBar()->maximum());
+        }
+    });
+
+    auto *actionFilter = new DkActionEventFilter(this);
+    actionFilter->addAction(scrollOnePageUpAction);
+    actionFilter->addAction(scrollOnePageDownAction);
+    actionFilter->addAction(scrollToTopAction);
+    actionFilter->addAction(scrollToEndAction);
+    this->installEventFilter(actionFilter);
 }
 
 void DkThumbsView::onScroll()


scrubbbbs:
Thanks for the PR! I am trying to get rid of keyPressEvent if possible so we can clean up key bindings in nomacs at some point. That means either:
- using a QAction tied to that widget
- using the `DkShortcutEventFilter::reserveKeys()` to allow certain standard keys to bypass shortcuts reliably.

If there is no way to use actions for this, then I would suggest finding out why the underlying widget is not receiving the PgUp/PgDown first. Make sure to do a debug build to get logging from the shortcut arbitration. For example, it appears PgUp/PgDown should be working via QAbstractScrollArea in this context, but is not for some reason.

`
[DEBUG] [shortcuts]  nmc::DkThumbsView "DkThumbsView" blocking QKeySequence("PgDown") reserved by QAbstractScrollArea ""
`

scrubbbbs:
I also noticed you are using the space bar, that could be problematic due to confusion with default play/pause key. It is nice however not to have to reach for pgup/pgdown. Maybe there is a better option?

Is space bar a standard scroll key besides web browsers?

spaghetti22:
Hi,

thanks for the feedback. I do not yet know enough about Qt or nomacs to immediately be able to reimplement it in any of the ways you mention. I may look into it some time in the next weeks.

As for your question about the space bar: you already mentioned web browsers, which is obviously where I got the idea from. But the space bar also scrolls in e.g. Thunderbird, Okular, the KDE help center, `less`, `man` (which may be using `less` in the background?) and probably other software too. Admittedly those are all things that display some kind of document, which isn't what is going on here, so it's indeed not quite the same thing; it's your decision, I can also remove that part if desired.

Does the space bar really do anything else already *in the thumbnail view*? If it does, that part should certainly be removed.

scrubbbbs:
> thanks for the feedback. I do not yet know enough about Qt or nomacs to immediately be able to reimplement it in any of the ways you mention. I may look into it some time in the next weeks.

That's no problem. We are feature freeze until 3.22 release, maybe for the next few weeks to see if any major bugs are discovered. When you resume work on this I would immediately look into why these keys do not make it to the scroll area first.

> Does the space bar really do anything else already _in the thumbnail view_? If it does, that part should certainly be removed.

I don't know but I would guess OK to use at the moment. My main thought is a consistency issue we might be creating, space bar would not work the same in different vertical scrolling situations or the main image view.

We also have a top feature request for vertical scrolling of the image itself. If one pair of keys could bind to this we could make it a configurable application-wide keybind.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -1735,6 +1739,32 @@ DkThumbsView::DkThumbsView(DkThumbScene *scene, QWidget *parent /* = 0 */)
 
     connect(horizontalScrollBar(), &QScrollBar::valueChanged, this, &DkThumbsView::onScroll);
     connect(verticalScrollBar(), &QScrollBar::valueChanged, this, &DkThumbsView::onScroll);
+
+    QAction *scrollOnePageUpAction = new QAction(this);
+    scrollOnePageUpAction->setShortcuts({
+        QKeySequence(Qt::Key_PageUp),
```
scrubbbbs: There are Qt::StandardKey for these shortcuts, I think they would be slightly better choice here (and for all others except space bar).

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -1735,6 +1739,32 @@ DkThumbsView::DkThumbsView(DkThumbScene *scene, QWidget *parent /* = 0 */)
 
     connect(horizontalScrollBar(), &QScrollBar::valueChanged, this, &DkThumbsView::onScroll);
     connect(verticalScrollBar(), &QScrollBar::valueChanged, this, &DkThumbsView::onScroll);
+
+    QAction *scrollOnePageUpAction = new QAction(this);
```
scrubbbbs: We prefer "auto * scrollOnePageUpAction=..." since type is on the same line. I recommend running clang-tidy on the files you changed for these types of hints.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -1735,6 +1739,32 @@ DkThumbsView::DkThumbsView(DkThumbScene *scene, QWidget *parent /* = 0 */)
 
     connect(horizontalScrollBar(), &QScrollBar::valueChanged, this, &DkThumbsView::onScroll);
     connect(verticalScrollBar(), &QScrollBar::valueChanged, this, &DkThumbsView::onScroll);
+
```
scrubbbbs, line 1702: If you are going to make the actions non-configurable by user, you must use DkActionEventFilter. Search the source for examples.

Your other choice is to define new DkActionManager::HiddenActions - then they will be configurable.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -1735,6 +1739,32 @@ DkThumbsView::DkThumbsView(DkThumbScene *scene, QWidget *parent /* = 0 */)
...
+    QAction *scrollOnePageUpAction = new QAction(this);
+    scrollOnePageUpAction->setShortcuts({
+        QKeySequence(Qt::Key_PageUp),
+        QKeySequence(Qt::SHIFT | Qt::Key_Space)
+    });
+    connect(scrollOnePageUpAction, &QAction::triggered, this, &DkThumbsView::scrollOnePageUp);
```
scrubbbbs: Since the signal handlers are only a few lines, it is cleaner to use lambda connection here

```
connect(scrollOnePageUpAction, &QAction::triggered, this, [this] {
   // ...
});

File: ImageLounge/src/DkGui/DkThumbsWidgets.cpp [hunk 5]
```diff
@@ -1910,6 +1940,38 @@ void DkThumbsView::dropEvent(QDropEvent *event)
...
+}
+
+void DkThumbsView::scrollToTop()
+{
+    if (verticalScrollBar()->isVisible()) {
+        verticalScrollBar()->setValue(verticalScrollBar()->minimum());
+    }
+    mThumbScene->selectAllThumbs(false);
+    mThumbScene->selectThumb(0, true);
+}
```
scrubbbbs: There is an inconsistency wrt selections. Should PageUp/PageDown do a similar thing, e.g. move the selected image to the next/previous page?

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
spaghetti22: done, but it seems from the documentation that in order to make it work on macOS too, I need to add more than one of them (I have only tested my changes on KDE Plasma)

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
spaghetti22: DkActionEventFilter done, not many examples in the existing codebase, but I hope I understood this correctly.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
spaghetti22: It's true that this is inconsistent, and I obviously thought about this.

My thought was that someone might be using PgUp/PgDown to scroll around a huge directory while already having an existing selection, and not want that discarded. (It would also be non-obvious which of the several images should be selected by pressing those keys: the one in the top left of the page, middle of the page, bottom right of the page?) Meanwhile having Home and End select the first/last image is useful in order to make it possible for people to start a selection there using only the keyboard.

If we want to resolve the inconsistency, I think it's better to make Home and End no longer change the selection either, not add such functionality to PgUp/PgDown. This isn't very important to me. I can remove it if desired, but thought it would be a useful feature.

scrubbbbs: I feel it should be removed, and later we can add Shift+(shortcut) to be the way to modify selections, similar to a text document.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -1735,6 +1735,45 @@ DkThumbsView::DkThumbsView(DkThumbScene *scene, QWidget *parent /* = 0 */)
...
+
+    auto *scrollOnePageUpAction = new QAction(this);
+    scrollOnePageUpAction->setShortcuts({QKeySequence::MoveToPreviousPage, QKeySequence(Qt::SHIFT | Qt::Key_Space)});
+    connect(scrollOnePageUpAction, &QAction::triggered, this, [this] {
+        if (verticalScrollBar()->isVisible()) {
+            verticalScrollBar()->setValue(verticalScrollBar()->value() - verticalScrollBar()->pageStep());
+        }
+    });
+
+    QAction *scrollOnePageDownAction = new QAction(this);
```
scrubbbbs: Please also use `auto` here as above, and also below.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -1735,6 +1735,45 @@ DkThumbsView::DkThumbsView(DkThumbScene *scene, QWidget *parent /* = 0 */)
...
+    scrollOnePageDownAction->setShortcuts({QKeySequence::MoveToNextPage, QKeySequence(Qt::Key_Space)});
+    connect(scrollOnePageDownAction, &QAction::triggered, this, [this] {
+        if (verticalScrollBar()->isVisible()) {
+            verticalScrollBar()->setValue(verticalScrollBar()->value() + verticalScrollBar()->pageStep());
+        }
+    });
+
+    QAction *scrollToTopAction = new QAction(this);
+    scrollToTopAction->setShortcuts({QKeySequence::MoveToStartOfLine, QKeySequence::MoveToStartOfDocument});
```
scrubbbbs: We only need start/end of document. The issue with macOS is that the shortcut is mapped to (`Command+Up/Command+Down` and DkThumScene is eating the arrow keys regardless of modifier keys pressed or not. We can leave fixing macOS for a later date.

spaghetti22: No, that doesn't work if we want home/end to work. "start/end of document" is ctrl+home and ctrl+end, "start/end of line" is home and end. Users expect home and end to scroll to the top/bottom even without ctrl pressed (like in a browser window), so the present solution seems to be correct.

scrubbbbs: OK, since we are not using Home/End in a standard way, then let's be explicit and don't use any standard key for that (as in the original commit). While also including the standard key for document begin/end which is different on macOS and Windows/Linux.

Also, you will need to rebase again to pass CI checks.

scrubbbbs:
Merging now, thanks!

Issue #1442, opened 2025-11-05
nomacs 3.22.0 unable to rotate


fcore117:
**Describe the bug**

**To Reproduce**
1. Open any image
2. Click on any rotate button or menu
3. Unable to rotate

**Expected behavior**
Should rotate image like intended

**Desktop (please complete the following information):**
nomacs 3.22.0, x86_64
becd1d87, 2025-10-30
Windows 11 Version 25H2
Qt 6.9.2, windows, scale=1.00
Exiv2 0.28.7
LibRAW 0.21.4-Release
OpenCV 4.12.0
TIFF 4.6.0
QuaZip v1

scrubbbbs:
Thanks for testing the new release! 

There are two ways to rotate files in nomacs. The 90-degree rotation buttons are actually only for metadata rotation - e.g. just setting the "Orientation" flag on the image file EXIF data. If the image has no EXIF metadata we don't create it automatically. I'm not sure if nomacs ever did this before, I have a feeling it probably did which is why you see this now.

Adjustments/Rotation changes the image pixels so it should always work. Are you seeing this on your end or is something else wrong here?

I would propose we either disable the buttons when EXIF is missing, create some default EXIF metadata if the file format supports it, or have different buttons for normal vs metadata rotation. I like the idea of creating default metadata but that still means we have to disable the buttons if the format doesn't support it.

fcore117:
I used NoMacs for years and years to simply edit, crop, rotate etc photos, any photo or even clipboard pasted image without issues. Disabling rotate would be disaster.

scrubbbbs:
I don't want to either, maybe a rotate-90 button with "multiple personalities" is what people want, as long as we could make it clear what it was doing. When working with camera photos you typically do not want the rotation to modify the image data/pixels, but then when pasting images it has to.

fcore117:
Reverting back to 3.19.1.205 until heavy bugs fixed.

scrubbbbs:
Thanks for reporting, this was a mistake I made in refactoring. Will be fixed in 3.22 final.

fcore117:
Thanks i hope it works again in next release even images without exif data destructively like it was previously or even with other viewers as example like irfan etc...

scrubbbbs:
It will try to do it non-destructive if there is EXIF data, and fallback to destructive. So a rotation will always happen one way or the other. I think this was how we had it in the past, though it has bugs with some image formats which I am trying to improve on now.

PR #1453, opened 2025-11-16
90-Degree Rotation Improvements

scrubbbbs:
Fix rotation regression #1442 and try to prevent unexpected behavior like wrong thumbnails or reloading the file changes the image orientation. See commit message for details.

There is still much to be desired, like lossless rotation for RAW and TIFF (exiv2 supports this, but we cannot due to how DkMetaData and DkEditImage work). Maybe that will come later, but not for 3.22.
 
The most visible change, when orientation is ignored either because of the image format or the checkbox in settings, *and* metadata rotation will be used: 

image

New dialog when orientation saving is disabled (default enabled).

image

### DIFF
--- a/ImageLounge/src/DkCore/DkBasicLoader.cpp
+++ b/ImageLounge/src/DkCore/DkBasicLoader.cpp
@@ -94,24 +94,18 @@ DkEditImage::DkEditImage()
     , mNewMetaData(false)
 {
 }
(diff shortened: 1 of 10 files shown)


novomesk:
If you are sure about it, let's merge. We can make RC2 and let people to re-test.

PR #1454, opened 2025-11-18
Misc Fixes for 3.22

scrubbbbs:
New dialog for permanently deleting files:

image

Basic color management added. Images are displayed in sRGB, while the source colorspace is preserved when editing and saving. CMYK is always converted to sRGB since we cannot display it correctly and adjustments/edits do not work with it. Here are some sample images (right side is new version):

Rec 2020:
img-2025-11-21 14 16 51

Display P3:
img-2025-11-21 14 21 45

CMYK:
img-2025-11-21 14 27 40

Display P3:
img-2025-11-21 15 03 02

### DIFF
--- a/ImageLounge/src/DkCore/DkBaseViewPort.cpp
+++ b/ImageLounge/src/DkCore/DkBaseViewPort.cpp
@@ -559,7 +559,10 @@ void DkBaseViewPort::draw(QPainter &frontPainter, double opacity)
 
     QPainter &painter = backPainter ? *(backPainter.get()) : frontPainter;
 
-    eraseBackground(painter);
+    // do not color-manage the viewport background; defer for better locality when swapping
+    if (!backPainter) {
+        eraseBackground(frontPainter);
+    }
 
     // opacity == 1.0f -> do not show pattern if we crossfade two images
     if (DkSettingsManager::param().display().tpPattern && img.hasAlphaChannel() && opacity == 1.0)
@@ -603,6 +606,7 @@ void DkBaseViewPort::draw(QPainter &frontPainter, double opacity)
     if (backPainter) {
         backPainter->end();
         mBackBuffer.convertToColorSpace(targetColorSpace);
+        eraseBackground(frontPainter);
         frontPainter.setWorldMatrixEnabled(false);
         frontPainter.setRenderHint(QPainter::SmoothPixmapTransform, false);
         frontPainter.drawImage(QPoint{0, 0}, mBackBuffer);


novomesk:
Color management is really very important feature.

I'd postpone this work after final 3.22 release.

As a side effect of the color management, performance could be lower. Maybe there could be a setting to turn in on/off.

For a proper color management, there is a need to detect monitor's ICC profile on all platforms or to allow user to select from a list of most common options.

scrubbbbs:
> As a side effect of the color management, performance could be lower. Maybe there could be a setting to turn in on/off.

You are right to be concerned, but the impact turns out to be quite minimal. While it takes some 100-200ms to convert large images, that isn't required if we double-buffer the drawing - then only the display surface needs to be converted.

For example, on a 1080p monitor the conversion happens (at worst) on ~1920x1080 image, regardless of the image resolution. In addition, the conversion itself is in-place and multi-threaded on the Qt side. This puts the viewport redraw around 10-20ms vs 1-2ms non-converted, on my machine at 2560x1440. And I think that would hardly be noticeable.

So I would propose we do not have a setting to turn it off for now as I do not expect any noticeable performance dip.

scrubbbbs:
> For a proper color management, there is a need to detect monitor's ICC profile on all platforms or to allow user to select from a list of most common options.

The detection part might be difficult to do and/or support, I suspect it is why Qt has not added support for this for many years now. I think that would need to be in unstable for quite a while.

So for now I have stubbed this into `DkImage::targetColorSpace()` to return sRGB since it is the most common case. Also, default windows/surfaces are generally assumed to be sRGB so even on wide gamut displays the conversion will give something more accurate that it has been.

Therefore I would at least like to try this for RC2 to see what the response is as I do not see any significant downsides here.

novomesk:
sRGB is probably most common case.

However when using Plasma with Wayland I saw some info that gamma 2.2 should be used, while running under X, sRGB is better.
( https://zamundaaa.github.io/wayland/2023/12/18/update-on-hdr-and-colormanagement-in-plasma.html https://planet.kde.org/xavers-blog-2024-07-15-how-to-profile-your-display-in-the-plasma-wayland-session/ )

P3 should be probably used on macOS

It's difficult to make to correct choice automatically, that's why I think it is good to allow user to configure it manually.

novomesk:
example of settings from [qView](https://github.com/jurplel/qView):

image

scrubbbbs:
> However when using Plasma with Wayland I saw some info that gamma 2.2 should be used, while running under X, sRGB is better. 

sRGB gamma is [very very nearly gamma 2.2](https://www.sjbrown.co.uk/posts/gamma-correct-rendering/). So I do not think we need to be concerned. But maybe it is possible to define a target with "true" gamma 2.2.

( https://zamundaaa.github.io/wayland/2023/12/18/update-on-hdr-and-colormanagement-in-plasma.html https://planet.kde.org/xavers-blog-2024-07-15-how-to-profile-your-display-in-the-plasma-wayland-session/ )
> 
> P3 should be probably used on macOS
> 
> It's difficult to make to correct choice automatically, that's why I think it is good to allow user to configure it manually.

This is trivial to add, would it be enough to get this into 3.22? I would skip the "Auto" option and only include the useful target colorspaces (sRGB,Display P3). User profiles are also easily added but I'm not sure how useful this is at the moment as we can't communicate this to the window server, it will assume sRGB/Display P3.

scrubbbbs:
The color management prefs UI. User ICC profiles are supported, listed after the built-in ones that are useful. Here I have selected a red/green swapped one I found.

image

image

novomesk:
sRGB and gamma 2.2 are close but some people inducing me notice the difference in dark areas.

Please add one more target with `QColorSpace::Primaries::SRgb` and 2.2f `QColorSpace::TransferFunction::Gamma`

and then when you feel comfortable, merge.

Afterwards, RC2 will be tagged.

scrubbbbs:
My understanding is that with SDR rendering (which we are constrained to at the moment in Qt), the non-piecewise gamma 2.2 will crush dark areas, e.g. a large swath of the low values map to 0 when the range is 0-255. But I'll go ahead and add it, it can't hurt. If we ever add HDR rendering then it will make more sense as it doesn't have that problem.

PR #1485, opened 2025-12-31
metadata: Add bugtracker and vcs-browser URLs

yakushabb:
Add bugtracker and vcs-browser URLs to show related pages.

More information: https://docs.flathub.org/docs/for-app-authors/metainfo-guidelines#url


novomesk:
Thanks for the contribution!

Issue #1493, opened 2026-01-09
Colormanagement for grayscale images doesn't work


novomesk:
Hi @scrubbbbs 

These images should look visually very similar:

https://raw.githubusercontent.com/nomacs/formats_testset/refs/heads/main/AVIF/profile_gray.avif

https://raw.githubusercontent.com/nomacs/formats_testset/refs/heads/main/AVIF/profile_gray_gamma22.avif

https://raw.githubusercontent.com/nomacs/formats_testset/refs/heads/main/AVIF/profile_gray_gamma28.avif

https://raw.githubusercontent.com/nomacs/formats_testset/refs/heads/main/AVIF/profile_gray_linear.avif

https://raw.githubusercontent.com/nomacs/formats_testset/refs/heads/main/AVIF/profile_gray_prophoto.avif

However, in nomacs 3.22.0, there is a big difference in brightness.

I think the problem is that grayscale QColorSpace cannot be used with RGB QImage::Format.

I think this Qt 6.8+ function `QImage QImage::convertedToColorSpace(targetColorSpace, QImage::Format_RGB32` should be used when converting from valid grayscale colorspace to RGB.

scrubbbbs:
> I think the problem is that grayscale QColorSpace cannot be used with RGB QImage::Format.

Yes. Apparently gray profiles cannot be converted (mathematically) to rgb profiles. I would like to try to keep images in their native format (so save-as and other features work accordingly). This patch appears to work but shifts brightness slightly with the custom trc case.

```c++
    // construct an rgb colorspace based on gray colorspace; cannot
    // be used to draw anything besides grayscale images
    if (srcColorSpace.colorModel() == QColorSpace::ColorModel::Gray) {
        if (srcColorSpace.transferFunction() == QColorSpace::TransferFunction::Custom) {
            // qt does not expose the profile transfer table, approximate one
            int N = 4096;
            QImage ramp(N, 1, QImage::Format_Grayscale16);
            ramp.setColorSpace(srcColorSpace);
            for (int i = 0; i < N; ++i) {
                double v = double(i) / (N - 1);
                ramp.setPixelColor(i, 0, QColor::fromRgbF(v, v, v));
            }
            ramp.convertToColorSpace(QColorSpace::SRgbLinear);
            QList<uint16_t> trc(N);
            for (int i = 0; i < N; ++i) {
                QColor c = ramp.pixelColor(i, 0);
[...]

PR #1501, opened 2026-01-14
Nicer slideshow fade and refactoring

scrubbbbs:
The current fade animation blends images with the window background color which gives an unpleasant brightness dip/pulse or strobe effect - the severity of which is proportional to the background darkness/brightness. If the background is not a shade of gray the color is shifted as well.

Ideally composited images should only blend with each other and in linear colorspace. This PR fixes the blending issue, the linear light issue seems too expensive to do in software rendering. There are 3 colorspace conversions needed, two of which could be cached to remedy this somewhat. However, I did not notice a big difference between linear vs display colorspace (as images are animated it can be hard to see however).

The solution here is fairly quick; only the bottom image of the fade must be drawn twice and only if the two images do not overlap completely.

Attached is a clip for comparison; the first is current nomacs and the second with the new fade. I added the vector scope to more easily see the "dip" in brightness and now smoother transition.

https://github.com/user-attachments/assets/02158341-e457-4ef1-a54b-bd4d68810ff1

### DIFF
--- a/ImageLounge/src/DkCore/DkImageStorage.cpp
+++ b/ImageLounge/src/DkCore/DkImageStorage.cpp
@@ -1314,7 +1314,9 @@ QImage::Format DkImage::renderFormat(QImage::Format imageFormat)
     case QImage::Format_RGBA64:
     case QImage::Format_Grayscale8:
     case QImage::Format_Grayscale16:
+#if QT_VERSION >= QT_VERSION_CHECK(6, 8, 0)
     case QImage::Format_CMYK8888:
+#endif
         // native formats for colorspace conversion - see: QImage::colorTransformed
         // premultiplied formats excluded since it saves some processing when color converting
         break;
@@ -1373,8 +1375,11 @@ QImage DkImage::convertToColorSpaceInPlace(const QColorSpace &target, QImage &im
     if (target == QColorSpace{}) {
         return img; // colorspace conversion is disabled
     }
-
+#if QT_VERSION < QT_VERSION_CHECK(6, 8, 0)
+    if (!target.isValid()) {
+#else
     if (!target.isValidTarget()) {
+#endif
         qWarning() << "[convertToColorSpace] invalid target" << target;
         return img;
     }
@@ -2027,8 +2032,11 @@ DkImageStorage::ScaledImage DkImageStorage::scaleImage(const QImage &src,
     scaled = src.scaled(size, Qt::IgnoreAspectRatio, mode);
 #endif
     Q_ASSERT(scaled.size() == size);
-
+#if QT_VERSION < QT_VERSION_CHECK(6, 8, 0)
+    if (colorSpace.isValid()) {
+#else
     if (colorSpace.isValidTarget()) {
+#endif
         scaled = DkImage::convertToColorSpaceInPlace(colorSpace, scaled);
         // Q_ASSERT(scaled.colorSpace() == colorSpace); // FIXME: sometimes fails due to qImage2Mat
     }


novomesk:
The two CR2 images (the fox and the bat) have wrong orientation/rotation.

ImageLounge/src/DkCore/DkBaseViewPort.h
```diff
@@ -195,8 +195,44 @@ public slots:
     QTimer *mZoomTimer;
     QImage mBackBuffer;
 
+    // flags to draw() call for multi-pass rendering or special cases
+    enum RenderFlag {
+        draw_image = 0, // only draw image
+        draw_background = 1, // draw background behind image
+        draw_pattern = 2, // draw transparency pattern above background, behind image
```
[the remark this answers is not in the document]
scrubbbbs: draw_image is 0 because it cannot be disabled; the image is always drawn

scrubbbbs:
Changed to draft. There are a bunch of issues and changes coming.

scrubbbbs:
I am ready to call this good now. I felt I needed to refactor some parts to get all of the features I wanted:
- proper crossfade
- no "pop-in" when animations switch between AA and non-AA images
- works correctly with transparency pattern enabled
- performs well with huge images and colorspace conversion

Along the way a few issues were fixed
- misalignment of AA/non-AA images
- misalignment of transparency pattern and image
- transparency pattern "crawls" when zooming
- color-managing transparency pattern
- check if transparency is actually used to make render decisions
- partially fixes #1493, still broken for downsampled case

ImageLounge/src/DkCore/DkBaseViewPort.h
```diff
@@ -203,6 +203,34 @@ public slots:
...
+     * @brief get parameters needed to draw image
+     * @param devicePixelRatio pixel ratio of widget we intend to paint on
+     * @param worldMatrix transformation from imgViewRect to viewport (zoom and pan)
+     * @param imgViewRect full image rectangle when scaled to fit widget
+     *
+     * @note This is intentionally static for use besides normal drawing
+     */
+    static RenderParams getRenderParams(double devicePixelRatio,
+                                        const QTransform &worldMatrix,
+                                        const QRectF &imgViewRect);
```
leejuyuu, line 232: This function seems to depend only on the three parameters. I suggest either making it into a non-member function or make `RenderParams` a class and `getRenderParams` its constructor.

ImageLounge/src/DkCore/DkBaseViewPort.cpp
```diff
@@ -516,44 +516,36 @@ DkBaseViewPort::RenderParams DkBaseViewPort::getRenderParams(double devicePixelR
...
-    const QRectF displayRect = worldMatrix.mapRect(imgViewRect); // use float size for QPainter
+    QRectF viewRect = worldMatrix.mapRect(imgViewRect);
 
-    // HiDPI: mImgViewRect is in logical pixels, we need physical for 100% scale and AA to work correctly
-    const QSizeF imgSizeF = displayRect.size() * devicePixelRatio;
+    // Round to physical/screen pixels to align resampled(AA) and original images
+    auto tx = QTransform::fromScale(devicePixelRatio, devicePixelRatio);
+    QRectF deviceRectF = tx.mapRect(viewRect);
+    double x = std::floor(deviceRectF.x());
+    double y = std::floor(deviceRectF.y());
```
leejuyuu: Why floor instead of round?

ImageLounge/src/DkCore/DkBaseViewPort.cpp
```diff
@@ -516,44 +516,36 @@ DkBaseViewPort::RenderParams DkBaseViewPort::getRenderParams(double devicePixelR
...
-    // the image may appear to shift slightly when toggling AA
-    QSize imgSize{qCeil(imgSizeF.width()), qCeil(imgSizeF.height())};
+    // rect to draw image, with world matrix disabled
+    viewRect = tx.inverted().mapRect(deviceRectF);
 
-    return {devicePixelRatio, imgViewRect, worldMatrix, displayRect, imgSize};
+    // rect to draw image, with world matrix enabled
+    QRectF dstRect = worldMatrix.inverted().mapRect(viewRect);
+
+    QRect deviceRect{int(x), int(y), int(w), int(h)};
```
leejuyuu: This could be

```cpp
QRect deviceRect = deviceRectF.toRect();
```

leejuyuu: Also, I noted that you started using brace initializations?

ImageLounge/src/DkCore/DkBaseViewPort.h
scrubbbbs: Yes my intent is for `RenderParams` concept to stand on it's own, but is not quite ready to be separated into another class, e.g. `renderComposite()`, while const, still depends on some virtual methods of `DkBaseViewPort`. At some point I think it will make sense to do the further refactoring needed to extract it. Maybe when we want to unlock new capabilities (maybe multi-pane views, opengl rendering, or frameless image being a real window).

ImageLounge/src/DkCore/DkBaseViewPort.cpp
scrubbbbs: I'm not sure, maybe it worked better. I'll take a look at it.

ImageLounge/src/DkCore/DkBaseViewPort.cpp
scrubbbbs: I guess I used std::floor() because I wanted to match the integer rect more precisely with the float rect, but that is not necessarily what is happening here.

ImageLounge/src/DkCore/DkBaseViewPort.cpp
scrubbbbs: It works fine to just round, I'm pushing changes

Issue #1500, opened 2026-01-14
Add options for initial zoom on loading new image


scrubbbbs:
> I was referring to the "always zoom to fit" checkbox which works perfectly in conjunction with the "keep zoom if the size is the same" option, as shown in my screenshot. I have the latest stable version.

_Originally posted by @tanakas323 in https://github.com/nomacs/nomacs/discussions/1498#discussioncomment-15491535_

3.22 changed "always zoom to fit" to a radio button, so it is no longer possible to have both keep-zoom-if-same-size and also scaling small images *up* to fit.

I think we need a new scheme for both keep zoom and size to fit that covers more use cases and is more usable by default.

The "size" considered in keep zoom should be more flexible:
- needs a fuzz factor so small size differences are considered the same, e.g. 10%
  
For keep zoom option:
- Always keep
- Keep if any dimensions match - default - now keeps if rotated
- Keep if same dimensions (current behavior)
- Keep if width is the same
- Keep if height is the same
- Keep if area is the same
- Never keep

For size to fit we also need an option:
- Fit longest side - default, current behavior
- Fit shortest side (Crop - nice for slideshows)
- Fixed zoom 100% (with optional percentage)
- Fit height (image only scrolls horizontally)
- Fit width (image only scrolls vertically)

And an option for how much to scale small images:
- Maximum zoom:
  - 100% - default
  - User -specified
  - None/Always fit

leejuyuu:
> have both keep-zoom-if-same-size and also scaling small images _up_ to fit.

Having the ability to combine would cause precedence problems. For example, if image 1 is smaller than the viewport already and was scaled to 90%, when loading image 2 that has the same size, should it zoom to fit the viewport, or should it keep the same zoom level? It is better to make the options mutually exclusive (as the current state).

leejuyuu:
I would like nomacs to stay as simple as possible. Here's the set of options that I think would suffice.

When loading new image

- Downscale images larger than viewport, otherwise render at 100%. (current zoom_never_keep)
- Scale to a fixed level (#865)
- Scale to fit inside the viewport (current zoom_to_fit)
- Scale to fill the viewport (#1222)
- Keep zoom (and position) with conditions (in sub-options), fall back to zoom_never_keep
  - Always
  - Has the same size (I think this is not intuitive (which is mentioned in #425), but it is an existing feature.) (2026-05-02 edit: confirmed that this option is actually used, see #1571)

scrubbbbs:
[...]

PR #1516, opened 2026-02-17
Zoom and viewport related fixes

### LINKED ISSUE #425: zoom 100% does not work properly
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

 - OS: [e.g. Ubuntu]

**Additional context**

### LINKED ISSUE #556: “Always keep zoom” setting should also keep position
When “Always keep zoom” is selected the viewer keeps the zoom setting but recenters on the image any time you switch between files. This should not be happening, it should keep the same position even when it switches to a different file.

1. Go to Settings/Display and select “Always keep zoom”
2. Get two images of the same size. For example, a nice big PNG image and the same image but converted to JPG.
3. Zoom on any part of the image to inspect the differences between the two files. Pressing Left/Right will switch between the images. However, each time the image is recentered. :(

### LINKED ISSUE #856: Default zoom setting does not affect frameless mode. (Always zoom to fit)
**Describe the bug**
Even if I set it to Always zoom to fit, it does not work in frameless mode.

**To Reproduce**
1. Go to 'setting'
2. Click on 'Always zoom to fit'
3. Change to 'frameless mode'

**Expected behavior**
The image always expands to fill the screen.

**Screenshots**
image

 - OS: [Fedora]
 - Version [36]

### LINKED ISSUE #1388: Overview slider has incorrect values
- The minimum is 0% which is invalid scale factor
- The maximum does not match the viewport
- The initial value does not match the viewport

### LINKED ISSUE #1507: Frameless image is slightly off-center
The image is off-center sometimes (3.22.0). This might be a window manager/Qt issue or related to zoom settings, because I do not see it on GNOME/Qt 6.10 but it shows up on xfce/Qt 6.2

This is easiest to see when scaling using the overview panel - which *should* scale about the center.

If you scale with mouse cursor over the image then the zoom is centered on the cursor as intended and it won't be centered as a result.

leejuyuu:
Fixes #556
Fixes #1507
Fixes #425
Fixes #856
Fixes #1388

### DIFF
--- a/ImageLounge/src/DkGui/DkViewPort.cpp
+++ b/ImageLounge/src/DkGui/DkViewPort.cpp
@@ -1989,8 +1989,6 @@ DkViewPortFrameless::DkViewPortFrameless(DkThumbLoader *thumbLoader, QWidget *pa
     mResetWhenZoomPastFit = false;
 }
 
-DkViewPortFrameless::~DkViewPortFrameless() = default;
-
 DkBaseViewPort::ZoomPos DkViewPortFrameless::calcZoomCenter(const QPointF &center, double /* unused */) const
 {
     QRectF viewRect = getImageViewRect();
@@ -2060,11 +2058,6 @@ void DkViewPortFrameless::resizeEvent(QResizeEvent *event)
 
         offset.setX(offset.x() + margin + iconSize.width());
     }
-}
-
-void DkViewPortFrameless::draw(QPainter &painter, double opacity, int flags)
-{
-    DkViewPort::draw(painter, opacity, flags);
 }
 
 void DkViewPortFrameless::eraseBackground(QPainter &painter) const
--- a/ImageLounge/src/DkGui/DkViewPort.h
+++ b/ImageLounge/src/DkGui/DkViewPort.h
@@ -232,7 +232,7 @@ class DllCoreExport DkViewPortFrameless : public DkViewPort
 
 public:
     explicit DkViewPortFrameless(DkThumbLoader *thumbLoader, QWidget *parent = nullptr);
-    ~DkViewPortFrameless() override;
+    ~DkViewPortFrameless() override = default;
 
 public slots:
     void moveViewInWidgetCoords(const QPointF &delta) override;
@@ -245,18 +245,16 @@ protected:
     void resizeEvent(QResizeEvent *event) override;
 
     // functions
-    void draw(QPainter &painter, double opacity, int flags) override;
     void drawFrame(QPainter &painter);
     void eraseBackground(QPainter &painter) const override;
 
-    // variables
+private:
     QVector<QAction *> mStartActions;
     QVector<QIcon> mStartIcons;
     QVector<QRectF> mStartActionsRects;
     QVector<QPixmap> mStartActionsIcons;
     QRectF mStartBgRect;
 
-private:
     [[nodiscard]] ZoomPos calcZoomCenter(const QPointF &center, double factor) const override;
     void controlImagePosition() override;
     void centerImage() override;


leejuyuu:
Preparing review...

scrubbbbs:
I am just starting to review with some testing, it seems to operate well but I found two regressions in frameless mode:

1. The translation/panning in frameless mode is no longer anchored to the mouse cursor, it is quite slow when zoomed out, and too fast when zoomed in. Also affects the overview panel. Does not seem to depend on the default zoom/fit setting.
2. The ALT key no longer makes the main menu visible

ImageLounge/src/DkCore/DkBaseViewPort.cpp
```diff
@@ -740,9 +740,12 @@ QTransform DkBaseViewPort::getScaledImageMatrix() const
 
 QTransform scaleKeepAspectRatioAndCenter(const QSizeF &src, const QSizeF &tgt, qreal paddingRatio)
 {
+    Q_ASSERT(paddingRatio >= 0 && paddingRatio < 1);
+
     qreal s = 1;
     if (!src.isNull()) {
-        s = std::min(tgt.width() / src.width(), tgt.height() / src.height());
+        const qreal r = 1 - paddingRatio;
+        s = std::min(tgt.width() * r / src.width(), tgt.height() * r / src.height());
```
scrubbbbs: Can you add a comment for what the padding ratio is and why we need it here? It could working around Qt bug that is no longer needed. My guess is the frameless window was overlapping window manager task bar at some point and this was the fix. Maybe this feature can be deleted rather than try to make it work (it is currently broken - see video below)

I did an experiment with `imageMatrixPaddingRatio` set to 0.25 to make it easier to see. It seems to do what it says - that is take some extra space around the image and viewport when fit-to-window is in effect or asked for manually (Ctrl+0).

But it behaves strangely, the padding will be completely ignored at certain sizes, notice the sudden jump and there is no padding in this clip:

https://github.com/user-attachments/assets/c2c04e28-8192-4293-8f65-a1e3bd888acb

I did a few tests with ratio set to 0 in frameless, it seems to work fine on Windows, macOS and Gnome/xorg.  The photos now fill the full screen instead of leaving gaps at the top and bottom.

If you want to keep the feature that's fine but this bug should be fixed. I can see the case for aesthetics here as it makes the window seem more like it is floating on the desktop. Or perhaps functionality as the padding gives a place to click through to the desktop.

ImageLounge/src/DkCore/DkBaseViewPort.cpp
```diff
@@ -740,9 +740,12 @@ QTransform DkBaseViewPort::getScaledImageMatrix() const
 
 QTransform scaleKeepAspectRatioAndCenter(const QSizeF &src, const QSizeF &tgt, qreal paddingRatio)
 {
+    Q_ASSERT(paddingRatio >= 0 && paddingRatio < 1);
+
     qreal s = 1;
     if (!src.isNull()) {
```
scrubbbbs: src.isNull() won't catch the case where either w/h is 0, they both have to be zero...could this possibly occur when zooming to a very small size?

ImageLounge/src/DkGui/DkViewPort.cpp
```diff
@@ -1173,17 +1173,9 @@ void DkViewPort::drawPolygon(QPainter &painter, const QPolygon &polygon)
...
-    centerImage();
-    changeCursor();
-
-    // mController->getOverview()->setViewPortRect(geometry());
-    mController->resize(width(), height());
-
-    return QGraphicsView::resizeEvent(event);
+    DkBaseViewPort::resizeEvent(event);
+    const QSize s = event->size();
+    mController->resize(s.width(), s.height());
```
scrubbbbs, line 1126: Bug: If the zoom is >100% with always-fit enabled, resizing the window does not resize the image. If <= 100% it works fine. There appears to be some test for when the zoom should follow the window resize that misses this case.

scrubbbbs:
Does not compile on Ubuntu 22.04, I'm still trying to keep things working there. Maybe we should raise minimum Qt to 6.4 in cmake?

```
../ImageLounge/src/DkCore/DkBaseViewPort.cpp: In static member function ‘static nmc::DkBaseViewPort::RenderParams nmc::DkBaseViewPort::getRenderParams(double, const QTransform&, const QRectF&)’:
../ImageLounge/src/DkCore/DkBaseViewPort.cpp:547:49: error: ‘class QRect’ has no member named ‘toRectF’; did you mean ‘QRect’?
  547 |     viewRect = tx.inverted().mapRect(deviceRect.toRectF());
```

novomesk:
Regarding required Qt version, 6.5 will be End of Support soon:

https://www.qt.io/blog/qt-6.5-reaches-end-of-support

KDE frameworks required 6.8 already.

Oldest Qt version in Gentoo Linux is 6.10.1 now.

scrubbbbs:
> Regarding required Qt version, 6.5 will be End of Support soon:

While Ubuntu 24.04 is stuck on 6.4.2 through 2029. As someone who doesn't upgrade the OS every few months or use a rolling release I can appreciate supporting old versions - so I don't mind it so much. I would like to get c++20 as a requirement however which would make 24.04 the minimum AFACT.

ImageLounge/src/DkCore/DkBaseViewPort.cpp
leejuyuu: Was this video taken with the frameless view? I think the padding might not work well with DkViewPort. DkViewPortFrameless is quite different in the panning and scaling logic.

I would like to remove this, just wasn't sure about it because it was included in a bug fix. Given that this does not seem to have an observable problem, maybe we can do this.

leejuyuu:
Thanks for testing!

> The translation/panning in frameless mode is no longer anchored to the mouse cursor

bisection points to 41d85c4

> The ALT key no longer makes the main menu visible

(This happens for frameless.) I am not very short, but ffec722 seems to be the first bad commit.

ImageLounge/src/DkCore/DkBaseViewPort.cpp
leejuyuu: Yes, this might happen. I use isValid() now.

ImageLounge/src/DkCore/DkBaseViewPort.cpp
scrubbbbs: Yes, I had set it on the base viewport. The frameless window is not resizable it is less likely but still a possiblity as it seems to depend on the viewport and image dimensions.

I support removing it, testing shows it is not needed.

ImageLounge/src/DkGui/DkViewPort.cpp
leejuyuu: Nice catch! This seems to be an old bug.

scrubbbbs:
> > The ALT key no longer makes the main menu visible
> 
> (This happens for frameless.) I am not very short, but [ffec722](https://github.com/nomacs/nomacs/commit/ffec7221ddda611cd92b472fc0d854f99b17ec52) seems to be the first bad commit.

Yes, `keyReleaseEvent()` cannot be removed here, despite the unused signal, it also has another function; to propagate the event around QGraphicsView which wants to eat it.

ImageLounge/src/DkGui/DkViewPort.cpp
leejuyuu: This is not specific to always fit. When the zoom level is different from the natural state (max(100%, viewport size)), updateImageMatrix() will restore the zoom level after operation, so it looks like the image does not shrink with the viewport. After thinking which should be the desired behavior, I would like this to be configurable. 

I suggest leave this as a future work.

leejuyuu:
I've addressed all the issues. @scrubbbbs, please take a look. Also, are you familiar with the build error for macOS?

ImageLounge/src/DkGui/DkViewPort.cpp
scrubbbbs: Configurable, sure. The issue is consistency, the window resize behavior should not change based on the size of the image being viewed. Either it is pinned to the viewport size or it isn't. As it is an old bug I suppose delaying to fix is fine. I will open an issue for it.

scrubbbbs:
> Also, are you familiar with the build error for macOS?

This is fixed in master now.

scrubbbbs:
When dragging the image in frameless, the cursor does not change back to pointer on mouse-up. Also in old nomacs.

Have you thought about anything we could do to hold the translation when switching between portrait and landscape photos? This is quite common because photos are usually the same size, just rotated. For example, if the image clips one of the edges of the viewport it could be shifted away from the edge rather than always centering on the viewport.

> One more issue I found with zoom to cursor, when zooming in or out, the image always re-centers in the viewport. It seems to happen either at 100% or the autofit size.

Did you take a look at this? Note in these clips my cursor was over the birds/bats eye the entire time.

https://github.com/user-attachments/assets/2686dfea-5b98-40a3-8b73-8026ed524537

https://github.com/user-attachments/assets/049def6b-3d2a-4b1d-a988-e6d6ec2fd5b8

leejuyuu:
> When dragging the image in frameless, the cursor does not change back to pointer on mouse-up. Also in old nomacs.

Should be fixed now.

> Have you thought about anything we could do to hold the translation when switching between portrait and landscape photos? This is quite common because photos are usually the same size, just rotated. For example, if the image clips one of the edges of the viewport it could be shifted away from the edge rather than always centering on the viewport.

Could you describe more the reason for this? When the sizes differ, handling the translation with re-centering might be the most predictable behavior for the user.

> > One more issue I found with zoom to cursor, when zooming in or out, the image always re-centers in the viewport. It seems to happen either at 100% or the autofit size.
> 
> Did you take a look at this? Note in these clips my cursor was over the birds/bats eye the entire time.
> Screencast.from.2026-03-01.10-06-39.mp4
> Screencast.from.2026-03-01.10-08-52.mp4

Thanks for catching this. I think it is fixed now.

scrubbbbs:
> > Have you thought about anything we could do to hold the translation when switching between portrait and landscape photos? This is quite common because photos are usually the same size, just rotated. For example, if the image clips one of the edges of the viewport it could be shifted away from the edge rather than always centering on the viewport.
> 
> Could you describe more the reason for this?

The primary use case is for frameless mode, where we cannot simply move the window to where we want the image on screen.

There are several mentions in issues about moving the frameless image around, and why it won't drag to another monitor, etc. This seems like something people want to do often. The only explanation I have for that is they want to have something else in the center of the monitor while keeping the frameless image in it's own spot off to the side. Perhaps as a reference image, background slideshow, or just for aesthetics as that is what frameless is geared towards.

>  When the sizes differ, handling the translation with re-centering might be the most predictable behavior for the user.

Since the user had to place the image where they wanted, I would think having the image sometimes stay-put and sometimes move back to center could be frustrating.

leejuyuu:
For the frameless mode, perhaps we should consider changing the implementation. The current implementation is a maximized window with a transparent background, and we handle the panning inside DkViewPortFrameless. I am thinking about something like a draggable window that always fits the image size. This could solve the cross monitor issue, and eliminate the need to remember the pan. What do you think?

scrubbbbs:
Yes, that is the ultimate solution! I would go that route but I see a lot of hurdles to overcome. Maybe we have to drop a lot of features in frameless mode as a result. I'm not sure how well this will play with all the different window managers on Linux. On Windows, macOS, and traditional Linux WMs it should be predictable. However, modern WMs seem to think that applications only need to have one window like a phone application.

scrubbbbs:
Confirmed last two issues fixed. But we've lost translation with the always-keep-zoom mode. Translation is working with keep-if-same-size mode.

scrubbbbs:
Another regression, when using the explorer to switch images rather than arrow keys, the animation picks up the old transformation/scale before the last zoom/resize.

1. Press a fit button (or zoom with scroll wheel), use arrows to move to next image, works fine
2. Press different fit button, click in DKExplorer. Uses the old transformation, and remains wrong until an arrow key is pressed.

This also applies when using File->Open.

https://github.com/user-attachments/assets/4e14d160-8f8a-4e71-8cd9-3f05bf4f74ca

leejuyuu:
> Confirmed last two issues fixed. But we've lost translation with the always-keep-zoom mode. Translation is working with keep-if-same-size mode.

Translation was not kept in 3.22.0. This branch keeps translation for zoom_always_keep only if the size remains the same. 

> Another regression, when using the explorer to switch images rather than arrow keys, the animation picks up the old transformation/scale before the last zoom/resize.

I cannot reproduce on 96b394b.

https://github.com/user-attachments/assets/c380d4da-7f52-4747-a9c0-a10b8e9188d7

scrubbbbs:
> I cannot reproduce on [96b394b]

Yes, this is invalid. I somehow wasn't on the correct commit.

> > [...] But we've lost translation with the always-keep-zoom mode.[...]
>>
> Translation was not kept in 3.22.0. This branch keeps translation for zoom_always_keep only if the size remains the same.

The image size is remaining the same in my test. I found the problem: the translation only sticks if the image has a zoom factor != 1.0 and is sufficiently large (maybe larger than viewport?). If the image is smaller than the viewport at 100% it won't keep translation.

This has an another interesting side-effect. If the image is != 100% and moved, then changed back to 100%, it now cannot be dragged to a new position, it always returns back to where it was before changing to 100%.

https://github.com/user-attachments/assets/c9914f74-f161-4299-bbe1-d56afefd873d

scrubbbbs:
On my previous observation, it is likely related to your previous comment "When the zoom level is different from the natural state (max(100%, viewport size)), updateImageMatrix() will restore the zoom level after operation"

That is there is some bifurcation of the logic when image is at 100% scale and smaller than the viewport. You wanted to push this for later work but if it is affecting more than just window resizing should we tackle it now?

scrubbbbs:
A few more nitpicks.

We should add tooltips for the different modes, there is nothing on three of these right now. My suggestion:
- Always keep zoom: Always keep zoom percentage. Keep position if images size is unchanged.
- Keep zoom if size is the same: Keep zoom and position if image size is unchanged. Scale small images up to 100%, large images to fit.
- Never keep: Scale small images up to 100%, large images to fit.
- Always fit: Scale small and large images to fit.

In sync mode (also in old nomacs) when passing 100%/viewport-size there is a discontinuity (tested in always-keep mode). Similar to what I was seeing elsewhere. In old nomacs sync does not work well for comparing images with different sizes anyways so we can push this one.

https://github.com/user-attachments/assets/303e525b-9de6-4947-9ff5-4366fcb397dc

leejuyuu:
> > 
> 
> The image size is remaining the same in my test. I found the problem: the translation only sticks if the image has a zoom factor != 1.0 and is sufficiently large (maybe larger than viewport?). If the image is smaller than the viewport at 100% it won't keep translation.

Let me confirm, this only happens in frameless mode for you?
I can reproduce with frameless mode, only for images smaller than the viewport.

> On my previous observation, it is likely related to your previous comment "When the zoom level is different from the natural state (max(100%, viewport size)), updateImageMatrix() will restore the zoom level after operation"
> 
> That is there is some bifurcation of the logic when image is at 100% scale and smaller than the viewport. You wanted to push this for later work but if it is affecting more than just window resizing should we tackle it now?

Part of it is, part of it isn't. The major cause of the above issue is that the translation of frameless is applied to mImageMatrix, while in the normal viewport, it is applied to mWorldMatrix. However, mImageMatrix resets when loading new images, hence the reset of translation. I am going to try fixing this first.

The use of two matrices for the transformation is quite interesting. I assume it was supposed to simplify some operation. When mWorldMatrix is unit, mImageMatrix would make the image fit the viewport if it is larger, or render at 100% in the center otherwise. This is the source of different logic. I would like to combine these two; however, they are used in many places, so it might not be easy at this moment.

leejuyuu:
> We should add tooltips for the different modes, there is nothing on three of these right now. My suggestion:

These are subject to change when we do https://github.com/nomacs/nomacs/issues/1500. I suggest postponing until then.


> In sync mode (also in old nomacs) when passing 100%/viewport-size there is a discontinuity (tested in always-keep mode). Similar to what I was seeing elsewhere. In old nomacs sync does not work well for comparing images with different sizes anyways so we can push this one.

I can't reproduce this. Did you observe any dependence on the image size?
However, I notice there's a bug that the zoom level in the status bar and overview does not update properly in the video.

scrubbbbs:
> Let me confirm, this only happens in frameless mode for you?

Yes, but possibly because there is no translation in normal mode (that is it is locked to center below the viewport size)

> The use of two matrices for the transformation is quite interesting.

Reminds me of OpenGL with its separate model-view and projection matrix. A lot of APIs have a separate matrix for model and view as well. Probably makes some things simpler to reason about, e.g. "scene" vs "view" vs screen coordinates.

scrubbbbs:
> I can't reproduce this. Did you observe any dependence on the image size?

If the images are different sizes (but same aspect ratio), is the main problem. If they are the same size it doesn't seem to matter.

> However, I notice there's a bug that the zoom level in the status bar and overview does not update properly in the video.

Seems to be old quirk with sync, I'm noticing basic things that do not sync that seem to be missing:
- overview controls
- navigation buttons

leejuyuu:
> > Let me confirm, this only happens in frameless mode for you?
> 
> Yes, but possibly because there is no translation in normal mode (that is it is locked to center below the viewport size)

Thanks. I think I fixed this.

leejuyuu:
> > I can't reproduce this. Did you observe any dependence on the image size?
> 
> If the images are different sizes (but same aspect ratio), is the main problem. If they are the same size it doesn't seem to matter.

Thanks for the hint. I can reproduce now. 

I intend to focus on the viewport a bit more and see if I can fix these strange problems. However, as this branch is getting quite long, I suggest we merge first. What do you think?

The most pressing problem is that the viewport logic is fairly complicated, but only testable by hand. I plan to decompose the viewport classes into Views (the actual widgets) and ModelViews (contain abstracted UI state and logic), similar to MVVM design, so that the ModelView part could be unit tested. This is likely going to be a big refactor.

scrubbbbs:
I went ahead and merged. Everything working except for:
- resize window issue (#1529)
- sync issue

> This is likely going to be a big refactor.

Probably worthwhile as it is difficult to test, and we want to add more features. Since you have spent a lot of time in viewport code by now you probably understand the issues better than anyone now. I think you should also consider the theoretical sub-window frameless mode when deciding on the design as well.

PR #1521, opened 2026-02-19
Adjustments Optimizations and Improvements

scrubbbbs:
This PR includes revisions of all adjustments defined in DkImage with several broad objectives:
- keep image pixel format the same if we can
- add multithreading to speed them up
- use less memory if possible

Pixel format support is expanded via several approaches:
- use QImage/QPainter operations if they are available (cropping, background fill etc)
- created DkNativeImage to handle conversions to/from cv::Mat  with more pixel formats besides ARGB/BGR888
- created DkImageProc to manage writing our own per-pixel processing kernels on multiple formats

UI changes:
- layer progress bar above the viewport so it doesn't push the image down and back up again when adjusting
- drop the preview image from edit dock

Some changes to adjustment algos (not much):
- auto adjust now clips both ends of histogram, previously only the high bins (highlights) were clipped. This is more in-line with other image editors
- grayscale uses linear light instead of L'ab, sets color profiles and outputs 16-bit grayscale. This gives a more neutral and adjustable output
- rescale "gamma correction" uses correct linear conversion and back to source colorspace - previously assumed sRGB.
- hue slider moves through +/- 180 degrees hue, previously +/- 360 (though the slider value was 180)

### DIFF
--- a/ImageLounge/src/DkCore/DkImageStorage.cpp
+++ b/ImageLounge/src/DkCore/DkImageStorage.cpp
@@ -194,10 +194,17 @@ QImage DkImage::resizeImage(const QImage &src,
         auto ipl = interpolation == ipl_nearest ? Qt::FastTransformation : Qt::SmoothTransformation;
         QImage outImg = inImg.scaled(nSize, Qt::IgnoreAspectRatio, ipl);
 #endif
+        // Output indexed to ARGB32 for consistency with other manipulators.
+        // Also we have no dither options so it looks bad, usually.
+        QImage::Format outFormat = src.format();
+        if (outFormat == QImage::Format_Indexed8) {
+            outFormat = DkImage::alphaChannelUsed(src) ? QImage::Format_ARGB32 : QImage::Format_BGR888;
+        }
+
         if (correctGamma) {
-            outImg = convertToColorSpace(outImg, src.format(), src.colorSpace());
+            outImg = convertToColorSpace(outImg, outFormat, src.colorSpace());
         } else {
-            outImg.convertTo(src.format());
+            outImg.convertTo(outFormat);
         }
         return outImg;
     } catch (...) {


ImageLounge/src/DkCore/DkImageStorage.cpp
```diff
@@ -539,490 +481,1025 @@ QImage rotateImageFast(const QImage &img, double angle)
...
-    for (int rIdx = 0; rIdx < img.height(); rIdx++) {
-        for (int cIdx = 0; cIdx < bpl; cIdx++, mPtr++) {
-            if (hasAlpha && cIdx % 4 == 3)
-                continue;
+    // FIXME: getAngle() is imprecise for tiny crops (1x1, 2x2 etc), so we need large epsilon
+    // This works, but requires angle has limited precision (in degrees, 2 decimal places)
+    const double epsilon = 1e-5;
+
+    double radians = rect.getAngle();
+    double rightAngle = radians / (CV_PI * 0.5);
```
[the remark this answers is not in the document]
scrubbbbs: CV_PI is defined in DkMath.h if OpenCV isn't available

ImageLounge/src/DkCore/DkManipulatorsIpl.cpp
```diff
@@ -261,9 +261,11 @@ DkUnsharpMaskManipulator::DkUnsharpMaskManipulator(QAction *action)
...
 QImage DkUnsharpMaskManipulator::apply(const QImage &img) const
 {
-    QImage imgC = img.copy();
-    DkImage::unsharpMask(imgC, (float)sigma(), 1.0f + amount() / 100.0f);
-    return imgC;
+    QImage tmp = img;
+    if (DkImage::unsharpMask(tmp, (float)sigma(), 1.0f + amount() / 100.0f)) {
+        return tmp;
+    }
+    return {};
```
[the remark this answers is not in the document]
scrubbbbs: blur/unsharp would only return false on an actual failure, a "no-op" would be invalid input, which should not be ignored.

leejuyuu:
Compiler warnings

```
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp: In instantiation of ‘void nmc::normalize(cv::Mat&, const DkWorkRange&, const std::array<typename Format::ChannelType, 3>&, const std::array<typename Format::ChannelType, 3>&) [with Format = PixFormat<nmc::ImgType::gray, unsigned char, 1>; typename Format::ChannelType = unsigned char]’:
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:891:30:   required from ‘static bool nmc::DkNormalizeKernel::kernel(const std::any&, const nmc::DkWorkRange&) [with Format = nmc::PixFormat<nmc::ImgType::gray, unsigned char, 1>]’
  891 |             normalize<Format>(mat, slice, {mn, mn, mn}, {mx, mx, mx});
      |             ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageProc.h:373:47:   required from ‘static constexpr nmc::DkKernelBase::DispatchTable nmc::DkKernelBase::makeTable(FmtList) [with Kernel = nmc::DkNormalizeKernel; DispatchTable = std::array<bool (*)(const std::any&, const nmc::DkWorkRange&), 9>; FmtList = std::array<nmc::ImgFmt, 9>]’
  373 |             case ImgFmt::Gray8:    table[i] = &Kernel::template kernel<PixFmt_Gray8>; break;
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:899:73:   required from here
  899 |     static constexpr DispatchTable kTable = makeTable<DkNormalizeKernel>(kFormats);
      |                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:825:20: warning: variable ‘isShort’ set but not used [-Wunused-but-set-variable]
  825 |     constexpr bool isShort = std::is_same_v<ChannelType, uint16_t>;
      |                    ^~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp: In instantiation of ‘void nmc::normalize(cv::Mat&, const DkWorkRange&, const std::array<typename Format::ChannelType, 3>&, const std::array<typename Format::ChannelType, 3>&) [with Format = PixFormat<nmc::ImgType::bgr, unsigned char, 3>; typename Format::ChannelType = unsigned char]’:
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:891:30:   required from ‘static bool nmc::DkNormalizeKernel::kernel(const std::any&, const nmc::DkWorkRange&) [with Format = nmc::PixFormat<nmc::ImgType::bgr, unsigned char, 3>]’
  891 |             normalize<Format>(mat, slice, {mn, mn, mn}, {mx, mx, mx});
      |             ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageProc.h:375:47:   required from ‘static constexpr nmc::DkKernelBase::DispatchTable nmc::DkKernelBase::makeTable(FmtList) [with Kernel = nmc::DkNormalizeKernel; DispatchTable = std::array<bool (*)(const std::any&, const nmc::DkWorkRange&), 9>; FmtList = std::array<nmc::ImgFmt, 9>]’
  375 |             case ImgFmt::BGR888:   table[i] = &Kernel::template kernel<PixFmt_BGR888>; break;
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:899:73:   required from here
  899 |     static constexpr DispatchTable kTable = makeTable<DkNormalizeKernel>(kFormats);
      |                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:825:20: warning: variable ‘isShort’ set but not used [-Wunused-but-set-variable]
  825 |     constexpr bool isShort = std::is_same_v<ChannelType, uint16_t>;
      |                    ^~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp: In instantiation of ‘void nmc::normalize(cv::Mat&, const DkWorkRange&, const std::array<typename Format::ChannelType, 3>&, const std::array<typename Format::ChannelType, 3>&) [with Format = PixFormat<nmc::ImgType::rgb, unsigned char, 3>; typename Format::ChannelType = unsigned char]’:
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:891:30:   required from ‘static bool nmc::DkNormalizeKernel::kernel(const std::any&, const nmc::DkWorkRange&) [with Format = nmc::PixFormat<nmc::ImgType::rgb, unsigned char, 3>]’
  891 |             normalize<Format>(mat, slice, {mn, mn, mn}, {mx, mx, mx});
      |             ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageProc.h:376:47:   required from ‘static constexpr nmc::DkKernelBase::DispatchTable nmc::DkKernelBase::makeTable(FmtList) [with Kernel = nmc::DkNormalizeKernel; DispatchTable = std::array<bool (*)(const std::any&, const nmc::DkWorkRange&), 9>; FmtList = std::array<nmc::ImgFmt, 9>]’
  376 |             case ImgFmt::RGB888:   table[i] = &Kernel::template kernel<PixFmt_RGB888>; break;
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:899:73:   required from here
  899 |     static constexpr DispatchTable kTable = makeTable<DkNormalizeKernel>(kFormats);
      |                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:825:20: warning: variable ‘isShort’ set but not used [-Wunused-but-set-variable]
  825 |     constexpr bool isShort = std::is_same_v<ChannelType, uint16_t>;
      |                    ^~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp: In instantiation of ‘void nmc::normalize(cv::Mat&, const DkWorkRange&, const std::array<typename Format::ChannelType, 3>&, const std::array<typename Format::ChannelType, 3>&) [with Format = PixFormat<nmc::ImgType::bgr, unsigned char, 4>; typename Format::ChannelType = unsigned char]’:
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:891:30:   required from ‘static bool nmc::DkNormalizeKernel::kernel(const std::any&, const nmc::DkWorkRange&) [with Format = nmc::PixFormat<nmc::ImgType::bgr, unsigned char, 4>]’
  891 |             normalize<Format>(mat, slice, {mn, mn, mn}, {mx, mx, mx});
      |             ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageProc.h:377:47:   required from ‘static constexpr nmc::DkKernelBase::DispatchTable nmc::DkKernelBase::makeTable(FmtList) [with Kernel = nmc::DkNormalizeKernel; DispatchTable = std::array<bool (*)(const std::any&, const nmc::DkWorkRange&), 9>; FmtList = std::array<nmc::ImgFmt, 9>]’
  377 |             case ImgFmt::ARGB32:   table[i] = &Kernel::template kernel<PixFmt_ARGB32>; break;
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:899:73:   required from here
  899 |     static constexpr DispatchTable kTable = makeTable<DkNormalizeKernel>(kFormats);
      |                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:825:20: warning: variable ‘isShort’ set but not used [-Wunused-but-set-variable]
  825 |     constexpr bool isShort = std::is_same_v<ChannelType, uint16_t>;
      |                    ^~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp: In instantiation of ‘void nmc::normalize(cv::Mat&, const DkWorkRange&, const std::array<typename Format::ChannelType, 3>&, const std::array<typename Format::ChannelType, 3>&) [with Format = PixFormat<nmc::ImgType::rgb, unsigned char, 4>; typename Format::ChannelType = unsigned char]’:
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:891:30:   required from ‘static bool nmc::DkNormalizeKernel::kernel(const std::any&, const nmc::DkWorkRange&) [with Format = nmc::PixFormat<nmc::ImgType::rgb, unsigned char, 4>]’
  891 |             normalize<Format>(mat, slice, {mn, mn, mn}, {mx, mx, mx});
      |             ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageProc.h:378:47:   required from ‘static constexpr nmc::DkKernelBase::DispatchTable nmc::DkKernelBase::makeTable(FmtList) [with Kernel = nmc::DkNormalizeKernel; DispatchTable = std::array<bool (*)(const std::any&, const nmc::DkWorkRange&), 9>; FmtList = std::array<nmc::ImgFmt, 9>]’
  378 |             case ImgFmt::RGBA8888: table[i] = &Kernel::template kernel<PixFmt_RGBA8888>; break;
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:899:73:   required from here
  899 |     static constexpr DispatchTable kTable = makeTable<DkNormalizeKernel>(kFormats);
      |                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:825:20: warning: variable ‘isShort’ set but not used [-Wunused-but-set-variable]
  825 |     constexpr bool isShort = std::is_same_v<ChannelType, uint16_t>;
      |                    ^~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp: In instantiation of ‘void nmc::normalize(cv::Mat&, const DkWorkRange&, const std::array<typename Format::ChannelType, 3>&, const std::array<typename Format::ChannelType, 3>&) [with Format = PixFormat<nmc::ImgType::rgb, float, 4>; typename Format::ChannelType = float]’:
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:891:30:   required from ‘static bool nmc::DkNormalizeKernel::kernel(const std::any&, const nmc::DkWorkRange&) [with Format = nmc::PixFormat<nmc::ImgType::rgb, float, 4>]’
  891 |             normalize<Format>(mat, slice, {mn, mn, mn}, {mx, mx, mx});
      |             ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageProc.h:380:47:   required from ‘static constexpr nmc::DkKernelBase::DispatchTable nmc::DkKernelBase::makeTable(FmtList) [with Kernel = nmc::DkNormalizeKernel; DispatchTable = std::array<bool (*)(const std::any&, const nmc::DkWorkRange&), 9>; FmtList = std::array<nmc::ImgFmt, 9>]’
  380 |             case ImgFmt::RGBAFP32: table[i] = &Kernel::template kernel<PixFmt_RGBAFP32>; break;
      |                                               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:899:73:   required from here
  899 |     static constexpr DispatchTable kTable = makeTable<DkNormalizeKernel>(kFormats);
      |                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:824:20: warning: variable ‘isByte’ set but not used [-Wunused-but-set-variable]
  824 |     constexpr bool isByte = std::is_same_v<ChannelType, uint8_t>;
      |                    ^~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkImageStorage.cpp:825:20: warning: variable ‘isShort’ set but not used [-Wunused-but-set-variable]
  825 |     constexpr bool isShort = std::is_same_v<ChannelType, uint16_t>;
      |                    ^~~~~~~
```

scrubbbbs:
> Compiler warnings

Thanks, I did not see that anywhere else, what gcc version is this?

ImageLounge/src/DkCore/DkNativeImage.h
```diff
@@ -0,0 +1,119 @@
...
+
+namespace nmc
+{
+#ifdef WITH_OPENCV
+
+// container for image in a "native" format supported by nomacs, which
+// for the moment means it can map directly to/from cv::Mat
+class DkNativeImage
+{
+protected:
```
leejuyuu: Since we use inheritance, I suggest making members private by default. This enforces stronger encapsulation and makes refactoring easier.

ImageLounge/src/DkCore/DkNativeImage.h
```diff
@@ -0,0 +1,119 @@
...
+#include "opencv2/core/mat.hpp"
+#endif
+
+namespace nmc
+{
+#ifdef WITH_OPENCV
+
+// container for image in a "native" format supported by nomacs, which
+// for the moment means it can map directly to/from cv::Mat
+class DkNativeImage
```
leejuyuu, line 23: "Native" seems to be more commonly associated with OSs or platforms. As this class holds QImage and cv::Mat and does not create a new format, maybe it should be named `DkImageConverter`, `DkImageBridge`, or `DkUniversalImage` to reflect its purpose.

ImageLounge/src/DkCore/DkNativeImage.h
```diff
@@ -0,0 +1,119 @@
...
+    const QImage &img() const // return const; any write to mImg will break link with mat
+    {
+        return mImg;
+    }
+
+    // note: if constructed from QImage, must use .clone() to persist a copy
+    cv::Mat &mat()
+    {
+        return mMat;
+    }
```
leejuyuu, line 70: Seems dangerous to return a mutable reference to an internal member.

ImageLounge/src/DkCore/DkMath.cpp
```diff
@@ -262,8 +262,8 @@ void DkRotatingRect::setCenter(const QPointF &center)
 double DkRotatingRect::getAngle() const
 {
     // default upper left corner is 0
-    DkVector xV = mRect[1] - mRect[0];
-    return xV.angle();
+    QPointF vec = mRect[1] - mRect[0];
+    return atan2(vec.y(), vec.x());
```
leejuyuu: Mark std call with namespace.

```suggestion
    return std::atan2(vec.y(), vec.x());
```

ImageLounge/src/DkCore/DkImageProc.h
```diff
@@ -0,0 +1,517 @@
...
+    case ImgFmt::BGR888:   return QImage::Format_BGR888;
+    case ImgFmt::RGB888:   return QImage::Format_RGB888;
+    case ImgFmt::ARGB32:   return QImage::Format_ARGB32;
+    case ImgFmt::RGBA8888: return QImage::Format_RGBA8888;
+    case ImgFmt::RGBA64:   return QImage::Format_RGBA64;
+    case ImgFmt::RGBAFP32: return QImage::Format_RGBA32FPx4;
+    default:               return QImage::Format_Invalid;
+    }
+    // clang-format on
+}
```
leejuyuu, line 251: Nitpick: I actually like these to format normally, making this side-by-side makes it easy to misalign different branches. Maybe it's just how I read.

ImageLounge/src/DkCore/DkImageProc.h
```diff
@@ -0,0 +1,517 @@
...
+#endif
+
+    static bool isCompatibleWith(const QImage &img)
+    {
+        return sizeof(T) * channels == img.depth() / 8;
+    }
+
+    // load pixel to float {r,g,b,a} [0,1]
+    // HDR (fp32) images may have values outside of [0,1]
+    static std::tuple<float, float, float, float> loadFloat(const T *rgba)
```
leejuyuu: Replacing the tuple with a struct might be clearer.

ImageLounge/src/DkCore/DkImageStorage.cpp
```diff
@@ -969,68 +969,161 @@ QImage DkImage::cropToImage(const QImage &src, const DkRotatingRect &rect, const
...
-            iPtr[cIdx + 2] = (unsigned char)v;
-
-            // adopt saturation
-            int s = qRound(iPtr[cIdx + 1] * satN);
-            if (s < 0)
-                s = 0;
-            if (s > 255)
-                s = 255;
-            iPtr[cIdx + 1] = (unsigned char)s;
+    static std::tuple<float, float, float> hsvToRgb(float h, float s, float v)
```
leejuyuu: This is why I think they should be structs; the input and output have exactly the same pattern.

ImageLounge/src/DkCore/DkImageProc.h
```diff
@@ -0,0 +1,517 @@
...
+
+    QList<DkWorkRange> partition() const // partiton by cpu thread count
+    {
+        QList<DkWorkRange> parts;
+        int partSize = qMax(1, (end - begin) / QThreadPool::globalInstance()->maxThreadCount());
+        for (int y0 = begin; y0 < end; y0 += partSize) {
+            parts.append({y0, qMin(end, y0 + partSize)});
+        }
+        return parts;
+    }
```
leejuyuu, line 270: This attempts to saturate the machine by default, regardless of whether other processes are running? Maybe we need to bound this to a lower count.

leejuyuu (commented):
I've tested a bit, seems to have no problem for me.

I want to take a deeper look at the DkNativeImage class. Sharing memory is a great optimization, but given the footguns of C++, I want to make sure I understand the lifetime of the values involved.

I have not come to appreciate the need for the DkImageProc yet. I will think more about it. Also, are you aware that OpenCV might already be using multi-threading under the hood? It should be supported, but I am not sure Nomacs has those configured.

scrubbbbs:
> I want to take a deeper look at the DkNativeImage class. Sharing memory is a great optimization, but given the footguns of C++, I want to make sure I understand the lifetime of the values involved.

The lifetime issue is when trying to separate the mat/image from the nativeImage. The nativeImage owns the shared buffer, and the mat/image members must be treated as views to that buffer.

> I have not come to appreciate the need for the DkImageProc yet. I will think more about it. Also, are you aware that OpenCV might already be using multi-threading under the hood? It should be supported, but I am not sure Nomacs has those configured.

I wanted to use OpenCV methods. I did change the blur/unsharp to use the MP version for example. CV methods were dropped only when they could not meet the requirements. Commonly the alpha channel is not supported or format support is not all there. Sometimes there was no MP method or the MP approach needs multiple conversion passes and/or 2-3 times more memory.

ImageLounge/src/DkCore/DkNativeImage.h
scrubbbbs: This was a struggle. I ended up with "native" image as it is "native to nomacs" or that was the concept. DkImage or NomacsImage would also be suitable.

I originally named it ImageView as it gives a view to a qimage...but as it also owns the memory and can convert images this is not quite the same thing.

- ImageConverter: it doesn't always convert the image
- ImageBridge: closer
- ImageAdapter:
- ImageWrapper:
- DkImage - taken

ImageLounge/src/DkCore/DkNativeImage.h
scrubbbbs: It must be mutable of course or it can't be used for anything. As to why it is  a reference, the thinking was to prevent some misuse of the member. As returning `cv::Mat` by value would give a shallow copy, returning a reference makes it more obvious that it's lifetime is tied to the native image.

ImageLounge/src/DkCore/DkImageProc.h
scrubbbbs: I am the opposite, it is easier to scan some things when formatted like a table. The vertical alignment helps see a typo or missing symbol quickly.

ImageLounge/src/DkCore/DkImageProc.h
scrubbbbs: I was only after the structured binding so sure a struct would be fine here.

ImageLounge/src/DkCore/DkImageProc.h
scrubbbbs: I am not concerned about taking resources from other processes for a short-lived operation. By short-lived I mean worst case 5-10 seconds. We don't have to worry about taking foreground resources because those are waiting for the adjustment to complete anyways.

I am concerned about a 64-core processor and each thread needs 1000MB of temporary memory... but we don't have anything like that yet. The closest is the grayscale kernel which needs one scanline of temporary memory which isn't much.

leejuyuu:
> > Compiler warnings
> 
> Thanks, I did not see that anywhere else, what gcc version is this?

15.2.1 20260209

ImageLounge/src/DkCore/DkNativeImage.h
scrubbbbs: There are no private members; DkConstNativeImage subclass needs access.  But I could scope things a little tighter

ImageLounge/src/DkCore/DkNativeImage.h
leejuyuu: After looking deeper into the code, I suggest something like `DkImageCVBridge`. Due to the necessity to convert to a format that all the related type supports, the allowed formats will be stricter if we add more image types other than cv::Mat. Therefore, I would expect this class to be bound to only cv::Mat and QImage.

`qtImageFormatToNative` in DkImageProc.h also contains potentially confusing "native".

leejuyuu:
I have some suggestions on the DkNativeImage interface:

- Accepting a mutable reference makes the impression that the source image is modified in-place. However, due to the potential conversion or detachment, this might not be true, hence the comment 'always call img().' I suggest changing the signature to accept either an rvalue reference that can be moved in or a const reference. This makes it clear from the signature that, when constructing DkNativeImage for mutation, we always need to retrieve the image explicitly (via `img()`). This also reduces some seemingly unnatural `QImage tmp = img` copies.
- Some of the image processing functions originally accept `QImage&` to modify the image in-place and return a `bool` to indicate success or not. To make them play well with the move of images and make the signature clearer, instead of modifying in-place, return a std::optional<QImage>. This also makes it clear that these transforms images.
- DkNativeImage originally only returns mutable `cv::Mat&`. This makes it less general since we can definitely have a const view on it. Introducing a `constMat()` similar to `constBits()` could make the intent clear and utilize the constness of DkNativeImage to return different types of references. DkConstNativeImage now only serves as a thin wrapper that removes the mutable view getter. It could also be removed.

<details>
<summary>Draft diff</summary>

```diff
diff --git a/ImageLounge/src/DkCore/DkImageStorage.cpp b/ImageLounge/src/DkCore/DkImageStorage.cpp
index f1d17b7e..6c933343 100755
--- a/ImageLounge/src/DkCore/DkImageStorage.cpp
+++ b/ImageLounge/src/DkCore/DkImageStorage.cpp
@@ -190,7 +190,7 @@ QImage DkImage::resizeImage(const QImage &src,
 
         auto input = DkConstNativeImage::fromImage(inImg);
         auto output = input.allocateLike(nSize);
-        cv::resize(input.mat(), output.mat(), cv::Size(nSize.width(), nSize.height()), 0, 0, ipl);
+        cv::resize(input.constMat(), output.mat(), cv::Size(nSize.width(), nSize.height()), 0, 0, ipl);
         QImage outImg = output.img();
 #else
         auto ipl = interpolation == ipl_nearest ? Qt::FastTransformation : Qt::SmoothTransformation;
@@ -525,7 +525,7 @@ protected:
         using DstType = typename DstFmt::ChannelType;
 
         auto &self = *(std::any_cast<DkGrayScaleKernel *>(arg));
-        const auto &src = self.mSrc.mat();
+        const auto &src = self.mSrc.constMat();
         auto &dst = self.mDst.mat();
         auto &srcToLinear = self.mSrcToLinear;
 
@@ -645,7 +645,7 @@ public:
         }
         dst.setColorSpace(dstColorSpace);
 
-        mDst = DkNativeImage::fromImage(dst);
+        mDst = DkNativeImage::fromImage(std::move(dst));
 
         return dispatch(table, kCaps, mSrc.img().format(), this, {0, mSrc.img().height()});
     }
@@ -862,8 +862,8 @@ public:
     DkNormalizeKernel() = delete;
     ~DkNormalizeKernel() override = default;
 
-    DkNormalizeKernel(QImage &img)
-        : mImg(DkNativeImage::fromImage(img))
+    DkNormalizeKernel(QImage &&img)
+        : mImg(DkNativeImage::fromImage(std::move(img)))
     {
     }
 
@@ -913,18 +913,17 @@ public:
 };
 #endif // WITH_OPENCV
 
-bool DkImage::normImage(QImage &src)
+std::optional<QImage> DkImage::normImage(QImage &&src)
 {
 #if WITH_OPENCV
-    DkNormalizeKernel kernel{src};
+    DkNormalizeKernel kernel{std::move(src)};
     if (kernel.run()) {
-        src = kernel.result();
-        return true;
+        return kernel.result();
     }
 #else
     Q_UNUSED(src)
 #endif
-    return false;
+    return std::nullopt;
 }
 
 #if WITH_OPENCV
@@ -938,8 +937,8 @@ public:
     DkAutoAdjustKernel() = delete;
     ~DkAutoAdjustKernel() override = default;
 
-    DkAutoAdjustKernel(QImage &img)
-        : mImg(DkNativeImage::fromImage(img))
+    DkAutoAdjustKernel(QImage &&img)
+        : mImg(DkNativeImage::fromImage(std::move(img)))
     {
     }
 
@@ -1003,18 +1002,17 @@ public:
 };
 #endif // WITH_OPENCV
 
-bool DkImage::autoAdjustImage(QImage &img)
+std::optional<QImage> DkImage::autoAdjustImage(QImage &&img)
 {
 #if WITH_OPENCV
-    DkAutoAdjustKernel kernel{img};
+    DkAutoAdjustKernel kernel{std::move(img)};
     if (kernel.run()) {
-        img = kernel.result();
-        return true;
+        return kernel.result();
     }
 #else
     Q_UNUSED(img)
 #endif
-    return false;
+    return std::nullopt;
 }
 
 QPixmap DkImage::makeSquare(const QPixmap &pm)
@@ -1157,8 +1155,8 @@ class DkHsvKernel : public DkKernelBase
 public:
     DkHsvKernel() = delete;
     Q_DISABLE_COPY(DkHsvKernel)
-    DkHsvKernel(QImage &img, float hue, float saturation, float brightness)
-        : mImg(DkNativeImage::fromImage(img, DkNativeImage::map_anyrgb))
+    DkHsvKernel(QImage &&img, float hue, float saturation, float brightness)
+        : mImg(DkNativeImage::fromImage(std::move(img), DkNativeImage::map_anyrgb))
         , mHue(hue)
         , mSaturation(saturation)
         , mBrightness(brightness)
@@ -1286,22 +1284,21 @@ public:
 };
 #endif // WITH_OPENCV
 
-bool DkImage::hueSaturation(QImage &img, float hue, float sat, float brightness)
+std::optional<QImage> DkImage::hueSaturation(QImage &&img, float hue, float sat, float brightness)
 {
     if (hue == 0 && sat == 0 && brightness == 0) {
-        return false;
+        return std::nullopt;
     }
 
 #ifdef WITH_OPENCV
-    DkHsvKernel kernel{img, hue, sat, brightness};
+    DkHsvKernel kernel{std::move(img), hue, sat, brightness};
     if (kernel.run()) {
-        img = kernel.result();
-        return true;
+        return kernel.result();
     }
 #else
     Q_UNUSED(img)
 #endif
-    return false;
+    return std::nullopt;
 }
 
 #if WITH_OPENCV
@@ -1397,8 +1394,8 @@ class DkLutKernel : DkKernelBase
 public:
     Q_DISABLE_COPY(DkLutKernel)
     DkLutKernel() = delete;
-    DkLutKernel(QImage &img, cv::Mat &lut)
-        : mImg(DkNativeImage::fromImage(img, DkNativeImage::map_anyrgb))
+    DkLutKernel(QImage &&img, cv::Mat &lut)
+        : mImg(DkNativeImage::fromImage(std::move(img), DkNativeImage::map_anyrgb))
         , mLut{lut}
     {
     }
@@ -1457,10 +1454,10 @@ public:
 };
 #endif // WITH_OPENCV
 
-bool DkImage::exposure(QImage &img, double exposure, double offset, double gamma)
+std::optional<QImage> DkImage::exposure(QImage &&img, double exposure, double offset, double gamma)
 {
     if (exposure == 0.0 && offset == 0.0 && gamma == 1.0) {
-        return false;
+        return std::nullopt;
     }
 
 #ifdef WITH_OPENCV
@@ -1473,16 +1470,15 @@ bool DkImage::exposure(QImage &img, double exposure, double offset, double gamma
         lut = combineLuts(lut, exposureLut(exposure));
     }
 
-    DkLutKernel lutKernel{img, lut};
+    DkLutKernel lutKernel{std::move(img), lut};
     if (lutKernel.run()) {
-        img = lutKernel.result();
-        return true;
+        return lutKernel.result();
     }
 #else
     Q_UNUSED(img)
 #endif // WITH_OPENCV
 
-    return false;
+    return std::nullopt;
 }
 
 // blend two colors with the standard source-over operator
@@ -2005,7 +2001,7 @@ QIcon DkImage::loadIcon(const QString &filePath, const QColor &color)
 
 cv::Mat DkImage::qImage2Mat(const QImage &img)
 {
-    return DkConstNativeImage::fromImage(img, DkNativeImage::map_bgr).mat().clone();
+    return DkConstNativeImage::fromImage(img, DkNativeImage::map_bgr).constMat().clone();
 }
 
 QImage DkImage::mat2QImage(cv::Mat mat, const QImage &srcImg)
@@ -2095,50 +2091,48 @@ QImage DkImage::tinyPlanet(const QImage &img, double scaleLog, double angle, con
 
     tmp = tmp.scaled(size, Qt::IgnoreAspectRatio, Qt::SmoothTransformation);
 
-    auto view = DkNativeImage::fromImage(tmp);
+    auto view = DkNativeImage::fromImage(std::move(tmp));
     logPolar(view.mat(), view.mat(), cv::Point2d(view.mat().cols * 0.5, view.mat().rows * 0.5), scaleLog, angle);
     return view.img();
 }
 
 #endif
 
-bool DkImage::gaussianBlur(QImage &img, float sigma)
+std::optional<QImage> DkImage::gaussianBlur(QImage &&img, float sigma)
 {
     if (sigma <= 0.0f) {
-        return false; // kernel would be 1x1 => no-op
+        return std::nullopt; // kernel would be 1x1 => no-op
     }
 
 #ifdef WITH_OPENCV
-    auto native = DkNativeImage::fromImage(img);
+    auto native = DkNativeImage::fromImage(std::move(img));
     cv::GaussianBlur(native.mat(), native.mat(), cv::Size{}, sigma);
-    img = native.img();
-    return true;
+    return native.img();
 #else
     Q_UNUSED(img);
     Q_UNUSED(sigma);
 #endif
-    return false;
+    return std::nullopt;
 }
 
-bool DkImage::unsharpMask(QImage &img, float sigma, float weight)
+std::optional<QImage> DkImage::unsharpMask(QImage &&img, float sigma, float weight)
 {
     if (sigma <= 0.0f) {
-        return false; // kernel would be 1x1 => no-op
+        return std::nullopt; // kernel would be 1x1 => no-op
     }
 
 #ifdef WITH_OPENCV
-    auto native = DkNativeImage::fromImage(img);
+    auto native = DkNativeImage::fromImage(std::move(img));
     cv::Mat blurred;
     cv::GaussianBlur(native.mat(), blurred, cv::Size{}, sigma);
     cv::addWeighted(native.mat(), weight, blurred, 1.0f - weight, 0.0, native.mat());
-    img = native.img();
-    return true;
+    return native.img();
 #else
     Q_UNUSED(img);
     Q_UNUSED(sigma);
     Q_UNUSED(weight);
 #endif
-    return false;
+    return std::nullopt;
 }
 
 QImage DkImage::createThumb(const QImage &image, int maxSize)
@@ -2419,7 +2413,7 @@ DkImageStorage::ScaledImage DkImageStorage::scaleImage(const QImage &src,
         try {
             const auto input = DkConstNativeImage::fromImage(resizedImg);
             auto output = input.allocateLike(s);
-            cv::resize(input.mat(), output.mat(), cv::Size(s.width(), s.height()), 0, 0, CV_INTER_AREA);
+            cv::resize(input.constMat(), output.mat(), cv::Size(s.width(), s.height()), 0, 0, CV_INTER_AREA);
             scaled = output.img();
         } catch (...) {
             qWarning() << "[ImageStorage]: OpenCV exception while resizing";
diff --git a/ImageLounge/src/DkCore/DkImageStorage.h b/ImageLounge/src/DkCore/DkImageStorage.h
index d7036833..9f2c21a4 100644
--- a/ImageLounge/src/DkCore/DkImageStorage.h
+++ b/ImageLounge/src/DkCore/DkImageStorage.h
@@ -95,10 +95,10 @@ public:
                               double factor = 1.0,
                               int interpolation = ipl_cubic,
                               bool correctGamma = true);
-    static bool normImage(QImage &img);
-    static bool autoAdjustImage(QImage &img);
-    static bool gaussianBlur(QImage &img, float sigma = 20.0f);
-    static bool unsharpMask(QImage &img, float sigma = 20.0f, float weight = 1.5f);
+    static std::optional<QImage> normImage(QImage &&img);
+    static std::optional<QImage> autoAdjustImage(QImage &&img);
+    static std::optional<QImage> gaussianBlur(QImage &&img, float sigma = 20.0f);
+    static std::optional<QImage> unsharpMask(QImage &&img, float sigma = 20.0f, float weight = 1.5f);
     static bool alphaChannelUsed(const QImage &img);
     static QImage thresholdImage(const QImage &img, double thr, bool color = false);
     static QImage flipImage(const QImage &image, Qt::Orientations flags);
@@ -126,8 +126,8 @@ public:
     static QPixmap makeSquare(const QPixmap &pm);
     static QPixmap merge(const QVector<QImage> &imgs);
     static QImage cropToImage(const QImage &src, const DkRotatingRect &rect, const QColor &fillColor = QColor());
-    static bool hueSaturation(QImage &img, float hue, float sat, float brightness);
-    static bool exposure(QImage &img, double exposure, double offset, double gamma);
+    static std::optional<QImage> hueSaturation(QImage &&img, float hue, float sat, float brightness);
+    static std::optional<QImage> exposure(QImage &&img, double exposure, double offset, double gamma);
     static QImage bgColor(const QImage &src, const QColor &col);
     static QByteArray extractImageFromDataStream(const QByteArray &ba,
                                                  const QByteArray &beginSignature = "‰PNG",
diff --git a/ImageLounge/src/DkCore/DkManipulatorsIpl.cpp b/ImageLounge/src/DkCore/DkManipulatorsIpl.cpp
index 34571adb..d841e88c 100644
--- a/ImageLounge/src/DkCore/DkManipulatorsIpl.cpp
+++ b/ImageLounge/src/DkCore/DkManipulatorsIpl.cpp
@@ -61,9 +61,10 @@ DkAutoAdjustManipulator::DkAutoAdjustManipulator(QAction *action)
 
 QImage DkAutoAdjustManipulator::apply(const QImage &img) const
 {
-    QImage imgR = img;
-    if (DkImage::autoAdjustImage(imgR))
-        return imgR;
+    std::optional<QImage> res = DkImage::autoAdjustImage(QImage(img));
+    if (res) {
+        return res.value();
+    }
 
     return QImage();
 }
@@ -81,9 +82,9 @@ DkNormalizeManipulator::DkNormalizeManipulator(QAction *action)
 
 QImage DkNormalizeManipulator::apply(const QImage &img) const
 {
-    QImage imgR = img;
-    if (DkImage::normImage(imgR)) {
-        return imgR;
+    std::optional<QImage> res = DkImage::normImage(QImage(img));
+    if (res) {
+        return res.value();
     }
 
     return QImage();
@@ -226,9 +227,9 @@ DkBlurManipulator::DkBlurManipulator(QAction *action)
 
 QImage DkBlurManipulator::apply(const QImage &img) const
 {
-    QImage tmp = img;
-    if (DkImage::gaussianBlur(tmp, (float)sigma())) {
-        return tmp;
+    std::optional<QImage> res = DkImage::gaussianBlur(QImage(img), (float)sigma());
+    if (res) {
+        return res.value();
     }
     return {};
 }
@@ -261,9 +262,9 @@ DkUnsharpMaskManipulator::DkUnsharpMaskManipulator(QAction *action)
 
 QImage DkUnsharpMaskManipulator::apply(const QImage &img) const
 {
-    QImage tmp = img;
-    if (DkImage::unsharpMask(tmp, (float)sigma(), 1.0f + amount() / 100.0f)) {
-        return tmp;
+    std::optional<QImage> res = DkImage::unsharpMask(QImage(img), (float)sigma(), 1.0f + amount() / 100.0f);
+    if (res) {
+        return res.value();
     }
     return {};
 }
@@ -435,9 +436,9 @@ DkHueManipulator::DkHueManipulator(QAction *action)
 
 QImage DkHueManipulator::apply(const QImage &img) const
 {
-    QImage tmp = img;
-    if (DkImage::hueSaturation(tmp, hue(), saturation(), value())) {
-        return tmp;
+    std::optional<QImage> res = DkImage::hueSaturation(QImage(img), hue(), saturation(), value());
+    if (res) {
+        return res.value();
     }
     return img;
 }
@@ -496,9 +497,9 @@ DkExposureManipulator::DkExposureManipulator(QAction *action)
 
 QImage DkExposureManipulator::apply(const QImage &img) const
 {
-    QImage tmp = img;
-    if (DkImage::exposure(tmp, exposure(), offset(), gamma())) {
-        return tmp;
+    std::optional<QImage> res = DkImage::exposure(QImage(img), exposure(), offset(), gamma());
+    if (res) {
+        return res.value();
     }
     return img;
 }
diff --git a/ImageLounge/src/DkCore/DkNativeImage.cpp b/ImageLounge/src/DkCore/DkNativeImage.cpp
index 4dbe5871..bc0883e2 100644
--- a/ImageLounge/src/DkCore/DkNativeImage.cpp
+++ b/ImageLounge/src/DkCore/DkNativeImage.cpp
@@ -69,8 +69,9 @@ int DkNativeImage::compatibleCvFormat(QImage::Format qtFormat, int options)
     return -1;
 }
 
-DkNativeImage DkNativeImage::fromImage(QImage &img, int options)
+DkNativeImage DkNativeImage::fromImageInner(QImage &&imgIn, int options, bool readOnly)
 {
+    QImage img(std::move(imgIn));
     Q_ASSERT(options & (map_bgr | map_rgb)); // choose one; if both, prefer bgr
 
     if (img.isNull()) {
@@ -148,14 +149,16 @@ DkNativeImage DkNativeImage::fromImage(QImage &img, int options)
         }
     }
 
+    // If we convert, it is safe to mutate.
+    readOnly = readOnly && converted.isNull();
+
     try {
         QImage &useImg = converted.isNull() ? img : converted; // careful not to refcount img: forces a deep copy
-        uchar *bits = options & map_readonly
-            ? const_cast<uchar *>(useImg.constBits()) // never deep copy, unsafe for writable views
-            : static_cast<uchar *>(useImg.bits()); // deep copy if img is a shallow copy
+        uchar *bits = readOnly ? const_cast<uchar *>(useImg.constBits()) // never deep copy, unsafe for writable views
+                               : static_cast<uchar *>(useImg.bits()); // deep copy if img is a shallow copy
         auto stride = static_cast<size_t>(useImg.bytesPerLine());
         cv::Mat mat{useImg.height(), useImg.width(), cvFormat, bits, stride};
-        return {useImg, mat};
+        return {std::move(useImg), std::move(mat), readOnly};
     } catch (...) {
         qWarning() << "[NativeImage] could not allocate cv::Mat";
     }
@@ -163,6 +166,16 @@ DkNativeImage DkNativeImage::fromImage(QImage &img, int options)
     return {};
 }
 
+DkNativeImage DkNativeImage::fromImage(QImage &&imgIn, int options)
+{
+    return fromImageInner(std::move(imgIn), options, false);
+}
+
+DkNativeImage DkNativeImage::fromConstImage(const QImage &imgIn, int options)
+{
+    return fromImageInner(QImage(imgIn), options, true);
+}
+
 DkNativeImage DkNativeImage::fromMat(cv::Mat &mat, const QImage &srcImg, int options)
 {
     int cvFormat = -1;
@@ -222,13 +235,12 @@ DkNativeImage DkNativeImage::fromMat(cv::Mat &mat, const QImage &srcImg, int opt
     QImage qImg(mat.data, mat.cols, mat.rows, mat.step, qtFormat);
     qImg.setColorSpace(srcImg.colorSpace());
 
-    return {qImg, mat};
+    return {std::move(qImg), cv::Mat(mat), false};
 }
 
 DkConstNativeImage DkConstNativeImage::fromImage(const QImage &img, int options)
 {
-    auto &nonConst = const_cast<QImage &>(img);
-    auto view = DkNativeImage::fromImage(nonConst, options | map_readonly);
+    auto view = DkNativeImage::fromConstImage(img, options);
     return DkConstNativeImage{std::move(view)};
 }
 
@@ -236,7 +248,7 @@ DkNativeImage DkNativeImage::allocateLike(const QSize &size) const
 {
     QImage tmp(size.isEmpty() ? mImg.size() : size, mImg.format());
     tmp.setColorSpace(mImg.colorSpace());
-    return fromImage(tmp, map_anyrgb);
+    return fromImage(std::move(tmp), map_anyrgb);
 }
 
 #endif // WITH_OPENCV
diff --git a/ImageLounge/src/DkCore/DkNativeImage.h b/ImageLounge/src/DkCore/DkNativeImage.h
index 793df8ee..dfef5641 100644
--- a/ImageLounge/src/DkCore/DkNativeImage.h
+++ b/ImageLounge/src/DkCore/DkNativeImage.h
@@ -11,59 +11,135 @@ namespace nmc
 
 // container for image in a "native" format supported by nomacs, which
 // for the moment means it can map directly to/from cv::Mat
+/**
+ * @brief A wrapper for sharing image buffers between Qt and OpenCV.
+ *
+ * DkNativeImage represents a shared image buffer that is accessible as a `const QImage&`
+ * and exposes a `cv::Mat&` view that can be processed by OpenCV.
+ * During construction, the image may be converted to a format that is supported by
+ * both QImage and cv::Mat.
+ *
+ * The exposed references remain valid as long as the lifetime of
+ * the DkNativeImage instance.
+ */
 class DkNativeImage
 {
     QImage mImg{};
+    cv::Mat mMat{};
+    bool mReadOnly = false;
 
     static int compatibleCvFormat(QImage::Format qtFormat, int options = map_anyrgb);
 
-protected:
-    cv::Mat mMat{};
-
-    DkNativeImage(const QImage &img, const cv::Mat &mat)
-        : mImg(img)
-        , mMat(mat)
+    DkNativeImage(QImage &&img, cv::Mat &&mat, bool readOnly)
+        : mImg(std::move(img))
+        , mMat(std::move(mat))
+        , mReadOnly{readOnly}
     {
     }
 
+    static DkNativeImage fromImageInner(QImage &&img, int options, bool readOnly);
+
 public:
     DkNativeImage() = default;
 
     // note: if constructed from cv::Mat, must use .copy() to persist a copy
-    const QImage &img() const // return const; any write to mImg will break link with mat
+    [[nodiscard]] const QImage &img() const // return const; any write to mImg will break link with mat
     {
         return mImg;
     }
 
-    // note: if constructed from QImage, must use .clone() to persist a copy
+    /**
+     * @brief Provides a mutable view to the underlying image data as OpenCV matrix.
+     *
+     * The cv::Mat& is valid as long as the DkNativeImage instance.
+     * Use cv::Mat::clone() to persist a copy.
+     *
+     * @warning Accessing this on a read-only image will trigger a fatal error.
+     * @return A reference to the cv::Mat for processing.
+     */
     cv::Mat &mat()
     {
+        if (mReadOnly) {
+            qFatal("attempt to access mutable mat of read only DkNativeImage");
+        }
+
         return mMat;
     }
 
+    /**
+     * @brief Provides a immutable view to the underlying image data as OpenCV matrix.
+     *
+     * The const cv::Mat& is valid as long as the DkNativeImage instance.
+     * Use cv::Mat::clone() to persist a copy.
+     *
+     * @return A reference to the cv::Mat for read.
+     */
+    [[nodiscard]] const cv::Mat &constMat() const
+    {
+        return mMat;
+    }
+
+    /**
+     * @brief Returns whether this instance is read-only.
+     *
+     * When the return value is true, calling `mat()` will trigger a fatal o
+     * error.
+     *
+     * @return Whether this instance is read-only.
+     */
+    [[nodiscard]] bool readOnly() const
+    {
+        return mReadOnly;
+    }
+
     // options for converting to native formats
     enum {
         map_bgr = 0x1, // cv::Mat is BGR(A) (or grayscale one channel) -- cv::cvtColor() wants this usually
         map_rgb = 0x2, // cv::Mat is RGB(A) (or grayscale one channel)
-        map_readonly = 0x4, // returned mat/image should not be modified
         map_anyrgb = map_bgr | map_rgb, // either order acceptable
     };
 
     /**
-     * @brief construct from QImage, converting if necessary
-     * @param img input reference to avoid taking a deep copy (see: QImage::bits())
+     * @brief construct mutable DkNativeImage from QImage, converting if necessary
+     *
+     * This factory method is used to construct a DkNativeImage and always returns
+     * an instance that is mutable via the reference returned from the `mat()` method.
+     *
+     *
+     * This factory method owns the `img` value,
+     * however, the image may be converted if the format is not supported by
+     * both QImage and cv::Mat, or the data could be copied if image refcount > 1.
+     * Therefore, always use the `DkNativeImage::img()` member to get the image
+     * after any mutation.
+     *
+     * The internal image is guarenteed to be detached from any other QImage that share
+     * the same buffer. Therefore any mutation will not affect other QImage.
+     * (Unless the QImage data is not managed by Qt, please avoid passing those in.)
+     *
+     * @param img input image It will be moved into the container.
      * @param options map_* options
+     * @return A mutable DkNativeImage.
+     */
+    static DkNativeImage fromImage(QImage &&img, int options = map_anyrgb);
+
+    /**
+     * @brief construct DkNativeImage from QImage, converting if necessary
      *
-     * @note  If the image refcount > 1, a deep copy of pixels is performed,
-     *        and the image is unmodified. So it is not possible to mutate other copies.
+     * This factory method is used to construct a DkNativeImage from a `const QImage&`.
+     * Conditioning on whether the image is converted, the resulting DkNativeImage may or may not
+     * allow mutation via `mat()`. The caller must check with `readOnly()` before
+     * attempting to call `mat()`.
      *
-     * @note If a direct mapping is not supported, there is a conversion, and
-     *       the passed image is unmodified; always use the "img" member to
-     *       get the QImage back
+     * If the conversion is not required, this method will not copy the image data.
+     * Due to the detach mechanisms of QImage, any modification of the same implicitly shared
+     * image outside of this instance will not be reflected.
+     * (Unless the QImage data is not managed by Qt, please avoid passing those in.)
      *
-     * @return see map_* options
+     * @param img input image
+     * @param options map_* options
+     * @return A mutable DkNativeImage.
      */
-    static DkNativeImage fromImage(QImage &img, int options = map_anyrgb);
+    static DkNativeImage fromConstImage(const QImage &img, int options = map_anyrgb);
 
     /**
      * @brief construct from QImage, converting if necessary
@@ -101,10 +177,7 @@ private:
     }
 
 public:
-    const cv::Mat &mat() const
-    {
-        return mMat;
-    }
+    cv::Mat &mat() = delete;
 
     /**
      * @brief immutable version
diff --git a/ImageLounge/tests/DkNativeImage_test.cpp b/ImageLounge/tests/DkNativeImage_test.cpp
index 582450e9..676309ff 100644
--- a/ImageLounge/tests/DkNativeImage_test.cpp
+++ b/ImageLounge/tests/DkNativeImage_test.cpp
@@ -129,7 +129,7 @@ TEST(ImageToMat, Default)
 
     for (int flags : options) {
         SCOPED_TRACE("flags: " + QString::number(flags).toStdString());
-        auto native = DkNativeImage::fromImage(nullImg, flags);
+        auto native = DkNativeImage::fromImage(std::move(nullImg), flags);
         EXPECT_TRUE(native.img().isNull());
         EXPECT_TRUE(native.mat().empty());
     }
@@ -137,7 +137,7 @@ TEST(ImageToMat, Default)
         SCOPED_TRACE("flags: " + QString::number(flags).toStdString());
         auto native = DkConstNativeImage::fromImage(nullImg, flags);
         EXPECT_TRUE(native.img().isNull());
-        EXPECT_TRUE(native.mat().empty());
+        EXPECT_TRUE(native.constMat().empty());
     }
 
     {
@@ -185,7 +185,7 @@ TEST(ImageToMat, Mapped)
 
         // normal mode, no copying
         {
-            auto native = DkNativeImage::fromImage(img);
+            auto native = DkNativeImage::fromImage(std::move(img));
             EXPECT_EQ(native.img().constBits(), native.mat().ptr<uchar>(0)); // correct linkage
             EXPECT_EQ(native.img().constBits(), origBits); // no copy should occur, img.refcount==1
 
@@ -203,9 +203,10 @@ TEST(ImageToMat, Mapped)
 
         img = {}; // now outerCopy is the only reference, should be no deep copy once again
         {
-            auto native = DkNativeImage::fromImage(outerCopy);
+            const auto outerCopyBits = outerCopy.constBits();
+            auto native = DkNativeImage::fromImage(std::move(outerCopy));
             EXPECT_EQ(native.img().constBits(), native.mat().ptr<uchar>(0)); // copy construction; no issues as well
-            EXPECT_EQ(native.img().constBits(), outerCopy.constBits());
+            EXPECT_EQ(native.img().constBits(), outerCopyBits);
             EXPECT_EQ(native.img().constBits(), origBits);
         }
 
@@ -218,25 +219,23 @@ TEST(ImageToMat, Mapped)
         QImage cowImg = img; // refcount of img data is now > 1, forces deep copy
         EXPECT_EQ(cowImg.constBits(), img.constBits());
         {
-            auto native = DkNativeImage::fromImage(cowImg);
-            EXPECT_EQ(native.img().constBits(), native.mat().ptr<uchar>(0));
-            EXPECT_EQ(native.img().constBits(), cowImg.constBits()); // no copy/correct linkage
+            auto native = DkNativeImage::fromImage(std::move(cowImg));
+            EXPECT_EQ(native.img().constBits(), native.mat().ptr<uchar>(0)); // correct linkage
             EXPECT_NE(native.img().constBits(), origBits); // COW enforced, deep copy of img
 
             auto copy = native;
             EXPECT_EQ(copy.img().constBits(), copy.mat().ptr<uchar>(0)); // copy construction; no issues as well
-            EXPECT_EQ(copy.img().constBits(), cowImg.constBits());
         }
 
         // read-only mode, no copying
         {
             auto native = DkConstNativeImage::fromImage(img);
-            EXPECT_EQ(native.img().constBits(), native.mat().ptr<uchar>(0));
+            EXPECT_EQ(native.img().constBits(), native.constMat().ptr<uchar>(0));
             EXPECT_EQ(native.img().constBits(), img.constBits());
             EXPECT_EQ(native.img().constBits(), origBits);
 
             auto copy = native;
-            EXPECT_EQ(copy.img().constBits(), copy.mat().ptr<uchar>(0)); // copy construction; no issues as well
+            EXPECT_EQ(copy.img().constBits(), copy.constMat().ptr<uchar>(0)); // copy construction; no issues as well
             EXPECT_EQ(copy.img().constBits(), img.constBits());
 
             // native.img().fill(0); // does not compile
@@ -250,7 +249,7 @@ TEST(ImageToMat, Mapped)
         EXPECT_EQ(cowImg.constBits(), img.constBits());
         {
             auto native = DkConstNativeImage::fromImage(cowImg);
-            EXPECT_EQ(native.img().constBits(), native.mat().ptr<uchar>(0));
+            EXPECT_EQ(native.img().constBits(), native.constMat().ptr<uchar>(0));
             EXPECT_EQ(native.img().constBits(), cowImg.constBits());
             EXPECT_EQ(native.img().constBits(), origBits); // COW relaxed: not allowed to mutate img/mat
         }
@@ -279,7 +278,7 @@ TEST(ImageToMat, ChannelOrder)
             SCOPED_TRACE("native format: " + qEnumToString(native.img().format()));
             SCOPED_TRACE("native option: " + QString::number(options[i]).toStdString());
 
-            EXPECT_TRUE(checkPixel(native.mat())); // check channel[0] is set to 0xFF
+            EXPECT_TRUE(checkPixel(native.constMat())); // check channel[0] is set to 0xFF
         }
     }
 }
@@ -316,7 +315,7 @@ TEST(ImageToMat, Scaled)
             auto out = in.allocateLike(size);
             EXPECT_EQ(out.img().size(), size);
 
-            cv::resize(in.mat(), out.mat(), cv::Size(size.width(), size.height()), 0, 0, cv::INTER_NEAREST);
+            cv::resize(in.constMat(), out.mat(), cv::Size(size.width(), size.height()), 0, 0, cv::INTER_NEAREST);
             src = out.img();
         }
         EXPECT_TRUE(checkPixel(src));
```

</details>

leejuyuu:
Another question:

The `rotateImageCVMat()` constructs image buffers and sets several properties after some bug fix. Should `DkNativeImage::allocateLike()` follow the same?

https://github.com/nomacs/nomacs/blob/daeb573017937f666225cbe84d51b3a2e4b66349/ImageLounge/src/DkCore/DkImageStorage.cpp#L416-L424

leejuyuu:
About the DkBaseKernel and its derivations, I see that you put in a great deal of effort to make static dispatch work. I worry that this part of the code will become too complicated to debug and change. Did you compare with a simpler inheritance pattern with virtual methods?

ImageLounge/src/DkCore/DkNativeImage.h
scrubbbbs: > Therefore, I would expect this class to be bound to only cv::Mat and QImage.

Thanks for the update.

I was thinking it could also map/convert more types in the future. And this will probably work due to the broad support that is already there and ability to convert things when needed. So naming it something Cv specific is going to discourage extending it for future work.

> `qtImageFormatToNative` in DkImageProc.h also contains potentially confusing "native".

I think we need a name to describe pixel formats that can be handled without conversions, copies, etc. If not "native", then something. "native" is fairly commonplace in graphics APIs. What about "direct" or "raw",  these have similar meaning as something that is ready to be consumed by an API. We would then have `DkDirectImage` and `qImageFormatToDirect`

scrubbbbs:
Update: I am looking at your patch more closely. I didn't realize you were passing "&&img" to the image filters now, which changes things if we can make that work.

- I thought std::move() on QImage would be a footgun, but the move operator makes the image null...so that is probably fine.
- The problem with accidental deep copies `QImage(img)` on non-const img remains. I can live with it, but it is not ideal.
- I'm not sure if we can chain operations without copies yet, if we can this patch is fine. Do you want to submit a PR (to my repo) or shall I commit with co-author?
--- 

> I suggest changing the signature to accept either an rvalue reference that can be moved in or a const reference. This makes it clear from the signature that, when constructing DkNativeImage for mutation, we always need to retrieve the image explicitly (via `img()`).

So the feature is zero-copy mutable mapping on (non-const) QImage, and we have 3 different ways use an image:
1. passed as `const QImage&`
2. passed as `QImage&`
3. local variable/local copy etc

So given what we know,
1. Doesn't work, `QImage::bits() const` forces a deep copy
2. May work, if refcount is 1,  works on arguments or local variables
3. May work, if recount is 1,  generally only works on local variables as we rarely pass `&&img` 

If we can only pass by `QImage &&` that eliminates case 2 from being possible. So it won't be possible to chain operations without copying the pixel buffer - unless working on a local variable.

What this will do is cause a deep copy when it isn't necessary, because there is nothing to prevent this:

```c++
QImage img; // refcount==1
// QPainter/QImage things
auto native=DkNativeImage::fromImage(QImage(img)); // refcount==2, QImage::bits() deep copy required
// OpenCV things
img=native.img();
```

You could have `std::move()` in there instead but that invalidates `img` until reassignment. Doing anything with `img` before then is undefined or crashes nomacs.

It seems like we would be giving up chaining operations and potentially introducing more deep copies and bugs.

scrubbbbs:
> About the DkBaseKernel and its derivations, I see that you put in a great deal of effort to make static dispatch work. I worry that this part of the code will become too complicated to debug and change. Did you compare with a simpler inheritance pattern with virtual methods?

I started from the idea that a kernel should be a template function on a pixel format and went from there. The static dispatch is what followed from that. I don't see this changing much or getting any simpler (the dispatch part). Maybe we add more pixel formats (doubtful) or use new C++ features like concepts to clean up warning messages or compiler errors.

If inheritance was used, each kernel would need a bunch of nearly-identical overloads for each pixel format. And when we add two pixel formats which can be different (a source and destination), the number of overloads explodes. Having a lot of nearly identical code doesn't seem great for maintenance.

The only other sane alternative is to convert everything to and from one or two formats, and accept the huge performance degradation from added memory use and conversions.

scrubbbbs:
> The `rotateImageCVMat()` constructs image buffers and sets several properties after some bug fix. Should `DkNativeImage::allocateLike()` follow the same?

Yes. I think it will have no effect for the moment, but a good idea.

scrubbbbs:
> * DkNativeImage originally only returns mutable `cv::Mat&`. This makes it less general since we can definitely have a const view on it. Introducing a `constMat()` similar to `constBits()` could make the intent clear and utilize the constness of DkNativeImage to return different types of references. DkConstNativeImage now only serves as a thin wrapper that removes the mutable view getter. It could also be removed.

I'd want to encourage using `DkConstNativeImage` (or something similar) if you needed a `const cv::Mat&` because otherwise you usually will make a deep copy. `DkConstNativeImage` is the only thing that can safely use `QImage::constBits()` to almost always get a direct mapping (if no conversion is needed - and it usually isn't).

For single class solution, there would have to be `const DkNativeImage fromConstImage()` to suggest proper usage, as that const is easily removed by assignment, and then the source image can be mutated. And these effects can be subtle for example a manipulator will modify the previous history entry or something in a cache.

leejuyuu:
> * The problem with accidental deep copies `QImage(img)` on non-const img remains. I can live with it, but it is not ideal.

These are already present with the `QImage tmp = img` assignments. I suppose they are not completely avoidable since sometimes we cannot modify the image in-place.

> * I'm not sure if we can chain operations without copies yet, if we can this patch is fine. Do you want to submit a PR (to my repo) or shall I commit with co-author?

Due to the conversion or detachment, `DkNativeImage` has an interesting property that, as long as the `img()` reference is not copied to another `QImage`, ensures the `mImg` is the only `QImage` that references its underlying buffer. Therefore, chaining without a copy should be possible if we ensure the previous `DkNativeImage` goes out of scope right after retrieving the `img()`. I think of this as I am typing, maybe we should replace the `const QImage& img()` method with a function `QImage moveDkNativeImage(DkNativeImage&& src)` to absorb the `DkNativeImage` by moving, further ensuring this only happens once.

If it is okay for you, you can commit it directly with co-author. Thanks!

> You could have `std::move()` in there instead but that invalidates `img` until reassignment. Doing anything with `img` before then is undefined or crashes nomacs.

Maybe we should add the `bugprone-use-after-move` clang-tidy rule to catch this.
(Note: This check requires static analysis, so it is currently not possible to make the warning show in editors via clangd.)

leejuyuu:
> For single class solution, there would have to be const DkNativeImage fromConstImage() to suggest proper usage, as that const is easily removed by assignment, and then the source image can be mutated. And these effects can be subtle for example a manipulator will modify the previous history entry or something in a cache.

This could be mitigated by deleting the copy constructor and the copy assignment of DkNativeImage.

scrubbbbs:
> > For single class solution, there would have to be const DkNativeImage fromConstImage() to suggest proper usage, as that const is easily removed by assignment, and then the source image can be mutated. And these effects can be subtle for example a manipulator will modify the previous history entry or something in a cache.
> 
> This could be mitigated by deleting the copy constructor and the copy assignment of DkNativeImage.

I don't think this is needed now. `fromConstImage()` sets a read-only flag such that taking the const away will produce a runtime error when `mat()` is called.

scrubbbbs:
@leejuyuu  Unless you have anything else I think this is ready. I'll test a bit more and plan to merge this weekend.

leejuyuu:
Looks great, thanks! I'd personally want a more descriptive name other than "native", but this should not be a blocker.

PR #1525, opened 2026-02-26
One key to toggle between "show 100%" and "fit image to window"

eddiezato:
I really missed this feature. So I tried to add it myself.


scrubbbbs:
Thanks! Can you please describe more how this should work?
- Is this essentially going to do Ctrl+0 or Ctrl+1 based on the current zoom level?
- What happens when resizing the window or moving to the next image, does it reset the view according to preference?
- How does it interact with the current zoom-to-fit preference, or should there be no influence

If you want to see this incorporated (and I think it can be), please rebase onto [@leejuyuu/nomacs:zoom](https://github.com/leejuyuu/nomacs/tree/zoom) from #1516 , there is a major refactoring/bug-hunt going on there for the zoom and fit features.

Please check carefully how you have implemented the changes in DkViewPort(Base) etc that it doesn't cause a problem with the new version.

If you can do that then we will be able to review when #1516 is completed.

leejuyuu:
I suggest postponing the rebase until #1516 is merged into master. I generally prefer rebasing only to master to avoid having to resolve conflicts multiple times. This change seems to have little or no  dependency on that one, so it should be safe to be reviewed as is.

However, since this functionality seems identical to Ctrl + 0 and Ctrl + 1, I'm not sure if repeating it is necessary

leejuyuu:
@scrubbbbs This is a one-time operation to change the zoom of the current image and is not related to the next image.

eddiezato:
> Can you please describe more how this should work?

This sets the zoom level to 100% if it's different from 100%; otherwise, it fits the image to the window.

The main purpose is to combine Ctrl+0 and Ctrl+1 into a single key. Many image viewers have this feature.

scrubbbbs:
> However, since this functionality seems identical to Ctrl + 0 and Ctrl + 1, I'm not sure if repeating it is necessary

Maybe it should be a fast mode toggle between "always 100%" fit (which we don't have yet) and zoom-to-fit preference. You get the same functionality but the added benefit of retaining behavior between images. Without taking a trip to the zoom settings.

scrubbbbs:
@leejuyuu  is refactoring viewport code in #1556 which likely break this fix. We can revisit it when that has been merged.

PR #1526, opened 2026-02-26
Histogram panel cleanup

scrubbbbs:
Summary:
- All image formats supported
- Fast rendering (threaded histogram compute,  render cache)
- Nice colors (less harsh, good contrast for C,M,Y mix)
- Fixed stats text layout, values properly scaled/displayed
- Don't count transparent pixels  
- Logarithmic scale option
  
image

### DIFF
--- a/ImageLounge/src/DkGui/DkWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkWidgets.cpp
@@ -2070,7 +2070,7 @@ void DkHistogram::paintEvent(QPaintEvent *)
     int binBottom = height() - margin - textHeight;
 
     // draw Histogram
-    if (mIsPainted && mMaxValue > 0) {
+    if (mIsValid && mMaxValue > 0) {
         for (int x = 0; x < 256; x++) {
             // get bounded values
             int rLineHeight = qMax(qMin(qRound((float)mHist[0][x] * binHeight * mScaleFactor / mMaxValue), binHeight),
@@ -2157,7 +2157,7 @@ void DkHistogram::loadSettings()
 void DkHistogram::drawHistogram(const QImage &imgQt)
 {
     if (!isVisible() || imgQt.isNull()) {
-        setPainted(false);
+        setValid(false);
         return;
     }
 
@@ -2262,7 +2262,7 @@ void DkHistogram::drawHistogram(const QImage &imgQt)
         }
     }
 
-    setPainted(true);
+    setValid(true);
     qDebug() << "drawing the histogram took me: " << dt;
     update();
 }
@@ -2272,19 +2272,19 @@ void DkHistogram::drawHistogram(const QImage &imgQt)
  **/
 void DkHistogram::clearHistogram()
 {
-    setPainted(false);
+    setValid(false);
     update();
 }
 
-void DkHistogram::setPainted(bool isPainted)
+void DkHistogram::setValid(bool isValid)
 {
-    this->mIsPainted = isPainted;
+    this->mIsValid = isValid;
 }
 
 void DkHistogram::setMaxHistogramValue(int maxValue)
 {
     if (maxValue == 0)
-        setPainted(false);
+        setValid(false);
 
     this->mMaxValue = maxValue;
 }
--- a/ImageLounge/src/DkGui/DkWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkWidgets.cpp
@@ -2038,7 +2038,12 @@ DkHistogramWidget::DkHistogramWidget(QWidget *parent)
     auto *showStats = new QAction(tr("Show Statistics"), this);
     showStats->setCheckable(true);
     showStats->setChecked(mDisplayMode == DisplayMode::mode_extended);
-    connect(showStats, &QAction::triggered, this, &DkHistogramWidget::onToggleStatsTriggered);
+    connect(showStats, &QAction::triggered, this, [&](bool checked) {
+        mDisplayMode = checked ? DisplayMode::mode_extended : DisplayMode::mode_simple;
+        DkSettingsManager::param().display().histogramStyle = static_cast<int>(mDisplayMode);
+        mIsDirty = true;
+        update();
+    });
 
     mContextMenu = new QMenu;
     mContextMenu->addAction(showStats);
@@ -2080,14 +2085,6 @@ void DkHistogramWidget::contextMenuEvent(QContextMenuEvent *event)
 
     // do not pass it on
     // DkFadeWidget::contextMenuEvent(event);
-}
-
-void DkHistogramWidget::onToggleStatsTriggered(bool show)
-{
-    mDisplayMode = (show) ? DisplayMode::mode_extended : DisplayMode::mode_simple;
-    DkSettingsManager::param().display().histogramStyle = (int)mDisplayMode;
-    mIsDirty = true;
-    update();
 }
 
 void DkHistogramWidget::loadSettings()
--- a/ImageLounge/src/DkGui/DkWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkWidgets.cpp
@@ -2040,7 +2040,7 @@ DkHistogram::DkHistogram(QWidget *parent)
     showStats->setChecked(mDisplayMode == DisplayMode::histogram_mode_extended);
     connect(showStats, &QAction::triggered, this, &DkHistogram::onToggleStatsTriggered);
 
-    mContextMenu = new QMenu(tr("Histogram Settings"));
+    mContextMenu = new QMenu;
     mContextMenu->addAction(showStats);
 }
 
--- a/ImageLounge/src/DkGui/DkWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkWidgets.cpp
@@ -2154,7 +2154,7 @@ void DkHistogram::loadSettings()
  * Goes through the image and counts pixels values. They are used to create the image histogram.
  * @param currently displayed image
  **/
-void DkHistogram::drawHistogram(QImage imgQt)
+void DkHistogram::drawHistogram(const QImage &imgQt)
 {
     if (!isVisible() || imgQt.isNull()) {
         setPainted(false);
--- a/ImageLounge/src/DkGui/DkWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkWidgets.cpp
@@ -2004,7 +2004,12 @@ void DkCropWidget::crop(bool cropToMetadata)
         QPolygonF poly = mat.map(mRect.getPoly());
         DkRotatingRect rect;
         rect.setPoly(poly);
-        emit cropImageSignal(rect, cropToolbar->getColor(), cropToMetadata);
+        // Viewport may not refresh panels that are currently hidden due
+        // to crop widget being active. Send after crop widget is hidden and
+        // presumably the viewport is shown.
+        QTimer::singleShot(0, [rect, cropToMetadata, this] {
+            emit cropImageSignal(rect, cropToolbar->getColor(), cropToMetadata);
+        });
     }
 
     setVisible(false);
--- a/ImageLounge/src/DkGui/DkWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkWidgets.cpp
@@ -2037,7 +2037,7 @@ DkHistogramWidget::DkHistogramWidget(QWidget *parent)
     // create context menu
     auto *showStats = new QAction(tr("Show Statistics"), this);
     showStats->setCheckable(true);
-    showStats->setChecked(mDisplayMode == DisplayMode::histogram_mode_extended);
+    showStats->setChecked(mDisplayMode == DisplayMode::mode_extended);
     connect(showStats, &QAction::triggered, this, &DkHistogramWidget::onToggleStatsTriggered);
 
     mContextMenu = new QMenu;
@@ -2065,7 +2065,7 @@ void DkHistogramWidget::paintEvent(QPaintEvent *)
 
         if (mIsDirty) {
             mHistImage.fill(Qt::transparent);
-            mHistogram->render(mHistImage, mScaleFactor, mDisplayMode == DisplayMode::histogram_mode_extended);
+            mHistogram->render(mHistImage, mScaleFactor, mDisplayMode == DisplayMode::mode_extended);
             mIsDirty = false;
         }
 
@@ -2084,7 +2084,7 @@ void DkHistogramWidget::contextMenuEvent(QContextMenuEvent *event)
 
 void DkHistogramWidget::onToggleStatsTriggered(bool show)
 {
-    mDisplayMode = (show) ? DisplayMode::histogram_mode_extended : DisplayMode::histogram_mode_simple;
+    mDisplayMode = (show) ? DisplayMode::mode_extended : DisplayMode::mode_simple;
     DkSettingsManager::param().display().histogramStyle = (int)mDisplayMode;
     mIsDirty = true;
     update();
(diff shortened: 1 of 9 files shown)


leejuyuu:
image

Can we have more bottom padding? The text looks almost at the edge of the panel.

Also, there is no visual indicator for the current scale (log or linear).

ImageLounge/src/DkGui/DkWidgets.h
```diff
@@ -578,10 +578,10 @@ class DkHistogramWidget : public DkFadeWidget
...
 
 public:
-    enum class DisplayMode {
-        histogram_mode_simple = 0, /// shows just the histogram
-        histogram_mode_extended = 1, /// shows histogram and data
-        histogram_mode_end = 2,
+    enum DisplayMode {
+        mode_simple = 0, // shows just the histogram
+        mode_extended = 1, // shows histogram and data
+        mode_end = 2,
```
leejuyuu, line 584: Using `enum class` has its benefits. The application logic knows that this variable can only have three different values, and `enum class` helps convey this through the type (as long as we do not assign an arbitrary int literal). The complexity of translating the setting (forced to be int by Qt) to something the application knows (i.e., the enum class) should remain in `DkSetting`, instead of forcing the application class to adapt to the type of the setting.

ImageLounge/src/DkGui/DkWidgets.cpp
```diff
@@ -2037,7 +2037,12 @@ DkHistogramWidget::DkHistogramWidget(QWidget *parent)
     auto *showStats = new QAction(tr("Show Statistics"), this);
     showStats->setCheckable(true);
     showStats->setChecked(mDisplayMode == DisplayMode::mode_extended);
-    connect(showStats, &QAction::triggered, this, &DkHistogramWidget::onToggleStatsTriggered);
+    connect(showStats, &QAction::triggered, this, [&](bool checked) {
+        mDisplayMode = checked ? DisplayMode::mode_extended : DisplayMode::mode_simple;
+        DkSettingsManager::param().display().histogramStyle = static_cast<int>(mDisplayMode);
+        mIsDirty = true;
+        update();
+    });
```
leejuyuu: It is likely that the lamba only needs to capture `this` instead of `&`. I think it is better to default to avoiding `&` whenever possible.

ImageLounge/src/DkGui/DkWidgets.cpp
```diff
@@ -2003,7 +2003,12 @@ void DkCropWidget::crop(bool cropToMetadata)
         QPolygonF poly = mat.map(mRect.getPoly());
         DkRotatingRect rect;
         rect.setPoly(poly);
-        emit cropImageSignal(rect, cropToolbar->getColor(), cropToMetadata);
+        // Viewport may not refresh panels that are currently hidden due
+        // to crop widget being active. Send after crop widget is hidden and
+        // presumably the viewport is shown.
+        QTimer::singleShot(0, [rect, cropToMetadata, this] {
+            emit cropImageSignal(rect, cropToolbar->getColor(), cropToMetadata);
+        });
```
leejuyuu, line 2011: Could you clarify what the QTimer does? Is its purpose to make the signal emit after this function returns?

scrubbbbs:
> Can we have more bottom padding? The text looks almost at the edge of the panel.

Strange, there is padding on everything I tested on. I'll try it a different way. Qt rich text is a bit fidgety.

> Also, there is no visual indicator for the current scale (log or linear).

The full extent of leftmost/rightmost bars must be visible to evaluate clipping, so top right/left corners can't have an icon for this. Unless a margin is added at the top or side which breaks the minimalist aesthetic. Maybe light-colored horizontal lines behind or over the bars?

ImageLounge/src/DkGui/DkWidgets.cpp
scrubbbbs: After it returns, and all immediate side-effects of the current event return as well. It is an old Qt idiom which is to do something after reaching the event loop which is where timers are processed.

In this case the side-effect we have is the old `QWidget::isVisible()` issue; visibility of  a widget does not mean the component/panel is active or inactive but it gets used that way everywhere.

ImageLounge/src/DkGui/DkWidgets.h
scrubbbbs: My reasoning at the time was that we just cast away the enum class constantly so what is the point.

If we were to do as you suggest, DkSettings would have to define and validate nearly all types like this in the program? But no validation for range-bound int, float, or other complex types?

I would prefer to weaken the enum and have modules validate their own settings where possible rather than add coupling to DkSettings. Ideally at some point we could remove all this coupling and have some kind of registry for modules to register handlers for their settings.

leejuyuu:
> > Can we have more bottom padding? The text looks almost at the edge of the panel.
> 
> Strange, there is padding on everything I tested on. I'll try it a different way. Qt rich text is a bit fidgety.

In case it matters, I am using System + fusion theme combination.

> > Also, there is no visual indicator for the current scale (log or linear).
> 
> The full extent of leftmost/rightmost bars must be visible to evaluate clipping, so top right/left corners can't have an icon for this. Unless a margin is added at the top or side which breaks the minimalist aesthetic. Maybe light-colored horizontal lines behind or over the bars?

I am okay with the horizontal line. An alternative is short ticks on the vertical axis.

scrubbbbs:
> I am okay with the horizontal line. An alternative is short ticks on the vertical axis.

I think this looks OK. Tick marks on 5ths but that leaves no marks until the top part on log scale.

image
image
image
image

ImageLounge/src/DkGui/DkWidgets.h
leejuyuu: > If we were to do as you suggest, DkSettings would have to define and validate nearly all types like this in the program? But no validation for range-bound int, float, or other complex types?

Ideally, we should do as much validation as we can. I think of this as establishing boundaries around our code. Once a value crosses, we treat it as reliable. Using an enum class can make it clear that "this value is validated and safe to use." 

> I would prefer to weaken the enum and have modules validate their own settings where possible rather than add coupling to DkSettings. Ideally at some point we could remove all this coupling and have some kind of registry for modules to register handlers for their settings.

There would be tradeoffs. I personally like validating as early and as centrally as possible. When I am writing Go, I tend to define structs in each of the modules that require them, and the global `Config` is composed of these structs. Config does not need to know how its members validate themselves; it just needs to call specific methods/functions that do it. I am not sure whether this plays well with QSettings, though.

leejuyuu:
> > I am okay with the horizontal line. An alternative is short ticks on the vertical axis.
> 
> I think this looks OK. Tick marks on 5ths but that leaves no marks until the top part on log scale.

Looks great!

leejuyuu:
The padding change does take effect for me.

image

leejuyuu:
I looked into the code, `DkHistogramKernel` mixes UI logic into business logic, which could be problematic in the future. The histogram processing function should only return data required to draw the histogram and statistics, but it now generates a part of the UI as an image.

leejuyuu:
The padding issue is an overflow because the statistics block has a constant 25% height. If the font is too large or too tall, it would not fit vertically. Calculating the document height dynamically can resolve this.

However, it might be simpler to utilize QLayout, etc.

```diff
diff --git a/ImageLounge/src/DkCore/DkImageStorage.cpp b/ImageLounge/src/DkCore/DkImageStorage.cpp
index c067d421..855ad794 100755
--- a/ImageLounge/src/DkCore/DkImageStorage.cpp
+++ b/ImageLounge/src/DkCore/DkImageStorage.cpp
@@ -2464,18 +2464,74 @@ protected:
         const int imgWidth = img.width();
 
         const float margin = 4; // even number required
-        const int statsHeight = 0.25 * imgHeight / dpr;
-        const int histHeight = showStats ? imgHeight / dpr - statsHeight : imgHeight / dpr - margin / 2;
-
-        const float y0 = histHeight; // bottom of bars
-        const float yScale = logScale ? float(histHeight) / std::log(static_cast<float>(maxCount)) * zoom
-                                      : float(histHeight) / maxCount * zoom;
-
+        //
         const int numChannels = qMin(3, imgChannels);
         const int numBins = h.kNumBins;
         const int numTicks = 4;
-
+        //
         QPainter painter(&img);
+        double histHeight = imgHeight / dpr - margin / 2;
+        if (showStats) {
+            double blackPct = h.numBlack * 100.0 / h.numPixels;
+            double whitePct = h.numWhite * 100.0 / h.numPixels;
+            double goodPct = (h.numPixels - h.numBlack - h.numWhite) * 100.0 / h.numPixels;
+
+            // percent of bins with zero sample count (across channels) == bands/gaps in histogram
+            double levelPct = (numBins - h.numEmptyLevels()) * 100.0 / numBins;
+
+            static constexpr QStringView styleSheet =
+                uR"(table,dt,dr { border-collapse: collapse; margin: %1px }
+                    td { margin: %1px; padding: 0px; font-size: %2px; font-family:"%3"; color:"%4"; })";
+
+            static constexpr QStringView html =
+                uR"(<table width="100%">
+                <tr>
+                    <td>Min:</td> <td>%1</td>
+                    <td>Blk:</td> <td>%3%</td>
+                    <td>OK:</td>  <td>%5%</td>
+                </tr>
+                <tr>
+                    <td>Max:</td> <td>%2</td>
+                    <td>Wht:</td> <td>%4%</td>
+                    <td>Lvl:</td> <td>%6%</td>
+                </tr>
+                </table>)";
+
+            QString text = html.toString()
+                               .arg(h.globalMin())
+                               .arg(h.globalMax())
+                               .arg(blackPct, 0, 'f', 2, ' ')
+                               .arg(whitePct, 0, 'f', 2, ' ')
+                               .arg(goodPct, 0, 'f', 1, ' ')
+                               .arg(levelPct, 0, 'f', 1, ' ');
+
+            QTextDocument doc;
+            doc.setTextWidth(img.width() / dpr);
+            doc.setDocumentMargin(0);
+
+            // monospace font to reduce jitter
+            QString systemMono = QFontDatabase::systemFont(QFontDatabase::FixedFont).family();
+            QString fontColor = DkSettingsManager::param().display().hudFgdColor.name(QColor::HexArgb);
+
+            QString style = styleSheet.toString()
+                                .arg(static_cast<int>(margin / 2.0)) // <table> margin
+                                .arg(13) // font pixel size
+                                .arg(systemMono)
+                                .arg(fontColor);
+            doc.setDefaultStyleSheet(style);
+
+            doc.setHtml(text);
+
+            painter.save();
+            histHeight = imgHeight / dpr - doc.size().height();
+            painter.translate(margin, histHeight);
+            doc.drawContents(&painter);
+            painter.restore();
+        }
+
+        const float y0 = histHeight; // bottom of bars
+        const float yScale = logScale ? float(histHeight) / std::log(static_cast<float>(maxCount)) * zoom
+                                      : float(histHeight) / maxCount * zoom;
 
         QList<QLineF> tickMarks;
         tickMarks.reserve(numTicks);
@@ -2534,61 +2590,6 @@ protected:
             painter.drawLines(lines);
         }
 
-        if (showStats) {
-            double blackPct = h.numBlack * 100.0 / h.numPixels;
-            double whitePct = h.numWhite * 100.0 / h.numPixels;
-            double goodPct = (h.numPixels - h.numBlack - h.numWhite) * 100.0 / h.numPixels;
-
-            // percent of bins with zero sample count (across channels) == bands/gaps in histogram
-            double levelPct = (numBins - h.numEmptyLevels()) * 100.0 / numBins;
-
-            static constexpr QStringView styleSheet =
-                uR"(table,dt,dr { border-collapse: collapse; }
-                    td { margin:%1px; padding:0px; font-size: %2px; font-family:"%3"; color:"%4"; })";
-
-            static constexpr QStringView html =
-                uR"(<table width="100%">
-                <tr>
-                    <td>Min:</td> <td>%1</td>
-                    <td>Blk:</td> <td>%3%</td>
-                    <td>OK:</td>  <td>%5%</td>
-                </tr>
-                <tr>
-                    <td>Max:</td> <td>%2</td>
-                    <td>Wht:</td> <td>%4%</td>
-                    <td>Lvl:</td> <td>%6%</td>
-                </tr>
-                </table>)";
-
-            QString text = html.toString()
-                               .arg(h.globalMin())
-                               .arg(h.globalMax())
-                               .arg(blackPct, 0, 'f', 2, ' ')
-                               .arg(whitePct, 0, 'f', 2, ' ')
-                               .arg(goodPct, 0, 'f', 1, ' ')
-                               .arg(levelPct, 0, 'f', 1, ' ');
-
-            QTextDocument doc;
-            doc.setTextWidth(img.width() / dpr);
-            doc.setDocumentMargin(0);
-
-            // monospace font to reduce jitter
-            QString systemMono = QFontDatabase::systemFont(QFontDatabase::FixedFont).family();
-            QString fontColor = DkSettingsManager::param().display().hudFgdColor.name(QColor::HexArgb);
-
-            QString style = styleSheet.toString()
-                                .arg(static_cast<int>(margin / 2.0)) // <table> margin
-                                .arg(13) // font pixel size
-                                .arg(systemMono)
-                                .arg(fontColor);
-            doc.setDefaultStyleSheet(style);
-
-            doc.setHtml(text);
-
-            painter.translate(margin, y0 + margin / 2.0);
-            doc.drawContents(&painter);
-        }
-
         painter.end();
 
         return true;

```

scrubbbbs:
> Calculating the document height dynamically can resolve this.

Thanks for the hint. I think I have it now using this approach.

> However, it might be simpler to utilize QLayout, etc.

Could also work, but I don't think much simpler due to text overflow issues which are handled automatically on the html side.

scrubbbbs:
> I looked into the code, `DkHistogramKernel` mixes UI logic into business logic [...]

I don't understand this, can you elaborate?

leejuyuu:
> > I looked into the code, `DkHistogramKernel` mixes UI logic into business logic [...]
> 
> I don't understand this, can you elaborate?

Sorry, I seem to have missed the comments. `DkHistogramKernel` should know the data and logic (e.g., how to calculate a histogram from image pixels), but not the representation of the histogram on the UI (e.g. the HTML layout and the text's position). I assume this was done to satisfy the multi-threading kernel interface, which returns an image; however, I think it is unexpected to find UI layout logic inside an image processing kernel.

scrubbbbs:
> > > I looked into the code, `DkHistogramKernel` mixes UI logic into business logic [...]
> > 
> > 
> > I don't understand this, can you elaborate?
> 
> Sorry, I seem to have missed the comments. `DkHistogramKernel` should know the data and logic (e.g., how to calculate a histogram from image pixels), but not the representation of the histogram on the UI (e.g. the HTML layout and the text's position). I assume this was done to satisfy the multi-threading kernel interface, which returns an image; however, I think it is unexpected to find UI layout logic inside an image processing kernel.

Thanks for following up. I went ahead and merged just now before I saw this. The design is to satisfy caching of the rendering (with stats as they are dependent). This is nice to have there are ~768 lines to draw in addition to the rich text rendering - which otherwise has to re-render when zooming/panning, etc.

I see your point, while there is some layout logic here, there isn't any interaction with the UI so the benefit of any separation is minimal.

leejuyuu:
>  there isn't any interaction with the UI so the benefit of any separation is minimal.

I think this is a bit of a trap. Mixing UI logic makes it hard to prevent tight coupling, even if the interaction seems to be minimal. The overflow issue above is an example. The rendered image will always depend on many details that are not easy to control, e.g., the system font. Although it is not a necessary change at this moment, I think this is something to watch out for. 

As for the caching, caching the histogram calculation result or the render result is fundamentally different from caching transformed/processed images; therefore, I'd probably create a different mechanism instead of shoehorning it into the existing one.

PR #1538, opened 2026-03-12
fix(cmake): build tests when ENABLE_TESTING is on

bernardopg:
## Summary
- build the `tests` subdirectory whenever `ENABLE_TESTING` is enabled
- ensure `ctest` works after a normal default build
- keep the existing `check` target behavior intact

## Problem
`ENABLE_TESTING=ON` enables CTest and registers GoogleTest cases, but the tests subdirectory is currently added with `EXCLUDE_FROM_ALL`. As a result, a normal `cmake --build` does not build `core_tests`, while `gtest_discover_tests()` still generates the `core_tests_NOT_BUILT` placeholder. Running `ctest` then fails until `core_tests` is built manually.

## Fix
Remove `EXCLUDE_FROM_ALL` from the tests subdirectory so the default build includes `core_tests` whenever testing is enabled.

## Validation
- configured a clean Ninja build with `-DENABLE_TESTING=ON -DENABLE_QUAZIP=ON`
- ran `cmake --build /home/bitter/.cache/nomacs-pr-build -j$(nproc)`
- ran `ctest --test-dir /home/bitter/.cache/nomacs-pr-build --output-on-failure`
- ran `cmake --build /home/bitter/.cache/nomacs-pr-build --target check`

Both test runs passed locally (3/3).


scrubbbbs:
We don't want to compile tests in the default build since it wastes time when developing nomacs. We want to have `ENABLE_TESTING=ON` and compile tests separately when we want them. 

I think that means this PR is a wontfix unless you can keep that behavior unchanged.

PR #1556, opened 2026-03-29
Refactor & fixes for viewport

leejuyuu:


scrubbbbs:
@leejuyuu is this ready for review? I just tried the build it it seems fine for the most part. If you feel it is almost there maybe now is a good time to review.

PR #1561, opened 2026-04-06
Fix rating persistence for images without existing metadata

### LINKED ISSUE #1482: Auto-create XMP rating container when assigning ratings
The following feature request stems from comments within #909 

**Is your feature request related to a problem? Please describe.**
When assigning image ratings in Nomacs (via number keys or star UI), the rating is not persisted if the image file does not already contain a valid rating metadata container (e.g. XMP-xmp:Rating). The UI briefly reflects the change, but the rating is lost when navigating away or restarting Nomacs.

Through testing, I confirmed that if a file is pre-seeded with XMP-xmp:Rating=0, Nomacs is then able to update the rating correctly using number keys. This indicates Nomacs can modify existing rating metadata, but does not currently create it when missing.

This behavior is confusing for users, as rating assignment appears to work but silently fails unless metadata is pre-initialized externally.

**Describe the solution you'd like**
When a user assigns a rating to an image and no rating metadata exists, Nomacs should automatically create a valid rating container and persist the value.

Specifically:
- On rating assignment, if XMP-xmp:Rating is not present:
-- Create the tag and write the assigned value
- If the tag is present:
-- Update it normally
- Writes should be metadata-only (no pixel recompression)
- Behavior should be consistent between number-key rating and star-click UI

Optionally (but ideally):
- Use XMP-xmp:Rating as the canonical field
- Avoid creating duplicate rating tags

**Describe alternatives you've considered**
As a workaround, users can batch-seed all image files with XMP-xmp:Rating=0 using external tools (e.g. ExifTool), after which Nomacs rating updates work as expected.

While effective, this requires users to:
- Discover the issue independently
- Modify files externally
- Understand metadata tooling

This is fragile and non-obvious for typical users.

**Additional context**
- This behavior was reproduced and confirmed by manually adding XMP-xmp:Rating=0 to an image that previously failed to retain ratings; Nomacs was then able to update the rating correctly.
- This suggests the core limitation is missing metadata initialization rather than inability to write ratings.
- Automating this initialization inside Nomacs would align behavior with user expectations and reduce reliance on external tools.

eeik:
Fixes: #1482

Nomacs already writes `Exif.Image.Rating`, `Exif.Image.RatingPercent`, `Xmp.xmp.Rating`, and `Xmp.MicrosoftPhoto.Rating`, but `DkMetaDataT::setRating()` returned early when the image started with empty metadata.

This update keeps the scope focused on rating persistence while addressing the review feedback from the initial attempt.

### What changed
- allow first-time non-zero ratings to be written when an Exiv2 image handle exists but the file starts with empty metadata
- treat unchanged ratings as a successful no-op
- reject invalid rating inputs before any metadata mutation occurs
- keep clear-rating as a no-op if no rating exists yet
- add focused tests for empty-metadata rating behavior and dirty-state transitions

### Addressed review points
1. Braces around single-line `if` statements
   - The touched guard clauses in `DkMetaDataT::setRating()` now use braces.
2. `getRating() == r` should return `true`
   - Re-applying the current rating now returns success instead of failure.
3. Invalid input should not write `ratingPercent = 0`
   - Ratings outside the supported range now return `false` immediately, before any metadata write or clear path can run.
4. Rebase on current `master`
   - This branch was recreated from current `master` as of April 12, 2026.
5. Unit test
   - Added `ImageLounge/tests/DkMetaData_test.cpp` and wired it into `ImageLounge/tests/CMakeLists.txt`.
   - Coverage includes:
     - initializing a rating on an image with empty metadata
     - unchanged-rating no-op success behavior
     - invalid-rating no-op behavior
     - dirty-state transitions after successful changes
     - clear-rating behavior
     - `clearExifState()` resetting the dirty flag after a rating change

### Scope note
Rotation/comment parity is intentionally out of scope for this PR. Those paths still have their own `no_data` guards and seemed better treated as a separate follow-up than folded into this rating-specific fix.

### Verification note
I could not run the nomacs test suite in my local environment because the required build/test toolchain was not available there, so the added test still needs maintainer or CI validation.

### DIFF
--- a/ImageLounge/src/DkCore/DkMetaData.cpp
+++ b/ImageLounge/src/DkCore/DkMetaData.cpp
@@ -403,6 +403,7 @@ int DkMetaDataT::getRating() const
     }
 
     // get Rating of Xmp Tag
+    // NOTE: -1 is a valid xmp rating (means rejected) but we ignore that
     if (!xmpData.empty()) {
         auto key = Exiv2::XmpKey("Xmp.xmp.Rating");
         auto pos = xmpData.findKey(key);
@@ -415,6 +416,7 @@ int DkMetaDataT::getRating() const
 
         // if xmpRating not found, try to find MicrosoftPhoto Rating tag
         if (xmpRating == -1) {
+            // FIXME: this fallback is incorrect as it gives rating percent
             key = Exiv2::XmpKey("Xmp.MicrosoftPhoto.Rating");
             pos = xmpData.findKey(key);
             if (pos != xmpData.end() && pos->count() != 0) {
--- a/ImageLounge/src/DkCore/DkMetaData.cpp
+++ b/ImageLounge/src/DkCore/DkMetaData.cpp
@@ -1231,8 +1231,22 @@ bool DkMetaDataT::setDescription(const QString &description)
 
 bool DkMetaDataT::setRating(int r)
 {
-    if (mExifState == not_loaded || mExifState == no_data || getRating() == r)
+    if (mExifState == not_loaded || !mExifImg) {
         return false;
+    }
+
+    const int currentRating = (mExifState == no_data) ? -1 : getRating();
+    if (currentRating == r) {
+        return false;
+    }
+
+    if (r < 0 || r > 5) {
+        return false;
+    }
+
+    if (currentRating < 0 && r == 0) {
+        return false;
+    }
 
     int16_t ratingPercent = 0;
     switch (r) {
@@ -1251,8 +1265,6 @@ bool DkMetaDataT::setRating(int r)
     case 5:
         ratingPercent = 99;
         break;
-    default:
-        r = 0;
     }
 
     Exiv2::ExifData &exifData = mExifImg->exifData(); // Exif.Image.Rating  - short
--- a/ImageLounge/src/DkCore/DkMetaData.cpp
+++ b/ImageLounge/src/DkCore/DkMetaData.cpp
@@ -1233,46 +1233,39 @@ bool DkMetaDataT::setDescription(const QString &description)
 
 bool DkMetaDataT::setRating(int r)
 {
+    if (r < 0 || r > 5) {
+        qWarning() << "[setRating] invalid argument";
+        return false;
+    }
+
     if (mExifState == not_loaded || !mExifImg) {
         return false;
     }
 
-    const int currentRating = (mExifState == no_data) ? -1 : getRating();
-    if (currentRating == r) {
-        return false;
-    }
-
-    if (r < 0 || r > 5) {
-        return false;
-    }
-
-    if (currentRating < 0 && r == 0) {
-        return false;
-    }
-
-    int16_t ratingPercent = 0;
-    switch (r) {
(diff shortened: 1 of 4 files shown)


ImageLounge/src/DkCore/DkMetaData.cpp
```diff
@@ -1231,7 +1231,11 @@ bool DkMetaDataT::setDescription(const QString &description)
 
 bool DkMetaDataT::setRating(int r)
 {
-    if (mExifState == not_loaded || mExifState == no_data || getRating() == r)
+    if (mExifState == not_loaded || !mExifImg)
```
scrubbbbs: We use braces around single-line if statement.

ImageLounge/src/DkCore/DkMetaData.cpp
```diff
@@ -1231,7 +1231,11 @@ bool DkMetaDataT::setDescription(const QString &description)
...
+        return false;
+
+    const int currentRating = (mExifState == no_data) ? -1 : getRating();
+    if (currentRating == r)
```
scrubbbbs: Shouldn't this be: 
```c++
if (getRating() == r) {
  return true;
}
```

scrubbbbs (changes requested):
Thanks for this! It appears to work at least for JPG and PNG images I tested.

I have a few nitpicks (see comments), also you need to rebase on current master branch.

Could a similar fix be applied to make rotation and comments work this way as well?

Also, it would be nice to have a unit test for this, I'm not sure if it depends on the exiv2 version being used.

ImageLounge/src/DkCore/DkMetaData.cpp
```diff
@@ -1255,6 +1259,9 @@ bool DkMetaDataT::setRating(int r)
         r = 0;
     }
 
+    if (mExifState == no_data && r == 0)
```
scrubbbbs: This line has a problem, if `no_data` is false, it writes a `ratingPercent` of 0 when given an invalid rating as input. I would suggest only returning false if input was invalid.

eeik:
This PR is refreshed with a cleaned single-commit history and your earlier review points addressed directly.

Summary of updates:
- added braces in the touched guard clauses
- changed unchanged-rating behavior to a successful no-op
- reject invalid ratings before any metadata mutation
- kept first-time non-zero rating writes working for empty-metadata images
- added focused tests for empty-metadata rating behavior and dirty-state transitions

Rotation/comments remain out of scope for this PR so the change remains narrow and easier to review.

scrubbbbs:
Thank you for the update. Is this PR AI-generated? This appears you did not run the unit test to validate it. And did not adhere to point "changed unchanged-rating behavior to a successful no-op"

Your unit test is failing,
```
[ RUN      ] DkMetaData.RatingInitializesAndClearsEmptyMetadata
[Exiv2] metadata is empty
/home/test/sw/nomacs/ImageLounge/tests/DkMetaData_test.cpp:45: Failure
Value of: meta.setRating(5)
  Actual: false
Expected: true
```

Returns false here if rating is the same as before:

image

Issue #1549, opened 2026-03-22
Metadata ribbon isn't updated when switching image through thumbnail view


spaghetti22:
**Describe the bug**
When switching from one image to the next through the thumbnail preview, the metadata ribbon initially still displays the previous image's metadata

**To Reproduce**
1. Open any image (A) in nomacs that is in a folder that has more than one image in it
2. Make sure the "metadata ribbon" is visible (Panels → Metadata Ribbon)
3. Go to the "thumbnail preview" (Panels → Thumbnail Preview)
4. Double click any image (B) different from A
5. Erroneous behavior: The metadata ribbon shows metadata of image A.

**Expected behavior**
The metadata ribbon shows metadata of image B.

**Screenshots**

<img width="1038" height="945" alt="Image" src="https://github.com/user-attachments/assets/adac5db8-a060-4fa2-84c0-9259aeb50d03" />

<img width="1038" height="945" alt="Image" src="https://github.com/user-attachments/assets/71a3944d-a13e-4248-b5f8-5e781dad20c4" />

<img width="1038" height="945" alt="Image" src="https://github.com/user-attachments/assets/4a08a0e7-de85-4e82-ae2e-801d5d51f4b1" />

after hiding and showing the metadata ribbon, the display is correct:

<img width="1038" height="945" alt="Image" src="https://github.com/user-attachments/assets/d6026fa6-2a81-4cfa-b98c-a557b50013c6" />

**Desktop (please complete the following information):**
 - Debian
 - 13
 - current version of nomacs from master branch

**Additional context**
I may try to fix this myself at some point, but do not currently have enough time to do so, so am reporting this so someone can look into it.

scrubbbbs:
Thanks, I can confirm the bug as well.

PR #1566, opened 2026-04-12
fix: issue #1549 - update metadata ribbon when switching between imag…

spaghetti22:
Fix issue #1549 where the metadata ribbon was not updated when switching from one image to another through the thumbnail view.


scrubbbbs:
The `isVisible()` check is intended to skip processing when the hud widget is disabled. I think this is important to avoid processing overhead for things that are not being used.

The problem as you noticed is it doesn't work when the widget is hidden due to being covered up by another widget, as `isVisible()` is a QWidget property and means visible to the UI layout/rendering system, but that is not what we want here.

The concept I'd like to try (this is a bit more involved, but I think it will work):
- make `isVisible()` a deleted function on `DkFadeMixin` (parent class of all hud widgets)
- add `isActive()` to `DkFadeMixin` for this purpose and tie it to when the widget is toggled on or off
- replace usages of `isVisible()` with `isActive()`

PR #1571, opened 2026-05-01
feat(viewport): add zoom_keep_same_size_or_fit setting

perk11:
Add a new setting to "When Displaying New Images". It combines the logic of "Keep Zoom if the size is the same" and "Zoom to fit".

I really like "Keep zoom if the size is the same" option but it breaks down when I later load smaller images and they are not zoomed in. I ended up frequently toggling this setting. This option helps combine what is the best of both worlds. If there is no zooming going on and an image if too small, it will be fit into the viewport. But scrolling between images while zoomed in is also possible.

- Fork [nomacs](https://github.com/nomacs/nomacs.git) and create your fix/feature branch from `master`
- Pull requests are only accepted if they pass all workflow checks. In addition: 
    - Must compile without warnings (check workflow logs)
	- Must pass clang-tidy checks (see `ImageLounge/.clang-tidy`)

Also have a look at our [CONTRIBUTING.md](https://github.com/nomacs/nomacs/blob/master/CONTRIBUTING.md) for additional coding guidelines.

### DIFF
--- a/ImageLounge/src/DkCore/DkBaseViewPort.cpp
+++ b/ImageLounge/src/DkCore/DkBaseViewPort.cpp
@@ -787,6 +787,11 @@ void DkBaseViewPort::updateImageMatrix(std::optional<DkSettings::keepZoom> keepZ
     case DkSettings::zoom_always_fit:
         zoomToFit();
         break;
+    case DkSettings::zoom_keep_same_size_or_fit:
+        if (!isSameSize) {
+            zoomToFit();
+        }
+        break;
 
     default:
         Q_UNREACHABLE();
--- a/ImageLounge/src/DkCore/DkSettings.h
+++ b/ImageLounge/src/DkCore/DkSettings.h
@@ -146,6 +146,7 @@ public:
         zoom_keep_same_size, // keep if prev image size is the same
         zoom_never_keep, // zoom to fit, disallow scale > 100%
         zoom_always_fit, // zoom to fit, allow scale > 100%
+        zoom_keep_same_size_or_fit, // keep if same size, otherwise zoom to fit, allow scale >100%
         zoom_end,
     };
 
--- a/ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
@@ -793,6 +793,9 @@ void DkDisplayPreference::createLayout()
         tr("If checked, the zoom level is only kept, if the image loaded has the same level as the previous."));
     keepZoomButtons[DkSettings::zoom_never_keep] = new QRadioButton(tr("Never keep zoom"), this);
     keepZoomButtons[DkSettings::zoom_always_fit] = new QRadioButton(tr("Always zoom to fit"), this);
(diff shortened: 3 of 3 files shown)


scrubbbbs:
Thanks for the PR. Did you read #1500? @leejuyuu thinks there is a problem with having an either/or condition on the scaling. E.g. what happens when a small image is zoomed and another small image of the same size is loaded (which behavior takes priority? Does it fit or keep the scaling?

Maybe we need a separate setting that places some constraint on how fit-to-window should work, e.g.
- default (no constraint)
- maximum 100% scale
- maximum x% scale (e.g 200%)
- something else??

There is in an open refactor of viewport #1556 which will affect this.

perk11:
I didn't see #1500, thanks for pointing out. I also had a thought that this setting is doing multiple things now, which is why so many options are needed for each combination. I'm open to trying to add a secondary setting for zoom to fit if that's the consensus.

>  what happens when a small image is zoomed and another small image of the same size is loaded (which behavior takes priority? Does it fit or keep the scaling?

Here I think the answer is it keeps scaling. Because if the user doesn't want the scaling, they can press Cltr+0 or use other ways to zoom out and it will persist between images. But if fit takes priority, it's impossible to easily keep it zoomed in.

As far as the refactor in #1556, so far this is a tiny change and should be easy to adapt to that, by just moving the code to the newly created DkViewPortTransformViewModel.

I personally really find the combination specifically enabled by the setting added here missing, but open to any other options, please let me know what you'd like me to do next:

1. Should I try to add a separate setting for how fit to window should work?

Here is how I see it could work: make it so that "Keep zoom if the size is the same" option is using this setting when the size is different, maybe we could change the wording there too to "Keep zoom if the size is the same, otherwise zoom to fit"). And then the could start with just 2 options: No constraint and "Enter a number in %" which would be another new setting. Other behaviors can probably be added later.  I'm open to discussing this based on what you'd like to add.

2. Should I rebase this work on top of the refactor going in #1556?

3. Or maybe we don't want to add more settings for now and just add this one radio, this seems to be the same ask as in #1500.

scrubbbbs:
> Should I try to add a separate setting for how fit to window should work?

I think so. Based on feedback I've seen there are two types of users, those who want to scale up small images and those who do not. I don't think there is a case for switching between one and the other all of the time - which would require another action/shortcut.

The option would be something like "Maximum zoom when fitting to window" with a few fixed values like 100%, 125%, 150%, 200%, etc and an unlimited option.

We could then make a few improvements to make it more clear:
- Change the wording so "fit to window" was used in place of "zoom to fit" or "scale to fit" etc to reduce the confusion. E.g. "Keep zoom if size is the same, otherwise fit to window" 
- Drop "Never Keep Zoom" as an option as it would be effectively the same as "Zoom to fit" as the new option toggles between these two.

@leejuyuu has been working a lot in this area so I'd be interested in his opinion before moving forward.

leejuyuu:
Thanks for the PR. I'd approach this differently. The problem here is that the two "keep zoom" options fall back to `zoom_never_keep` when there is no previous zoom or the same size condition is not met. Instead of adding new options, making the fallback configurable would be a more flexible way to handle this.

perk11:
Thanks, I can work on a separate setting. @leejuyuu  How close are you on #1556?. Should I be basing my work on your branch or master?

perk11:
I created a draft of the new setting in #1577 (based on master for now). Will do a bit more testing before submitting it for review. Closing this PR.

PR #1577, opened 2026-05-04
Add a settings for zoom level when using zoom to fit and remove "Never keep zoom" setting

perk11:

- Fork [nomacs](https://github.com/nomacs/nomacs.git) and create your fix/feature branch from `master`
- Pull requests are only accepted if they pass all workflow checks. In addition: 
    - Must compile without warnings (check workflow logs)
	- Must pass clang-tidy checks (see `ImageLounge/.clang-tidy`)

Also have a look at our [CONTRIBUTING.md](https://github.com/nomacs/nomacs/blob/master/CONTRIBUTING.md) for additional coding guidelines.

### DIFF
--- a/ImageLounge/src/DkCore/DkBaseViewPort.cpp
+++ b/ImageLounge/src/DkCore/DkBaseViewPort.cpp
@@ -293,8 +293,7 @@ void DkBaseViewPort::setImage(const QImage &newImg)
 {
     mImgStorage.setImage(newImg);
     mImgRect = QRectF(QPointF(), getImageSize());
-    const bool kz = DkSettingsManager::param().display().keepZoom;
-    updateImageMatrix(kz ? DkSettings::zoom_keep_same_size : DkSettings::zoom_never_keep);
+    updateImageMatrix(static_cast<DkSettings::keepZoom>(DkSettingsManager::param().display().keepZoom));
     update();
 }
 
@@ -781,9 +780,6 @@ void DkBaseViewPort::updateImageMatrix(std::optional<DkSettings::keepZoom> keepZ
         }
         break;
     }
-    case DkSettings::zoom_never_keep:
-        mWorldMatrix.reset();
-        break;
     case DkSettings::zoom_always_fit:
         zoomToFit();
         break;
--- a/ImageLounge/src/DkCore/DkBaseViewPort.cpp
+++ b/ImageLounge/src/DkCore/DkBaseViewPort.cpp
@@ -917,38 +917,8 @@ void DkBaseViewPort::zoomToFit()
     const QSizeF winSize = size();
     double zoomLevel = qMin(winSize.width() / imgSize.width(), winSize.height() / imgSize.height());
 
-    // apply maximum zoom limit when fitting to window
-    int maxZoomSetting = DkSettingsManager::param().display().maxZoomOnFit;
-    double maxZoom = 0;
-    switch (maxZoomSetting) {
-        case DkSettings::zoom_on_fit_100:
-            maxZoom = 1.0;
-            break;
-        case DkSettings::zoom_on_fit_125:
-            maxZoom = 1.25;
-            break;
-        case DkSettings::zoom_on_fit_150:
-            maxZoom = 1.5;
-            break;
-        case DkSettings::zoom_on_fit_200:
-            maxZoom = 2.0;
-            break;
-        case DkSettings::zoom_on_fit_300:
-            maxZoom = 3.0;
-            break;
-        case DkSettings::zoom_on_fit_400:
-            maxZoom = 4.0;
-            break;
-        case DkSettings::zoom_on_fit_500:
-            maxZoom = 5.0;
-            break;
-        case DkSettings::zoom_on_fit_unlimited:
-            maxZoom = 0;
-            break;
-        default:
-            Q_UNREACHABLE();
-    }
-
+    // maxZoomOnFit <= 0 means unlimited, otherwise it is the zoom factor (e.g., 1.0 = 100%, 2.0 = 200%)
+    double maxZoom = DkSettingsManager::param().display().maxZoomOnFit;
     if (maxZoom > 0 && zoomLevel > maxZoom) {
         zoomLevel = maxZoom;
     }
--- a/ImageLounge/src/DkCore/DkBaseViewPort.cpp
+++ b/ImageLounge/src/DkCore/DkBaseViewPort.cpp
@@ -776,7 +776,7 @@ void DkBaseViewPort::updateImageMatrix(std::optional<DkSettings::keepZoom> keepZ
         break;
     case DkSettings::zoom_keep_same_size: {
         if (!isSameSize) {
-            mWorldMatrix.reset();
+            zoomToFit();
         }
         break;
     }
@@ -915,11 +915,47 @@ void DkBaseViewPort::zoomToFit()
 {
     const QSizeF imgSize = getImageSize();
     const QSizeF winSize = size();
-    const double zoomLevel = qMin(winSize.width() / imgSize.width(), winSize.height() / imgSize.height());
+    double zoomLevel = qMin(winSize.width() / imgSize.width(), winSize.height() / imgSize.height());
+
+    // apply maximum zoom limit when fitting to window
+    int maxZoomSetting = DkSettingsManager::param().display().maxZoomOnFit;
+    double maxZoom = 0;
+    switch (maxZoomSetting) {
+        case DkSettings::zoom_on_fit_100:
+            maxZoom = 1.0;
+            break;
+        case DkSettings::zoom_on_fit_125:
+            maxZoom = 1.25;
+            break;
+        case DkSettings::zoom_on_fit_150:
+            maxZoom = 1.5;
+            break;
+        case DkSettings::zoom_on_fit_200:
+            maxZoom = 2.0;
+            break;
+        case DkSettings::zoom_on_fit_300:
+            maxZoom = 3.0;
+            break;
+        case DkSettings::zoom_on_fit_400:
+            maxZoom = 4.0;
+            break;
+        case DkSettings::zoom_on_fit_500:
+            maxZoom = 5.0;
+            break;
+        case DkSettings::zoom_on_fit_unlimited:
+            maxZoom = 0;
+            break;
+        default:
+            Q_UNREACHABLE();
+    }
+
+    if (maxZoom > 0 && zoomLevel > maxZoom) {
+        zoomLevel = maxZoom;
+    }
 
     if (zoomLevel > 1) {
         zoomTo(zoomLevel);
-    } else if (zoomLevel < 1 || (zoomLevel == 1 && mSvg)) {
+    } else {
         resetView();
     }
 }
--- a/ImageLounge/src/DkCore/DkSettings.cpp
+++ b/ImageLounge/src/DkCore/DkSettings.cpp
@@ -459,6 +459,10 @@ void DkSettings::load(QSettings &settings, bool defaults)
     settings.beginGroup("DisplaySettings");
 
     display_p.keepZoom = settings.value("keepZoom", display_p.keepZoom).toInt();
+    // Migrate removed zoom_never_keep to zoom_always_fit
+    if (display_p.keepZoom == removed_zoom_never_keep) {
+        display_p.keepZoom = zoom_always_fit;
+    }
     display_p.invertZoom = settings.value("invertZoom", display_p.invertZoom).toBool();
     display_p.highlightColor = QColor::fromRgba(
         settings.value("highlightColorRGBA", display_p.highlightColor.rgba()).toInt());
--- a/ImageLounge/src/DkCore/DkSettings.cpp
+++ b/ImageLounge/src/DkCore/DkSettings.cpp
@@ -463,6 +463,7 @@ void DkSettings::load(QSettings &settings, bool defaults)
     if (display_p.keepZoom == removed_zoom_never_keep) {
         display_p.keepZoom = zoom_always_fit;
     }
+    display_p.maxZoomOnFit = settings.value("maxZoomOnFit", display_p.maxZoomOnFit).toInt();
     display_p.invertZoom = settings.value("invertZoom", display_p.invertZoom).toBool();
     display_p.highlightColor = QColor::fromRgba(
         settings.value("highlightColorRGBA", display_p.highlightColor.rgba()).toInt());
@@ -717,6 +718,8 @@ void DkSettings::save(QSettings &settings, bool force)
 
     if (force || display_p.keepZoom != display_d.keepZoom)
         settings.setValue("keepZoom", display_p.keepZoom);
+    if (force || display_p.maxZoomOnFit != display_d.maxZoomOnFit)
+        settings.setValue("maxZoomOnFit", display_p.maxZoomOnFit);
     if (force || display_p.invertZoom != display_d.invertZoom)
         settings.setValue("invertZoom", display_p.invertZoom);
     if (force || display_p.highlightColor != display_d.highlightColor)
@@ -973,6 +976,7 @@ void DkSettings::setToDefaultSettings()
 #endif
 
     display_p.keepZoom = zoom_keep_same_size;
+    display_p.maxZoomOnFit = zoom_on_fit_unlimited;
     display_p.invertZoom = false;
     display_p.highlightColor = QColor(0, 204, 255);
     display_p.hudBgColor = QColor(0, 0, 0, 100);
--- a/ImageLounge/src/DkCore/DkSettings.cpp
+++ b/ImageLounge/src/DkCore/DkSettings.cpp
@@ -459,11 +459,14 @@ void DkSettings::load(QSettings &settings, bool defaults)
     settings.beginGroup("DisplaySettings");
 
     display_p.keepZoom = settings.value("keepZoom", display_p.keepZoom).toInt();
+    display_p.maxZoomOnFit = settings.value("maxZoomOnFit", display_p.maxZoomOnFit).toDouble();
     // Migrate removed zoom_never_keep to zoom_always_fit
     if (display_p.keepZoom == removed_zoom_never_keep) {
         display_p.keepZoom = zoom_always_fit;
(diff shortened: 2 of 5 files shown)


perk11:
@scrubbbbs @leejuyuu This removes "Never keep zoom" and implements a new setting as discussed  in #1571. Let me know what you think. I didn't find any issues in my testing.

ImageLounge/src/DkCore/DkSettings.h
```diff
@@ -209,6 +222,7 @@ class DllCoreExport DkSettings
 
     struct Display {
         int keepZoom;
+        int maxZoomOnFit;
```
scrubbbbs: I think changing this to mean <=0 is unlimited and otherwise it is the zoom percentage or factor could be cleaner. The enum and switch block can be dropped, the combo box setup can use the `userData` argument of `addItem()` to establish the mapping in one location.

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
```diff
@@ -788,28 +788,42 @@ void DkDisplayPreference::createLayout()
...
     keepZoomButtons.resize(DkSettings::zoom_end);
     keepZoomButtons[DkSettings::zoom_always_keep] = new QRadioButton(tr("Always keep zoom"), this);
-    keepZoomButtons[DkSettings::zoom_keep_same_size] = new QRadioButton(tr("Keep zoom if the size is the same"), this);
+    keepZoomButtons[DkSettings::zoom_keep_same_size] = new QRadioButton(
+        tr("Keep zoom if the size is the same, otherwise fit to window"), this);
     keepZoomButtons[DkSettings::zoom_keep_same_size]->setToolTip(
-        tr("If checked, the zoom level is only kept, if the image loaded has the same level as the previous."));
-    keepZoomButtons[DkSettings::zoom_never_keep] = new QRadioButton(tr("Never keep zoom"), this);
-    keepZoomButtons[DkSettings::zoom_always_fit] = new QRadioButton(tr("Always zoom to fit"), this);
+        tr("If selected, if the image loaded has the exact same size as the previous, keep zoom level, otherwise zoom the image to fit the window dimensions"));
```
scrubbbbs: Simplify to something like: "If new image has the same size as previous image, keep zoom level. Otherwise zoom the image to fit the window."

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
```diff
@@ -983,6 +998,12 @@ void DkDisplayPreference::onKeepZoomButtonClicked(int buttonId) const
         DkSettingsManager::param().display().keepZoom = buttonId;
 }
 
+void DkDisplayPreference::onMaxZoomOnFitChanged(int index) const
+{
+    if (DkSettingsManager::param().display().maxZoomOnFit != index)
```
scrubbbbs: If you use the userData of combo box you can validate and pull the factor here before assigning to settings.

ImageLounge/src/DkCore/DkBaseViewPort.cpp
```diff
@@ -777,13 +776,10 @@ void DkBaseViewPort::updateImageMatrix(std::optional<DkSettings::keepZoom> keepZ
         break;
     case DkSettings::zoom_keep_same_size: {
         if (!isSameSize) {
-            mWorldMatrix.reset();
+            zoomToFit();
```
scrubbbbs, line 779: Please check this is also working in frameless mode as it could have a different logic to account for.

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
```diff
@@ -788,28 +788,42 @@ void DkDisplayPreference::createLayout()
...
+    // Maximum zoom when fitting to window
+    auto *maxZoomOnFitCombo = new QComboBox(this);
+    maxZoomOnFitCombo->addItem(tr("100%"));
+    maxZoomOnFitCombo->addItem(tr("125%"));
+    maxZoomOnFitCombo->addItem(tr("150%"));
+    maxZoomOnFitCombo->addItem(tr("200%"));
+    maxZoomOnFitCombo->addItem(tr("300%"));
+    maxZoomOnFitCombo->addItem(tr("400%"));
+    maxZoomOnFitCombo->addItem(tr("500%"));
+    maxZoomOnFitCombo->addItem(tr("Unlimited"));
```
scrubbbbs: I think unlimited should be the first option, trying to keep the convention of defaults being the first option in combo boxes.

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
```diff
@@ -788,28 +788,42 @@ void DkDisplayPreference::createLayout()
...
     auto *keepZoomGroup = new DkGroupWidget(tr("When Displaying New Images"), this);
     keepZoomGroup->addWidget(keepZoomButtons[DkSettings::zoom_always_keep]);
     keepZoomGroup->addWidget(keepZoomButtons[DkSettings::zoom_keep_same_size]);
-    keepZoomGroup->addWidget(keepZoomButtons[DkSettings::zoom_never_keep]);
     keepZoomGroup->addWidget(keepZoomButtons[DkSettings::zoom_always_fit]);
 
+    // Maximum zoom when fitting to window
+    auto *maxZoomOnFitCombo = new QComboBox(this);
+    maxZoomOnFitCombo->addItem(tr("100%"));
+    maxZoomOnFitCombo->addItem(tr("125%"));
```
scrubbbbs: To reduce the translation footprint please use `tr("%1%").arg(100)` etc. Unfortunately, there is no percentage formatter in QLocale, but this will be one string to translate instead of 7.

leejuyuu:
- Why remove the "zoom_never_keep" option? The logic behind it is different from zoom to fit. 
- This changes the fallback of "keep zoom if same size" from "zoom_never_keep" to "zoom_to_fit", but maintains no flexibility.  
- Adding options with explicit numbers will be inflexible. They should be merged into a single option, with a numeric input to set the level. Also, the fixed level seems to be max zoom, meaning that images smaller will stay at 100%? This might not be intuitive as well. 

I am still in favor of the options list in [this comment in #1500](https://github.com/nomacs/nomacs/issues/1500#issuecomment-3940639942). When the keep zoom option is chosen, another dropdown should be activated to select a fallback from options 1~4.

scrubbbbs:
> * Why remove the "zoom_never_keep" option? The logic behind it is different from zoom to fit. 
>     * This changes the fallback of "keep zoom if same size" from "zoom_never_keep" to "zoom_to_fit", but maintains no flexibility.

never keep is confusing as it has a different fit logic than view/fit-to-window. There is no view action to invoke what never keep is doing. This PR makes that logic the same, and makes fit to window configurable instead.

>     * Adding options with explicit numbers will be inflexible. They should be merged into a single option, with a numeric input to set the level. Also, the fixed level seems to be max zoom, meaning that images smaller will stay at 100%? This might not be intuitive as well.

This is an attempt to simplify, scale factors < 100% are invalid and we also have the unlimited option. 100% and unlimited will be the most common use case.

> I am still in favor of the options list in [this comment in #1500](https://github.com/nomacs/nomacs/issues/1500#issuecomment-3940639942). When the keep zoom option is chosen, another dropdown should be activated to select a fallback from options 1~4.

Besides dropping never keep option I think that list remains valid.

If we want to have configurable fallbacks later, this PR is compatible. There just won't be a separate fallback for 100% max zoom and unlimited (or some other factor). We are just altering fit to window behavior to be consistent.

ImageLounge/src/DkCore/DkBaseViewPort.cpp
perk11: I checked and didn't notice anything working differently in frameless mode.

ImageLounge/src/DkCore/DkSettings.h
perk11: Changed it to double corresponding to zoomLevel directly. I was deliberating if should be an int with % amount and then divide it by 100 during the calculation, but keeping it as double seems to be one less line of code.

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
perk11: Updated

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
perk11: I implemented this inside the new lambda in connect()

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
perk11: Implemented

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
perk11: Implemented

perk11:
@scrubbbbs All the requested changes are implemented

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
```diff
@@ -788,28 +788,53 @@ void DkDisplayPreference::createLayout()
...
+    maxZoomOnFitCombo->addItem(tr("%1%").arg(400), 4.0);
+    maxZoomOnFitCombo->addItem(tr("%1%").arg(500), 5.0);
+    const int maxZoomIndex = maxZoomOnFitCombo->findData(DkSettingsManager::param().display().maxZoomOnFit);
+    maxZoomOnFitCombo->setCurrentIndex(maxZoomIndex >= 0 ? maxZoomIndex : 0);
+    connect(maxZoomOnFitCombo,
+            QOverload<int>::of(&QComboBox::currentIndexChanged),
+            this,
+            [maxZoomOnFitCombo](const int currentIndex) {
+                const double selectedOptionSettingValue = maxZoomOnFitCombo->itemData(currentIndex).toDouble();
+                if (DkSettingsManager::param().display().maxZoomOnFit != selectedOptionSettingValue) {
```
scrubbbbs: The if statement isn't necessary here. Settings in nomacs are just global variables (no operator overloading tricks here etc).

`selectedOptionSettingValue` is generic, why not call it what it is, e.g. `maxZoomOnFit` or just drop this variable entirely.

scrubbbbs (approved):
I just tested and it seems to be OK. Just the one nitpick left.

leejuyuu:
> never keep is confusing as it has a different fit logic than view/fit-to-window.

Ok. I understand what the "Maximum zoom when fitting to window" is trying to do now.


> This is an attempt to simplify, scale factors < 100% are invalid and we also have the unlimited option. 100% and unlimited will be the most common use case.

As you said, "100% and unlimited will be the most common use case." Therefore, it is not certain whether those options will be used at this time, and we cannot know in advance what other options users will request. Which means, setting these options is premature now. To make it the most flexible, I suggest making this setting into a check box that says "Limit max zoom level when fitting to window," which, when checked,  activates a numeric input for the zoom level that defaults to 100%.

 
> Besides dropping never keep option I think that list remains valid.
> 
> If we want to have configurable fallbacks later, this PR is compatible. There just won't be a separate fallback for 100% max zoom and unlimited (or some other factor). We are just altering fit to window behavior to be consistent.

This PR changes the fallback logic. This is a behavior change that I cannot accept lightly unless it is configurable.

ImageLounge/src/DkCore/DkSettings.cpp
```diff
@@ -459,6 +459,11 @@ void DkSettings::load(QSettings &settings, bool defaults)
     settings.beginGroup("DisplaySettings");
 
     display_p.keepZoom = settings.value("keepZoom", display_p.keepZoom).toInt();
+    // Migrate removed zoom_never_keep to zoom_always_fit
+    if (display_p.keepZoom == removed_zoom_never_keep) {
+        display_p.keepZoom = zoom_always_fit;
+    }
```
leejuyuu, line 469: `maxZoomOnFit` should be set to 1 here to preserve the old `zoom_never_keep` behavior.

perk11:
> To make it the most flexible, I suggest making this setting into a check box that says "Limit max zoom level when fitting to window," which, when checked, activates a numeric input for the zoom level that defaults to 100%.

image

The only thing that's not allowed right now is entering an arbitrary value, everything else is already possible. I was also leaning towards a text input but @scrubbbbs wanted a dropdown, I think it doesn't really matter, I doubt the users will need to adjust it this precisely, but open to changing it if there is a consensus.

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
perk11: Got it, I removed the temporary variable altogether.

ImageLounge/src/DkCore/DkSettings.cpp
perk11: Implemented. I had to add saving of the settings, otherwise this could override the existing `maxZoomOnFit` value.

scrubbbbs:
> > This is an attempt to simplify, scale factors < 100% are invalid and we also have the unlimited option. 100% and unlimited will be the most common use case.
> 
> As you said, "100% and unlimited will be the most common use case." Therefore, it is not certain whether those options will be used at this time, and we cannot know in advance what other options users will request. Which means, setting these options is premature now. To make it the most flexible, I suggest making this setting into a check box that says "Limit max zoom level when fitting to window," which, when checked, activates a numeric input for the zoom level that defaults to 100%.

I guess this would be fine, there isn't much difference in this case.

I would be mindful of choosing flexibility over ease of use because it leads to maintenance problems. We can always add flexibility, it is not so easy to take it away if we don't think it is needed anymore.

> > Besides dropping never keep option I think that list remains valid.
> > If we want to have configurable fallbacks later, this PR is compatible. There just won't be a separate fallback for 100% max zoom and unlimited (or some other factor). We are just altering fit to window behavior to be consistent.
> 
> This PR changes the fallback logic. This is a behavior change that I cannot accept lightly unless it is configurable.

So previously, the fallback for "keep zoom if same size" was "fit to window not exceeding 100%". Now it is "fit to window" which is configurable to "limit zoom to 100%". So transitively, the fallback behavior is configurable.

leejuyuu:
> So previously, the fallback for "keep zoom if same size" was "fit to window not exceeding 100%". Now it is "fit to window" which is configurable to "limit zoom to 100%". So transitively, the fallback behavior is configurable.

No, it isn't. This PR sets `maxZoomOnFit` default to 0, which is the original zoom to fit behavior. So, this **is** a behavior change for the "keep zoom if same size" users. However, it also cannot be set to 1 because it will then affect "zoom to fit" users. This design is making more trouble than the problem it is trying to solve.

scrubbbbs:
> > So previously, the fallback for "keep zoom if same size" was "fit to window not exceeding 100%". Now it is "fit to window" which is configurable to "limit zoom to 100%". So transitively, the fallback behavior is configurable.
> 
> No, it isn't. This PR sets `maxZoomOnFit` default to 0, This design is making more trouble than the problem it is trying to solve.

OK, I can see the problem of Ctrl+0 not always doing the same thing depending on settings. But if fit logic is going to be expanded then there will be a lot of options. Unfortunately I do not see a design where omitting some of these is going to satisfy everyone in the end.

The alternative expands into at least 8 different fit behaviors:
- Fit longest side (current fit-to-window), shortest side (fill window), fit width, fit height, and all of those doubled again:
- Fit longest side (alternate) maybe Ctrl+Shift+0, shortest side (alternate) .. etc

And the the "alternate" mode would apply features:
- max zoom setting
- fit faces, e.g. do face detection and zoom to all faces.
- fit content, e.g. detect and zoom past the border pixels/frame around image
- fit to metadata crop rectangle

This approach would unwind this PR and make the fallback configurable to one of these 8 options.

leejuyuu:
Mixing feature addition is making things more complicated. I think we need to focus the discussion back on the goal of the [first PR](https://github.com/nomacs/nomacs/pull/1571), which, as I understand, is to change the fallback logic of "Keep zoom if the size is the same" from `zoom_never_keep` to `zoom_to_fit`. I think the simplest way to achieve this is to revert the changes to the "When displaying new images" options, and add a separate dropdown to select the fallback for the keep zoom options.

scrubbbbs:
> Mixing feature addition is making things more complicated. I think we need to focus the discussion back on the goal of the [first PR](https://github.com/nomacs/nomacs/pull/1571), [...]

If @perk11 gets what he is after that's fine with me. Since you seem to want to own the viewport related issues I'll step aside on this one.

perk11:
@leejuyuu

What I want is behavior that existed in 3.21 with settings like this, but became impossible in 3.22, it was removed in this commit 70c5323e7ec1c690f3fc1238f7124f43a307da60
image

If you think the dropdown with % numbers is the solution, I can adjust the code.

If I revert the "When displaying new images" change. Does the dropdown that I already added work, or does it need any changes?
Should it also being controlling what happens when someone presses Ctrl+0? Right now it does.

leejuyuu:
> If you think the dropdown with % numbers is the solution, I can adjust the code.
> 
> If I revert the "When displaying new images" change. Does the dropdown that I already added work, or does it need any changes? 

The current implementation of "Maximum Zoom When Fitting to Window", i.e., the `maxZoomOnFit` is shared between "zoom_to_fit" and the fallback of "keep_zoom_if_same_size" options. However, I think it is impossible to set a default of `maxZoomOnFit` to preserve the behavior before this PR, since the original "zoom_to_fit" requires `maxZoomOnFit = 1` while "keep_zoom_if_same_size" requires `maxZoomOnFit = 0`. This configuration is a new feature and is not directly related to your request, so I think we should not add this dropdown before we have a way to address this.

> Should it also being controlling what happens when someone presses Ctrl+0? Right now it does.

The current behavior of `Ctrl + 0` is always zoom to fit and does not consider keeping the zoom level. Is this expected for you?

leejuyuu:
I am thinking a dropdown below the current "Always zoom to fit" that says "If cannot keep zoom, fallback to", which has two options "Never keep zoom" (we should probably change the wording later) and "Zoom to fit".

perk11:
@leejuyuu Sorry about delay, life got in the way of this. I posted a version similar to what you suggested in #1591, please review my comments there.

PR #1581, opened 2026-05-13
Allow comments or other metadata to be created if none exists

scrubbbbs:
This is a followup to #1561 that enables the same behavior for comments. I'm undecided what to do about rotation/orientation, that remains unchanged at the moment.


leejuyuu:
This seems to work for JPG, but not for WebP. The file saving seems to succeed, though. Is there any restriction on the file type? Can it be warned early so it does not fail silently?

scrubbbbs:
This is a bug, there will be a message now. While webp has write support, it doesn't work the way we are using exiv2 to save metadata or edit history.

leejuyuu:
Works now, thanks!

PR #1583, opened 2026-05-15
Refactor image load options; fix UI state, metadata display, settings, and batch processing

scrubbbbs:
Various low-impact bugs

Reworked prefs/file layout:
image


ImageLounge/src/DkCore/DkThumbs.h
```diff
@@ -62,7 +62,7 @@ enum class LoadThumbnailOption {
     none,
 
     // Only load EXIF thumbnail.
-    force_exif,
+    // force_exif,
```
leejuyuu: Does this mean loading the whole image for the thumbnail cannot be disabled?

ImageLounge/src/DkCore/DkThumbs.cpp
```diff
@@ -106,6 +106,7 @@ std::optional<LoadThumbnailResult> loadThumbnail(const LoadThumbnailRequest &req
     std::unique_ptr<DkCachedThumb> cachedThumb{};
 
     if (request.option != LoadThumbnailOption::force_full //
+        && !(loadOptions & DkLoadOption::source) //
```
leejuyuu, line 109: This is only used by DkThumbs, shouldn't it be part of `LoadThumbnailOption`?

ImageLounge/src/DkCore/DkBasicWidgets.h
```diff
@@ -61,6 +61,12 @@ class DllCoreExport DkSlider : public DkWidget
     int value() const;
     void setValueSuffix(const QString &format);
     void setFocus(Qt::FocusReason reason);
+
+    QSpinBox *getSpinBox() const
+    {
+        return sliderBox;
+    }
```
leejuyuu: This will expose the whole spinbox. If only `setFixedWidth()` is needed, maybe `setSpinBoxFixedWidth(int)` is enough and preserves encapsulation.

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
```diff
@@ -1108,54 +1108,57 @@ void DkFilePreference::createLayout()
...
-    tempFolderGroup->addWidget(tLabel);
+
+    const int maxRam = DkMemory::maxImageAlloc();
+
+    auto &res = DkSettingsManager::param().resources();
+
+    auto *maxAlloc = new DkSlider(tr("Image allocation limit"), this);
+    maxAlloc->setToolTip(tr("Maximum memory allowed for any image. If exceeded, image will not be loaded."));
+    maxAlloc->setRange(1024, maxRam);
+    maxAlloc->setValueSuffix(tr(" MB"));
```
leejuyuu: I think leaving leading or trailing whitespaces is going to make translation more complicated. Can it be done with `" " + tr("MB")`? Also, maybe we do not even need to translate "MB".

leejuyuu (commented):
I suggest changing the PR title, as this branch contains new features.

ImageLounge/src/DkCore/DkThumbs.h
scrubbbbs: Yes, full image is loaded if exif is not present. We do not use it anymore (or never used it). There is a different path for metadata dock.

ImageLounge/src/DkCore/DkThumbs.cpp
scrubbbbs: LoadThumbnailOption is different, more than one option cannot be set. It could be refactored to be flags but I don't see the benefit.

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
scrubbbbs: There are some languages where no space is often preferred (Chinese, Japanese, Korean). But maybe that is an unnecessary nitpick. As for do we need it, maybe not,  Russian and French have changed it in our current files but probably don't have to.

ImageLounge/src/DkCore/DkThumbs.cpp
leejuyuu: I would be cautious, as this blurs the class/module boundaries.

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
leejuyuu: If "MB" is used widely enough, I suggest keeping it untranslated.

PR #1591, opened 2026-05-29
feat(viewport) Add a fallback setting for "Keep zoom if size is the same"

perk11:
I went based off this suggestion
image
but since the option only really applies to "Keep zoom if the size is the same", I put it under that setting and made it disabled if another option is selected. Also, I named the version that matches behavior of "Never keep zoom" "Zoom to fit if image is bigger than view port, otherwise reset to 100%", since that's what describes what really is going on. Without such a description I think it's really hard for user to figure out how this works.

image
image

### DIFF
--- a/ImageLounge/src/DkCore/DkBaseViewPort.cpp
+++ b/ImageLounge/src/DkCore/DkBaseViewPort.cpp
@@ -777,7 +777,17 @@ void DkBaseViewPort::updateImageMatrix(std::optional<DkSettings::keepZoom> keepZ
         break;
     case DkSettings::zoom_keep_same_size: {
         if (!isSameSize) {
-            mWorldMatrix.reset();
+            switch (static_cast<DkSettings::keepZoomKeepSameSizeFallback>(
+                DkSettingsManager::param().display().keepZoomKeepSameSizeFallback)) {
+            case DkSettings::kzkss_fallback_fit_bigger_or_reset:
+                mWorldMatrix.reset();
+                break;
+            case DkSettings::kzkss_fallback_always_fit:
+                zoomToFit();
+                break;
+            default:
+                Q_UNREACHABLE();
+            }
         }
         break;
     }
--- a/ImageLounge/src/DkCore/DkSettings.cpp
+++ b/ImageLounge/src/DkCore/DkSettings.cpp
@@ -455,6 +455,8 @@ void DkSettings::load(QSettings &settings, bool defaults)
     settings.beginGroup("DisplaySettings");
 
     display_p.keepZoom = settings.value("keepZoom", display_p.keepZoom).toInt();
+    display_p.keepZoomKeepSameSizeFallback = settings.value("keepZoomFallback", display_p.keepZoomKeepSameSizeFallback)
+                                                 .toInt();
     display_p.invertZoom = settings.value("invertZoom", display_p.invertZoom).toBool();
     display_p.highlightColor = QColor::fromRgba(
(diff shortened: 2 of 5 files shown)


perk11:
@leejuyuu Could you please review?

ImageLounge/src/DkCore/DkSettings.cpp
```diff
@@ -454,6 +454,8 @@ void DkSettings::load(QSettings &settings, bool defaults)
     settings.beginGroup("DisplaySettings");
 
     display_p.keepZoom = settings.value("keepZoom", display_p.keepZoom).toInt();
+    display_p.keepZoomKeepSameSizeFallback = settings.value("keepZoomFallback", display_p.keepZoomKeepSameSizeFallback)
```
leejuyuu, line 458: I suggest keeping the name consistent to `keepZoomFallback` across this PR.

ImageLounge/src/DkCore/DkSettings.h
```diff
@@ -149,6 +149,12 @@ class DllCoreExport DkSettings
         zoom_end,
     };
 
+    enum keepZoomKeepSameSizeFallback {
+        kzkss_fallback_fit_bigger_or_reset,
+        kzkss_fallback_always_fit,
+        kzkss_fallback_end,
```
leejuyuu, line 155: Please replace `kzkss` with `keep_zoom`.

ImageLounge/src/DkCore/DkBaseViewPort.cpp
```diff
@@ -777,7 +777,17 @@ void DkBaseViewPort::updateImageMatrix(std::optional<DkSettings::keepZoom> keepZ
...
+            switch (static_cast<DkSettings::keepZoomKeepSameSizeFallback>(
+                DkSettingsManager::param().display().keepZoomKeepSameSizeFallback)) {
+            case DkSettings::kzkss_fallback_fit_bigger_or_reset:
+                mWorldMatrix.reset();
+                break;
+            case DkSettings::kzkss_fallback_always_fit:
+                zoomToFit();
+                break;
+            default:
+                Q_UNREACHABLE();
```
leejuyuu, line 789: Using `Q_UNREACHABLE` here would make the program crash if someone edits the value in the config file to something like 5. Either make this default to `mWorldMatrix.reset()` or validate the value in the settings.

ImageLounge/src/DkGui/DkPreferenceWidgets.cpp
```diff
@@ -809,8 +809,42 @@ void DkDisplayPreference::createLayout()
...
+
+    QVector<QRadioButton *> keepZoomKeepSameSizeFallbackButtons;
+    keepZoomKeepSameSizeFallbackButtons.resize(DkSettings::kzkss_fallback_end);
+    keepZoomKeepSameSizeFallbackButtons
+        [DkSettings::kzkss_fallback_fit_bigger_or_reset] = new QRadioButton(tr("Zoom to fit if image is bigger than "
+                                                                               "viewport, otherwise reset to 100%"),
+                                                                            this);
+    keepZoomKeepSameSizeFallbackButtons
+        [DkSettings::kzkss_fallback_always_fit] = new QRadioButton(tr("Always zoom to fit"), this);
+    keepZoomKeepSameSizeFallbackButtons[DkSettingsManager::param().display().keepZoomKeepSameSizeFallback]->setChecked(
```
leejuyuu, line 841: This also crashes when the config value is edited to out-of-range values.

leejuyuu (changes requested):

PR #1614, opened 2026-06-27
Add Unique ID to shortcuts and refactor

scrubbbbs:
Add unique id to shortcuts to fix shortcuts bound to translated text and provide a path to customize toolbars.  This is via the new `DkActionBuilder` which helps clean up action creation and requires a unique id for each action.

~The main issue with this PR is that all existing custom shortcuts must be rebound. I don't see that has a huge deal but we could do something about it.~


ImageLounge/src/DkCore/DkActionManager.h
```diff
@@ -91,6 +91,93 @@ public slots:
     bool mFirstTime = true;
 };
 
+struct DkActionId {
+    int index; // index into the associated actions array
+    const char *objectName; // uuid for settings, etc
```
leejuyuu: 1. From how the ACTION_ID seems to work, I think the `objectName` is not a UUID? I would avoid using this term in the comment if that is the case.
2. As these come from expression strings, I do not see how they are guaranteed to be unique.
3. I would recommend using `std::string_view` instead.

ImageLounge/src/DkCore/DkActionManager.cpp
```diff
@@ -64,25 +64,31 @@ DkAppManager::DkAppManager(QWidget *parent)
 {
     loadSettings();
 
+    QString name, id;
```
leejuyuu: Recommend avoid declaring multiple variables on the same line.

ImageLounge/src/DkCore/DkPluginManager.cpp
```diff
@@ -1425,38 +1425,36 @@ QVector<QMenu *> DkPluginActionManager::pluginSubMenus() const
 
 void DkPluginActionManager::updateMenu()
 {
-    qDebug() << "CREATING plugin menu";
-
     if (!mMenu) {
-        qWarning() << "plugin menu is NULL where it should not be!";
+        Q_ASSERT(mMenu);
```
leejuyuu: Redundant Q_ASSERT.

ImageLounge/tests/DkShortcutsModel_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+    // that is required as there is global state destructor won't clean up
+    static QApplication app(argc, argv);
+
+    // these are needed for nomacs to find its settings file, we use
+    // different values from release to isolate from development environment
+    QCoreApplication::setOrganizationName("nomacs-test");
+    QCoreApplication::setOrganizationDomain("https://test.nomacs.org");
+    QCoreApplication::setApplicationName("ImageLounge-test");
+    QCoreApplication::setApplicationVersion(NOMACS_VERSION_STR);
+}
```
leejuyuu: This seems to need its own special environment? If that is the case, I suggest separating this test from the others that can run without extra dependencies.

ImageLounge/tests/DkShortcutsModel_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+    // stand up QApplication compatible with headless CI environment
+    int argc = 3;
+    static char *argv[3];
+    argv[0] = const_cast<char *>("nomacs-test");
+    argv[1] = const_cast<char *>("-platform");
+    argv[2] = const_cast<char *>("offscreen");
+
+    // static is fine here since each test runs as separate process, and
+    // that is required as there is global state destructor won't clean up
+    static QApplication app(argc, argv);
```
leejuyuu: argc needs to be static as well.

ImageLounge/tests/DkShortcutsModel_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "DkSettings.h"
+#include "DkVersion.h"
+
+#include <QApplication>
+#include <QMainWindow>
+#include <QMenu>
+#include <QRectF>
+#include <gtest/gtest.h>
+
+using namespace nmc;
```
leejuyuu: I would avoid this.

ImageLounge/src/DkCore/DkActionManager.cpp
```diff
@@ -1786,17 +1786,29 @@ void DkActionManager::assignCustomShortcuts(QVector<QAction *> actions) const
...
     for (QAction *a : actions) {
         if (!a) {
             qWarning() << "NULL action detected!";
             continue;
         }
 
         QString actionId = a->objectName();
-        QString val = settings.value(actionId, "no-shortcut").toString();
+        QVariant val = settings.value(actionId);
+        if (val == unset) {
```
leejuyuu: Why not use isValid() or isNull()?

ImageLounge/src/DkCore/DkPluginManager.cpp
```diff
@@ -1518,12 +1518,16 @@ void DkPluginActionManager::runPluginFromShortcut()
...
-        settings.setValue(actions.at(i)->text(), actions.at(i)->text());
-    settings.endGroup();
+
+    // remove possibly outdated plugin data
+    settings.remove("PluginActions");
+
+    settings.beginGroup("PluginActions");
+    for (QAction *a : actions) {
+        settings.setValue(a->objectName(), a->text());
+    }
```
leejuyuu, line 1554: This changes the settings key? Is there an associated migration?

leejuyuu (commented):
Two major issues:

1. I think it might be better to separate the unique ID from the enum identifier name. It seems possible that these unique IDs end up in the config file, so they become a compatibility contract in the future. This makes changing the names of these enums, like several commits in this branch do, very hard or impossible. On the other hand, it is not obvious that the names of these enums need to be unique and fulfill this contract, and someone may change them without noticing. I suggest separating the unique identifier from the enum names in code for long-term stability.

2. How are third-party plugin action IDs validated and namespaced? The plugin interface does not currently require actions to provide a stable, unique objectName(). Will this break existing third-party plugins? How do we prevent accidental ID collision between plugins and between nomacs and a plugin?

ImageLounge/src/DkCore/DkActionManager.h
scrubbbbs: Not a UUID in the RFC sense but must be unique for all time once created. So maybe it should be renamed to clear that up. Since it is stored to the `objectName` property of `QAction` so it ended up this way.

The only guarantee of unique ids is currently the unit test since you correctly asserted there is no uuid validation here. A small change to `DkActionBuilder` could add each action to a common hashmap - which we will need for toolbar customization - so we could check this now or later on.

ImageLounge/src/DkCore/DkPluginManager.cpp
scrubbbbs: No migration is needed, IIRC the plugin shortcut binds end up under "CustomShortcuts" key.

scrubbbbs:
Thanks for taking time to review.

> 1. I think it might be better to separate the unique ID from the enum identifier name. It seems possible that these unique IDs end up in the config file, so they become a compatibility contract in the future. This makes changing the names of these enums, like several commits in this branch do, very hard or impossible.

The current contract is not to break custom shortcuts, in the future we'll add toolbar customization. So yes they (unique ids) will immediately go in the settings file and so we'd like to establish something stable in this PR. Maybe we should add some loud comments to that file that any changes to enum names (not groupings) will break user settings.

> On the other hand, it is not obvious that the names of these enums need to be unique and fulfill this contract, and someone may change them without noticing. I suggest separating the unique identifier from the enum names in code for long-term stability.

I think they need to be unique, since we will only have this to determine what action goes with a keybind or toolbar button. So the separation is not useful, but we could use an integer (enum) instead of a readable string. The main idea with the string id is to make debugging or editing the settings file itself simpler and it fits into the design that we already have (arrays indexed by enums).

> 2. How are third-party plugin action IDs validated and namespaced? The plugin interface does not currently require actions to provide a stable, unique objectName(). Will this break existing third-party plugins?

I don't know of any 3rd-party plugins but yes it might be a problem as you are correct the plugin system doesn't have stable unique ids by design. Currently the built-in plugins are validated with the unit test.

> How do we prevent accidental ID collision between plugins and between nomacs and a plugin?

Ideally, plugins that want to bind shortcuts have to call the plugin API to create their actions. For the moment plugins have to promise to use the convention "plugin_[plugin-name]_[action-name]".

leejuyuu:
> Maybe we should add some loud comments to that file that any changes to enum names (not groupings) will break user settings.

I fear this would not work as intended. The enum identifier would be used in multiple places, and LSPs nowadays make rename fairely easy.

Using string literals in a centralized way has the benefits:

- They appear only once, so it is harder to change and easier to spot during review.
- When placed closely, e.g. in one file, it is easier to find the comment and why these are part of a contract.
- Being explicit instead of implicit.
- We might have better ways to enforce that they are unique.

Some examples:

- Gwenview: https://invent.kde.org/graphics/gwenview/-/blob/5b4ed19a16da41ddb3aac73f11b24fe1e72da663/app/mainwindow.cpp#L387-408
- qView: https://github.com/jurplel/qView/blob/c5eca1c7176549e0f0718d11201547ddcdb1f8c9/src/actionmanager.cpp#L638-L654
- inkscape: https://gitlab.com/inkscape/inkscape/-/blob/d40156b03232e3f7363d31a3a2f0df29f995a1f1/src/actions/actions-paths.cpp#L277-300


> it might be a problem as you are correct the plugin system doesn't have stable unique ids by design.

I suggest trying a way to make existing plugins continue to work first; however, produce a message/dialog that warns the plugin needs an update. We could announce old plugins are compatible for a number of versions, then schedule the compatibility removal after that. If this is not feasible, then we need to communicate clearly that things may break when this eventually lands.

ImageLounge/tests/DkShortcutsModel_test.cpp
scrubbbbs: Each test runs as a separate process so no interference is possible. libnomacscore contains all of nomacs so there aren't any extra dependencies here. I will add a check that initApplication() is not called more than once per process since gtest has a mode where that could happen.

ImageLounge/src/DkCore/DkPluginManager.cpp
scrubbbbs: I just confirmed that custom plugin shortcuts are already correctly migrated

scrubbbbs:
> > it might be a problem as you are correct the plugin system doesn't have stable unique ids by design.

This isn't correct, plugins actually do have a stable GUID-style serial number we can use for binding instead of the plugin menu name which might be translated in the future. I added this feature.
 
> I suggest trying a way to make existing plugins continue to work first; 

I don't think there are any existing plugins practically. There are some [here](https://github.com/diemmarkus/nomacs-plugins) that Markus created but they no longer compile against current nomacs. If there are any, they will not compile anymore since I modified for `DkPluginInterface::createActions()` to formally assure unique action ids for plugins.

There are a few distros that still package nomacs-plugins separately which will break - but probably broken already before this work. We may have to inform the maintainers that nomacs-plugins repo is deprecated. I added a hint in case they end up reading the code.

> however, produce a message/dialog that warns the plugin needs an update.We could announce old plugins are compatible for a number of versions, then schedule the compatibility removal after that. If this is not feasible, then we need to communicate clearly that things may break when this eventually lands.

Handling depredations gracefully in plugins is a nice idea for web browsers etc but I think it is overkill given the unstable nature of nomacs plugins and practically non-existent interest at the moment.
- API/ABI of nomacscore changes constantly due to refactoring
- Semantic changes in addition to the API surface (example, this patch before the latest commits)
- Too painful to think about ABI stability with every change and go update the necessary files (version strings in DkPluginInterface etc)

ImageLounge/tests/DkShortcutsModel_test.cpp
leejuyuu: I mean this test requires files other than the built test binary, as in 

```

    // these are needed for nomacs to find its settings file, we use
    // different values from release to isolate from development environment
    QCoreApplication::setOrganizationName("nomacs-test");
    QCoreApplication::setOrganizationDomain("https://test.nomacs.org");
    QCoreApplication::setApplicationName("ImageLounge-test");
    QCoreApplication::setApplicationVersion(NOMACS_VERSION_STR);
}
```

This is the special setup, which makes this test more similar to an integration test. Separating it into another directory makes the original unit tests stand-alone.

ImageLounge/src/DkCore/DkActionManager.h
```diff
@@ -81,87 +81,117 @@ public slots:
...
+     * @param text action label
+     * @param statusTip action hover text
+     *
+     * @warning The id should not be changed (normally), as it will be saved to
+     * config files. Any modification is considered a breaking change to the user
+     * config file.
+     */
+    QAction *add(T idx, const char *id, const QIcon &icon, const QString &text, const QString &statusTip)
+    {
+        Q_ASSERT(idx >= 0 && idx < mActions.size());
```
leejuyuu, line 109: This looks interesting. I suppose you still want the actions to be ordered by the enum value. When we miss a single `add()`, it seems we could get a nullptr somewhere.

leejuyuu:
The simple plugin no longer compiles. I suppose it is an example.

```
[77/109] Building CXX object nomacs-plugins/SI...ir/SimplePlugin_autogen/mocs_compilation.cpp.
FAILED: [code=1] nomacs-plugins/SIMPLE_PLUGIN/CMakeFiles/SimplePlugin.dir/SimplePlugin_autogen/mocs_compilation.cpp.o 
ccache /usr/bin/c++ -DDEBUG -DHAVE_EXIV2_HPP -DNOMACS_VERSION=\"3.23\" -DQT_CONCURRENT_LIB -DQT_CORE5COMPAT_LIB -DQT_CORE_LIB -DQT_DLL -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_NO_DEBUG -DQT_PLUGIN -DQT_PRINTSUPPORT_LIB -DQT_SHARED -DQT_SVG_LIB -DQT_WIDGETS_LIB -DSimplePlugin_EXPORTS -DWITH_LIBRAW -DWITH_LIBTIFF -DWITH_OPENCV -DWITH_PLUGINS -DWITH_QUAZIP -I/home/tzu-yu/git_repos/nomacs/build/nomacs-plugins/SIMPLE_PLUGIN/SimplePlugin_autogen/include -I/home/tzu-yu/git_repos/nomacs/build -I/home/tzu-yu/git_repos/nomacs/ImageLounge/src -I/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore -I/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkGui -I/home/tzu-yu/git_repos/nomacs/ImageLounge/3rdparty/libqpsd -I/home/tzu-yu/git_repos/nomacs/ImageLounge/3rdparty/drif -I/home/tzu-yu/git_repos/nomacs/build/nomacs-plugins/SIMPLE_PLUGIN -isystem /usr/include/opencv5 -isystem /usr/include/QuaZip-Qt6-1.7.2 -isystem /usr/include/QuaZip-Qt6-1.7.2/quazip -isystem /usr/include/qt6/QtCore -isystem /usr/include/qt6 -isystem /usr/lib/qt6/mkspecs/linux-g++ -isystem /usr/include/qt6/QtCore5Compat -isystem /usr/include/qt6/QtWidgets -isystem /usr/include/qt6/QtGui -isystem /usr/include/qt6/QtNetwork -isystem /usr/include/qt6/QtPrintSupport -isystem /usr/include/qt6/QtConcurrent -isystem /usr/include/qt6/QtSvg -Wall -Wextra -Wpedantic -Wformat -Wformat-nonliteral -Wshadow -g -std=gnu++17 -fPIC -mno-direct-extern-access -MD -MT nomacs-plugins/SIMPLE_PLUGIN/CMakeFiles/SimplePlugin.dir/SimplePlugin_autogen/mocs_compilation.cpp.o -MF nomacs-plugins/SIMPLE_PLUGIN/CMakeFiles/SimplePlugin.dir/SimplePlugin_autogen/mocs_compilation.cpp.o.d -o nomacs-plugins/SIMPLE_PLUGIN/CMakeFiles/SimplePlugin.dir/SimplePlugin_autogen/mocs_compilation.cpp.o -c /home/tzu-yu/git_repos/nomacs/build/nomacs-plugins/SIMPLE_PLUGIN/SimplePlugin_autogen/mocs_compilation.cpp
In file included from /home/tzu-yu/git_repos/nomacs/build/nomacs-plugins/SIMPLE_PLUGIN/SimplePlugin_autogen/UVLADIE3JM/moc_PLUGIN_CLASS_NAME.cpp:9,
                 from /home/tzu-yu/git_repos/nomacs/build/nomacs-plugins/SIMPLE_PLUGIN/SimplePlugin_autogen/mocs_compilation.cpp:2:
/home/tzu-yu/git_repos/nomacs/build/nomacs-plugins/SIMPLE_PLUGIN/SimplePlugin_autogen/UVLADIE3JM/../../../../../ImageLounge/plugins/SIMPLE_PLUGIN/src/PLUGIN_CLASS_NAME.h:44:22: error: ‘QList<QAction*> nmc::PLUGIN_CLASS_NAME::createActions(QWidget*)’ marked ‘override’, but does not override
   44 |     QList<QAction *> createActions(QWidget *parent) override;
      |                      ^~~~~~~~~~~~~
In file included from /home/tzu-yu/git_repos/nomacs/build/nomacs-plugins/SIMPLE_PLUGIN/SimplePlugin_autogen/UVLADIE3JM/../../../../../ImageLounge/plugins/SIMPLE_PLUGIN/src/PLUGIN_CLASS_NAME.h:27:
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkPluginInterface.h:74:30: warning: ‘virtual QList<QAction*> nmc::DkPluginInterface::createActions(const QString&, QWidget*)’ was hidden [-Woverloaded-virtual=]
   74 |     virtual QList<QAction *> createActions(const QString &idPrefix, QWidget *)
      |                              ^~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/build/nomacs-plugins/SIMPLE_PLUGIN/SimplePlugin_autogen/UVLADIE3JM/../../../../../ImageLounge/plugins/SIMPLE_PLUGIN/src/PLUGIN_CLASS_NAME.h:44:22: note:   by ‘QList<QAction*> nmc::PLUGIN_CLASS_NAME::createActions(QWidget*)’
   44 |     QList<QAction *> createActions(QWidget *parent) override;
      |                      ^~~~~~~~~~~~~
[78/109] Building CXX object nomacs-plugins/SI...s/SimplePlugin.dir/src/PLUGIN_CLASS_NAME.cpp.
FAILED: [code=1] nomacs-plugins/SIMPLE_PLUGIN/CMakeFiles/SimplePlugin.dir/src/PLUGIN_CLASS_NAME.cpp.o 
ccache /usr/bin/c++ -DDEBUG -DHAVE_EXIV2_HPP -DNOMACS_VERSION=\"3.23\" -DQT_CONCURRENT_LIB -DQT_CORE5COMPAT_LIB -DQT_CORE_LIB -DQT_DLL -DQT_GUI_LIB -DQT_NETWORK_LIB -DQT_NO_DEBUG -DQT_PLUGIN -DQT_PRINTSUPPORT_LIB -DQT_SHARED -DQT_SVG_LIB -DQT_WIDGETS_LIB -DSimplePlugin_EXPORTS -DWITH_LIBRAW -DWITH_LIBTIFF -DWITH_OPENCV -DWITH_PLUGINS -DWITH_QUAZIP -I/home/tzu-yu/git_repos/nomacs/build/nomacs-plugins/SIMPLE_PLUGIN/SimplePlugin_autogen/include -I/home/tzu-yu/git_repos/nomacs/build -I/home/tzu-yu/git_repos/nomacs/ImageLounge/src -I/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore -I/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkGui -I/home/tzu-yu/git_repos/nomacs/ImageLounge/3rdparty/libqpsd -I/home/tzu-yu/git_repos/nomacs/ImageLounge/3rdparty/drif -I/home/tzu-yu/git_repos/nomacs/build/nomacs-plugins/SIMPLE_PLUGIN -isystem /usr/include/opencv5 -isystem /usr/include/QuaZip-Qt6-1.7.2 -isystem /usr/include/QuaZip-Qt6-1.7.2/quazip -isystem /usr/include/qt6/QtCore -isystem /usr/include/qt6 -isystem /usr/lib/qt6/mkspecs/linux-g++ -isystem /usr/include/qt6/QtCore5Compat -isystem /usr/include/qt6/QtWidgets -isystem /usr/include/qt6/QtGui -isystem /usr/include/qt6/QtNetwork -isystem /usr/include/qt6/QtPrintSupport -isystem /usr/include/qt6/QtConcurrent -isystem /usr/include/qt6/QtSvg -Wall -Wextra -Wpedantic -Wformat -Wformat-nonliteral -Wshadow -g -std=gnu++17 -fPIC -mno-direct-extern-access -MD -MT nomacs-plugins/SIMPLE_PLUGIN/CMakeFiles/SimplePlugin.dir/src/PLUGIN_CLASS_NAME.cpp.o -MF nomacs-plugins/SIMPLE_PLUGIN/CMakeFiles/SimplePlugin.dir/src/PLUGIN_CLASS_NAME.cpp.o.d -o nomacs-plugins/SIMPLE_PLUGIN/CMakeFiles/SimplePlugin.dir/src/PLUGIN_CLASS_NAME.cpp.o -c /home/tzu-yu/git_repos/nomacs/ImageLounge/plugins/SIMPLE_PLUGIN/src/PLUGIN_CLASS_NAME.cpp
In file included from /home/tzu-yu/git_repos/nomacs/ImageLounge/plugins/SIMPLE_PLUGIN/src/PLUGIN_CLASS_NAME.cpp:25:
/home/tzu-yu/git_repos/nomacs/ImageLounge/plugins/SIMPLE_PLUGIN/src/PLUGIN_CLASS_NAME.h:44:22: error: ‘QList<QAction*> nmc::PLUGIN_CLASS_NAME::createActions(QWidget*)’ marked ‘override’, but does not override
   44 |     QList<QAction *> createActions(QWidget *parent) override;
      |                      ^~~~~~~~~~~~~
In file included from /home/tzu-yu/git_repos/nomacs/ImageLounge/plugins/SIMPLE_PLUGIN/src/PLUGIN_CLASS_NAME.h:27:
/home/tzu-yu/git_repos/nomacs/ImageLounge/src/DkCore/DkPluginInterface.h:74:30: warning: ‘virtual QList<QAction*> nmc::DkPluginInterface::createActions(const QString&, QWidget*)’ was hidden [-Woverloaded-virtual=]
   74 |     virtual QList<QAction *> createActions(const QString &idPrefix, QWidget *)
      |                              ^~~~~~~~~~~~~
/home/tzu-yu/git_repos/nomacs/ImageLounge/plugins/SIMPLE_PLUGIN/src/PLUGIN_CLASS_NAME.h:44:22: note:   by ‘QList<QAction*> nmc::PLUGIN_CLASS_NAME::createActions(QWidget*)’
   44 |     QList<QAction *> createActions(QWidget *parent) override;
      |                      ^~~~~~~~~~~~~
[90/109] Building CXX object nomacs-plugins/Af...lugin.dir/src/DkImgTransformationsPlugin.cpp.
ninja: build stopped: subcommand failed.
```

ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -1707,58 +1707,52 @@ void DkNoMacs::openFileWith(QAction *action)
...
     if (appPath == DkAppManager::kOpenFileAppName) {
-        if (fileInfo.isFromZip()) {
-            fileInfo = extractToTmpFile(fileInfo);
-        }
-        if (fileInfo.exists()) {
-            QUrl url = QUrl::fromLocalFile(fileInfo.path());
-            QDesktopServices::openUrl(url);
-        }
+        QUrl url = QUrl::fromLocalFile(fileInfo.path());
+        QDesktopServices::openUrl(url);
```
leejuyuu, line 1735: Originally, there seems to be a `fileInfo.exists()` check after `extractToTmpFile(fileInfo)`. But it has been removed. Is this intended?

ImageLounge/src/DkCore/DkPluginManager.cpp
leejuyuu: Could you explain? It looks like this field is no longer read.

leejuyuu:
Sorry for the late response. There are a few questions left.

About the plugins, I think it might be better to deprecate them in the future and move the official ones into the app. They tie too deeply into internal stuff and make changing them difficult.

ImageLounge/tests/DkShortcutsModel_test.cpp
scrubbbbs: It isn't a external dependency but rather a small scaffold for a temporary config file, it doesn't need to exist but it could interact with the test I suppose. On Linux the path would be `~/.config/nomacs-test/ImageLounage-test.ini`.

I suppose it should be removed when the test starts and cleaned up when the test finishes so it can't possibly interact with other tests, something like this unless there is a better way provided by google-test.

```c++
static void deleteConfig() {
   nmc::DefaultSettings settings;
   QFile::remove(settings.filePath())
}

static void initApplication() {
    // [...]
    QCoreApplication::setApplicationName()
    // [...]
    deleteConfig();
    std::set_terminate(deleteConfig());
}

```

ImageLounge/src/DkCore/DkPluginManager.cpp
scrubbbbs: I renamed "CustomPluginShortcuts" to "PluginActions". There were never shortcuts stored there, rather it is a registry for plugin action names/ids. The actual plugin shortcuts were always stored under "CustomShortcuts" key along with all other shortcuts.

The function of the setting remains the same which is to enable shortcut actions while also lazy-loading plugins. So at startup, dummy actions can be bound to shortcuts before the plugin is loaded. The dummy then loads the plugin and calls the actual action.

The reason it is fine to delete the key is because it is regenerated every time nomacs starts with the current set of plugins; previous to this patch it would keep the old information permanently which was a problem as I refactored the plugin action ids.

ImageLounge/src/DkCore/DkActionManager.h
scrubbbbs: I had the same thought but at the moment it is impossible to miss this, for example the shortcuts binding loop at startup would immediately segfault as it touches every action. If that logic ever changed then the shortcuts prefs dialog would segfault.

ImageLounge/src/DkGui/DkNoMacs.cpp
scrubbbbs: Yes, if the file no longer exists we'd like to see something fail rather than fail silently. Generally it can't hurt to open a file path that doesn't exist and the application will show an error dialog. But in practice this test never fails because if the file doesn't exist the UI refreshes and makes it impossible in nomacs to invoke the open-with action on it.

scrubbbbs:
> Sorry for the late response. There are a few questions left.
> 
> About the plugins, I think it might be better to deprecate them in the future and move the official ones into the app. They tie too deeply into internal stuff and make changing them difficult.

They are no longer separate git repo so there isn't much stopping it. It would be nice to freeze that repo, @novomesk  do we have the authority?

I don't see much interest in C++ development so future plugin interface if any would need to be in some scripting language to see any use. And with some abstraction layer so refactoring won't break them constantly.

ImageLounge/tests/DkShortcutsModel_test.cpp
leejuyuu: It would be easier to control what test are run if these type of tests are separated from the tests that can run without the full configs.

ImageLounge/src/DkCore/DkActionManager.h
leejuyuu: Yeah. I do not see a nicer way to handle this. Let's leave it as is for now.

ImageLounge/tests/DkShortcutsModel_test.cpp
scrubbbbs: I'm not seeing any difference this makes, besides slowing down the build and adding more cmake configuration to maintain. Can you elaborate?

As long as all tests run with `make check` and cannot interact with each other, does it really matter one way or the other?

scrubbbbs: I did a little experiment to separate the test environment/harness out a bit using fixtures, see my latest commit.

The base fixture makes an empty dir for each test and sets environment variables TMPDIR, TMP, XDG_CONFIG_DIR so each test has its own sandbox to capture nomacs config file or any temp files written like thumbnail cache. There are subclasses for tests needing additional setup that are basic enough to be reused elsewhere, this will cut down on refactoring chores should we decide refactor the classes involved.

At the end of a test run we can upload any files written from all tests using a common path (/tmp/nomacs-test-artifacts), which will help debug failing tests in CI.

Is this more in-line with what you are thinking of?

leejuyuu: Sorry that I did not elaborate. I was thinking about separating this test into another test executable, for example, a core_test and a gui_test or integration_test. The comment that the QApplication has a process-level restriction further supports this. 

I would also expect this test to clean up its temporary files. Maybe adding a unique/random component in the path so that the chance of a previous state carrying over to the next run is lower.

scrubbbbs: Did you take a look at the fixture? It seems to take care of the separation and state pollution risk.

As I said each TEST() function already runs in its own process which is a good thing. Otherwise global variables (in nomacs or Qt/dependencies), temporary files, non-reentrant classes, etc will cause unreliable results. So in reality we already have a process-level restriction for all tests, so this isn't changing anything.

What we are testing up to this point doesn't seem to do these things but we have hardly begun testing.

Adding another cmake target/exe makes no difference to test reliability or future development headaches because we had no isolation either way. Maybe causes more headaches because one has to go through each exe separately when testing in an IDE (see capture):

https://github.com/user-attachments/assets/9b854d9c-9776-4630-8d31-a12e5895b369

leejuyuu: > Did you take a look at the fixture? It seems to take care of the separation and state pollution risk.

Sorry, I missed that line.

> TEST() function already runs in its own process

This is only true when running with the `check` target or using ctest with some filtering. Executing the test binary directly makes it single-process.

What I am trying to say is that we should separate "the tests that are self-contained with their executable and the in-tree test data" and "the tests that also depend on the environment." The former is what we probably should focus on at this moment, as they are easier to manage.

I am not familiar with the IDE you are using. Shouldn't they handle the differences when you do something like click the test case to run?

scrubbbbs: > > TEST() function already runs in its own process
> 
> This is only true when running with the `check` target or using ctest with some filtering. Executing the test binary directly makes it single-process.

Yes I was mistaken, but no filtering needed, e.g. `ctest <build-dir>` is sufficient so there are no hoops to jump through.

> What I am trying to say is that we should separate "the tests that are self-contained with their executable and the in-tree test data" and "the tests that also depend on the environment." The former is what we probably should focus on at this moment, as they are easier to manage.

> I am not familiar with the IDE you are using.
> Shouldn't they handle the differences when you do something like click the test case to run?

QtCreator has a unit tests thing but it doesn't recompile the test exe before running, so I ended up switching to the core_tests target (in my screenshot) and running that way (with no process isolation). Maybe that could be fixed.

I guess I don't see the management issue provided the tests are well written. But what if we structured it this way:
- `core_tests` => `reentrant_tests`  -- each TEST() can run safely in a separate thread
- DkShortcutsModel_test => `forked_tests` -- each TEST() must run in a separate process

leejuyuu: I think separating the tests by what is being tested (self-contained units such as individual functions or classes v.s. the whole application and the environment it depends on) is a more common choice.

scrubbbbs: This is the direction I'm going. I decided to go with "pure" tests vs "serial" tests for now. The pure tests are not run with process isolation. I found  way to do the environment variables so ctest works properly when run separate. This might mean we can test on msvc builds now.

```cmake
include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src/DkCore)

include(GoogleTest)

# dependencies used for all tests
add_library(test_libs INTERFACE)
target_link_libraries(
    test_libs #
    INTERFACE #
        ${DLL_CORE_NAME}
        ${OpenCV_LIBS}
        GTest::gtest_main
        Qt${QT_VERSION_MAJOR}::Core
        Qt${QT_VERSION_MAJOR}::Gui
)

add_custom_target(tests)

# pure tests
# test suites where *all* gtest TEST() may run in the same process
# the key property is they cannot modify global shared state, but reading is fine
add_executable(
    tests_pure #
    DkUtils_test.cpp
    DkBaseViewPort_test.cpp
    DkNativeImage_test.cpp
    DkMetaData_test.cpp
)
target_link_libraries(tests_pure test_libs)
add_dependencies(tests tests_pure)

# register the entire test binary with ctest (no process isolation)
# if a test segfaults it skips the rest of the tests
add_test(NAME tests_pure COMMAND tests_pure)

# serial tests
# contains suites where *any* gtest TEST() could modify global shared state
# but they should still be cleaning up or using fixtures to avoid side effects
add_executable(
    tests_serial #
    DkShortcutsModel_test.cpp
)
target_link_libraries(tests_serial test_libs)
add_dependencies(tests tests_serial)

# register each TEST() as a separate command line in ctest configuration (process isolation)
gtest_discover_tests(tests_serial)

# set environment variables so `ctest` is run against this build's products
file(
    WRITE
    "${CMAKE_CURRENT_BINARY_DIR}/GlobalCTestEnv.cmake"
    "# environment overrides for ctest
set(ENV{NOMACS_TESTING} 1)
set(ENV{QT_PLUGIN_PATH} \"${CMAKE_BINARY_DIR}/nomacs-plugins:\$ENV{QT_PLUGIN_PATH}\")
set(ENV{LD_LIBRARY_PATH} \"${CMAKE_BINARY_DIR}:\$ENV{LD_LIBRARY_PATH}\")
set(ENV{DYLD_LIBRARY_PATH} \"${CMAKE_BINARY_DIR}:\$ENV{DYLD_LIBRARY_PATH}\")
"
)
set_property(DIRECTORY PROPERTY TEST_INCLUDE_FILE "${CMAKE_CURRENT_BINARY_DIR}/GlobalCTestEnv.cmake")

# convenience script for running test binaries directly (gdb etc)
file(
    WRITE
    "${CMAKE_BINARY_DIR}/test.env"
    "# source this file when running nomacs test binaries directly (not required for ctest)
set -a
export NOMACS_TESTING=1
export QT_PLUGIN_PATH=\"${CMAKE_BINARY_DIR}/nomacs-plugins:\$QT_PLUGIN_PATH\"
export LD_LIBRARY_PATH=\"${CMAKE_BINARY_DIR}:\$LD_LIBRARY_PATH\"
export DYLD_LIBRARY_PATH=\"${CMAKE_BINARY_DIR}:\$DYLD_LIBRARY_PATH\"
"
)

# cmake's "make test" target does not have any dependencies, add "make check" for that
add_custom_target(
    check #
    COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure
    DEPENDS tests
    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
)


```

leejuyuu: I think this distinction fails to capture what helps to understand the test. That some tests needs to be run serially is one of the symptoms, not a cause. The cause is that these tests are testing different things at different levels.

scrubbbbs: is "tests_unit" vs "tests_integration" what you are looking for?

scrubbbbs: what about dropping tests_integration entirely, and we'll just have separate test binaries for anything that modifies shared state. So there will be a lot of them but the link is to nomacs core so maybe it doesn't slow us down too much

leejuyuu:
````diff
diff --git a/ImageLounge/plugins/CMakeLists.txt b/ImageLounge/plugins/CMakeLists.txt
index a9e6ca47..982fb199 100644
--- a/ImageLounge/plugins/CMakeLists.txt
+++ b/ImageLounge/plugins/CMakeLists.txt
@@ -20,6 +20,8 @@ option(ENABLE_PAGE "Compile Document Page Extraction plugin" ON)
 option(ENABLE_COMPOSITE "Compile Composite Image plugin" ON)
 option(ENABLE_SIMPLE "Compile SIMPLE PLUGIN" OFF)
 
+add_custom_target(nomacs_plugins)
+
 NMC_PREPARE_PLUGIN()
 NMC_FINDQT()
 NMC_FIND_OPENCV("core" "imgproc")
@@ -28,24 +30,30 @@ set(BUILDING_MULTIPLE_PLUGINS true)
 
 if(ENABLE_FAKE_MINIATURES)
     add_subdirectory(FakeMiniaturesPlugin)
+    add_dependencies(nomacs_plugins fakeMiniaturesPlugin)
 endif(ENABLE_FAKE_MINIATURES)
 
 if(ENABLE_TRANSFORM)
     add_subdirectory(AffineTransformations)
+    add_dependencies(nomacs_plugins affineTransformPlugin)
 endif(ENABLE_TRANSFORM)
 
 if(ENABLE_PAINT)
     add_subdirectory(PaintPlugin)
+    add_dependencies(nomacs_plugins paintPlugin)
 endif(ENABLE_PAINT)
 
 if(ENABLE_PAGE)
     add_subdirectory(PageExtractionPlugin)
+    add_dependencies(nomacs_plugins pageExtractionPlugin)
 endif(ENABLE_PAGE)
 
 if(ENABLE_COMPOSITE)
     add_subdirectory(CompositePlugin)
+    add_dependencies(nomacs_plugins CompositePlugin)
 endif(ENABLE_COMPOSITE)
 
 if(ENABLE_SIMPLE)
     add_subdirectory(SIMPLE_PLUGIN)
+    add_dependencies(nomacs_plugins SimplePlugin)
 endif(ENABLE_SIMPLE)
diff --git a/ImageLounge/tests/CMakeLists.txt b/ImageLounge/tests/CMakeLists.txt
index f6cb9454..d1169bb0 100644
--- a/ImageLounge/tests/CMakeLists.txt
+++ b/ImageLounge/tests/CMakeLists.txt
@@ -14,55 +14,44 @@ target_link_libraries(
         Qt${QT_VERSION_MAJOR}::Gui
 )
 
+add_custom_target(tests_unit)
+add_custom_target(tests_integration)
 add_custom_target(tests)
+add_dependencies(tests tests_unit tests_integration)
 
-function(add_ctest_test name)
+function(add_nomacs_test name category)
     add_executable(${name} ${ARGN})
-    target_link_libraries(${name} test_libs)
-    add_test(NAME ${name} COMMAND ${name})
-    add_dependencies(tests ${name})
+    target_link_libraries(${name} PRIVATE test_libs)
+    gtest_discover_tests(${name} PROPERTIES LABELS ${category})
+    add_dependencies(tests_${category} ${name})
+    set_property(DIRECTORY PROPERTY TEST_INCLUDE_FILE "${NOMACS_TEST_ENV_FILE}")
 endfunction()
 
-# tests with no side-effects can all go in the same binary
-add_ctest_test(
-    tests_core #
-    DkUtils_test.cpp
-    DkBaseViewPort_test.cpp
-    DkNativeImage_test.cpp
-    DkMetaData_test.cpp
-)
-
-add_ctest_test(
-    test_shortcutsmodel
-    DkShortcutsModel_test.cpp
-)
-
 # set environment variables so `ctest` is run against this build's products
+set(NOMACS_TEST_ENV_FILE "${CMAKE_CURRENT_BINARY_DIR}/GlobalCTestEnv.cmake")
 file(
-    WRITE
-    "${CMAKE_CURRENT_BINARY_DIR}/GlobalCTestEnv.cmake"
+    WRITE "${NOMACS_TEST_ENV_FILE}"
     "# environment overrides for ctest
-set(ENV{NOMACS_TESTING} 1)
-set(ENV{QT_PLUGIN_PATH} \"${CMAKE_BINARY_DIR}/nomacs-plugins:\$ENV{QT_PLUGIN_PATH}\")
+set(ENV{QT_PLUGIN_PATH} \"${CMAKE_BINARY_DIR}/nomacs-plugins\")
 set(ENV{LD_LIBRARY_PATH} \"${CMAKE_BINARY_DIR}:\$ENV{LD_LIBRARY_PATH}\")
 set(ENV{DYLD_LIBRARY_PATH} \"${CMAKE_BINARY_DIR}:\$ENV{DYLD_LIBRARY_PATH}\")
 "
 )
-set_property(DIRECTORY PROPERTY TEST_INCLUDE_FILE "${CMAKE_CURRENT_BINARY_DIR}/GlobalCTestEnv.cmake")
 
 # convenience script for running test binaries directly (gdb etc)
 file(
-    WRITE
-    "${CMAKE_BINARY_DIR}/test.env"
+    WRITE "${CMAKE_BINARY_DIR}/test.env"
     "# source this file when running nomacs test binaries directly (not required for ctest)
 set -a
-export NOMACS_TESTING=1
-export QT_PLUGIN_PATH=\"${CMAKE_BINARY_DIR}/nomacs-plugins:\$QT_PLUGIN_PATH\"
+export QT_PLUGIN_PATH=\"${CMAKE_BINARY_DIR}/nomacs-plugins\"
 export LD_LIBRARY_PATH=\"${CMAKE_BINARY_DIR}:\$LD_LIBRARY_PATH\"
 export DYLD_LIBRARY_PATH=\"${CMAKE_BINARY_DIR}:\$DYLD_LIBRARY_PATH\"
 "
 )
 
+add_subdirectory(unit)
+add_subdirectory(integration)
+
 # cmake's "make test" target does not have any dependencies, add "make check" for that
 add_custom_target(
     check #
@@ -70,3 +59,17 @@ add_custom_target(
     DEPENDS tests
     WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
 )
+
+add_custom_target(
+    check-unit
+    COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -L unit
+    DEPENDS tests_unit
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+)
+
+add_custom_target(
+    check-integration
+    COMMAND ${CMAKE_CTEST_COMMAND} --output-on-failure -L integration
+    DEPENDS tests_integration
+    WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
+)
diff --git a/ImageLounge/tests/integration/CMakeLists.txt b/ImageLounge/tests/integration/CMakeLists.txt
new file mode 100644
index 00000000..9ebac797
--- /dev/null
+++ b/ImageLounge/tests/integration/CMakeLists.txt
@@ -0,0 +1,11 @@
+# Tests that wish to have its own QApplication needs its own binary.
+# Add `add_nomacs_test` for each of them.
+
+add_nomacs_test(
+    test_shortcutsmodel
+    integration
+    DkShortcutsModel_test.cpp
+)
+if(TARGET nomacs_plugins)
+    add_dependencies(test_shortcutsmodel nomacs_plugins)
+endif()
diff --git a/ImageLounge/tests/DkShortcutsModel_test.cpp b/ImageLounge/tests/integration/DkShortcutsModel_test.cpp
similarity index 100%
rename from ImageLounge/tests/DkShortcutsModel_test.cpp
rename to ImageLounge/tests/integration/DkShortcutsModel_test.cpp
diff --git a/ImageLounge/tests/DkTestFixture.h b/ImageLounge/tests/integration/DkTestFixture.h
similarity index 100%
rename from ImageLounge/tests/DkTestFixture.h
rename to ImageLounge/tests/integration/DkTestFixture.h
diff --git a/ImageLounge/tests/unit/CMakeLists.txt b/ImageLounge/tests/unit/CMakeLists.txt
new file mode 100644
index 00000000..e2f29d9d
--- /dev/null
+++ b/ImageLounge/tests/unit/CMakeLists.txt
@@ -0,0 +1,10 @@
+# Tests with no side effects can share a binary.
+# Append new test files to the list.
+add_nomacs_test(
+    tests_core
+    unit
+    DkUtils_test.cpp
+    DkBaseViewPort_test.cpp
+    DkNativeImage_test.cpp
+    DkMetaData_test.cpp
+)
diff --git a/ImageLounge/tests/DkBaseViewPort_test.cpp b/ImageLounge/tests/unit/DkBaseViewPort_test.cpp
similarity index 100%
rename from ImageLounge/tests/DkBaseViewPort_test.cpp
rename to ImageLounge/tests/unit/DkBaseViewPort_test.cpp
diff --git a/ImageLounge/tests/DkMetaData_test.cpp b/ImageLounge/tests/unit/DkMetaData_test.cpp
similarity index 100%
rename from ImageLounge/tests/DkMetaData_test.cpp
rename to ImageLounge/tests/unit/DkMetaData_test.cpp
diff --git a/ImageLounge/tests/DkNativeImage_test.cpp b/ImageLounge/tests/unit/DkNativeImage_test.cpp
similarity index 100%
rename from ImageLounge/tests/DkNativeImage_test.cpp
rename to ImageLounge/tests/unit/DkNativeImage_test.cpp
diff --git a/ImageLounge/tests/DkUtils_test.cpp b/ImageLounge/tests/unit/DkUtils_test.cpp
similarity index 75%
rename from ImageLounge/tests/DkUtils_test.cpp
rename to ImageLounge/tests/unit/DkUtils_test.cpp
index 04841217..6e6ae654 100644
--- a/ImageLounge/tests/DkUtils_test.cpp
+++ b/ImageLounge/tests/unit/DkUtils_test.cpp
@@ -1,4 +1,4 @@
-#include "../src/DkCore/DkUtils.h"
+#include "DkUtils.h"
 #include "DkVersion.h"
 #include <gtest/gtest.h>
 #include <string>
@@ -8,17 +8,6 @@ TEST(TestEnvironment, Test)
     // If this test fails the test executable has loaded the wrong nomacsCore shared library
     // This can be caused by LD_LIBRARY_PATH containing another installation of nomacsCore.
     EXPECT_EQ(std::string(NOMACS_REVISION_STR), std::string(nmc::revisionString));
-
-    qInfo() << "NOMACS_TESTING: " << qgetenv("NOMACS_TESTING");
-    qInfo() << "QT_PLUGIN_PATH:" << qgetenv("QT_PLUGIN_PATH");
-    qInfo() << "LD_LIBRARY_PATH: " << qgetenv("LD_LIBRARY_PATH");
-    qInfo() << "DYLD_LIBRARY_PATH: " << qgetenv("DYLD_LIBRARY_PATH");
-
-    // If you are running test binary directly (not ctest) these must be set
-    EXPECT_NE(qgetenv("NOMACS_TESTING"), "") << "Test environment is invalid, NOMACS_TESTING is not set";
-    EXPECT_NE(qgetenv("QT_PLUGIN_PATH"), "")
-        << "Test environment is invalid, QT_PLUGIN_PATH is not set\n\n"
-           "Run 'source test.env' in the build directory before running tests directly";
 }
 
 TEST(DkFileNameConverterTest, Test)
diff --git a/README.md b/README.md
index 50d67c33..e1f26e84 100644
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ There are other optional features that can be enabled during build:
 | `ENABLE_QUAZIP`    | Quazip**       |                 | Enable reading from zip files
 | `ENABLE_OPENCV`    | OpenCV         |                 | RAW, TIFF, Adjustments, High-quality thumbnails, DRIF files, histogram, mosaic
 | `ENABLE_PLUGINS`   | nomacs/plugins | `ENABLE_OPENCV` | Paint on image, Composite, Affine Transform, Fake Miniatures, Page Extraction
-| `ENABLE_TESTING`   | GoogleTest     |                 | Enables `make check` target for unit testing
+| `ENABLE_TESTING`   | GoogleTest     |                 | Enables targets for unit and integration testing
 
 (**) Quazip is not enabled by default as of nomacs 3.19.1
 
@@ -185,6 +185,17 @@ Nomacs uses [GoogleTest](https://github.com/google/googletest), which is include
 make check
 ```
 
+Unit and integration tests can also be built and run separately:
+
+```bash
+make check-unit
+make check-integration
+
+# Alternatively, invoke with ctest
+ctest -L unit
+ctest -L integration
+```
+
 ### Install nomacs
 
 By default nomacs is installed to /usr/local/ unless you set `CMAKE_INSTALL_PREFIX`

````

leejuyuu: > is "tests_unit" vs "tests_integration" what you are looking for?

Yes, or core vs app tests.

The recent changes have several issues:

1. It masks the test counts discoverable by ctest to 2.
2. The new lines in TestEnvironment are unnecessary and make running the test binary directly fail.
3. QT_PLUGIN_PATH was overwrite, it became append.

Issue #1613, opened 2026-06-25
New build scripts for Windows


novomesk:
FYI,
I am preparing new scripts how to build nomacs for Windows.

Updating dependencies with the current scripts is often challenge.

Kimageformsts will be used on Windows too instead of
qt-jpegxl-image-plugin, qt-avif-image-plugin and similar.

Another problem is 1 hour timeout in AppVeyor. Dependencies will be prepared locally and only nomacs sources will be built on AppVeyor.

scrubbbbs:
Will there be scripts to prepare all dependencies (vc-pkg etc)?

I have a [prototype cross-compile setup](https://github.com/scrubbbbs/cbird/tree/main/docker) for github actions we could adapt for this instead of visual studio. Downside seems to be older gcc version, mingw-gcc seems to be lagging behind quite a bit. Upside is easy to test and work on locally since it just uses docker.

novomesk:
I already have the script to prepare all dependencies.

scrubbbbs:
Thanks for your work on this, I was able to get this going. I had only one issue, the `make-all-dependencies.bat` seems to be out of date so I ended up manually building 11 through 26. I would consider dropping the numbered files and instead keeping a list in make-all-dependencies with a loop to build each one in order. That still gives us the option to disable any dependency for testing.

novomesk:
It is out of date. I am going to update it after finishing work in kimageformats.

Then I'd like to see a new version.

Windows users needs updated libraries and Linux users OpenCV 5 compatibility.

novomesk:
@scrubbbbs There is a new release https://github.com/nomacs/nomacs/releases/tag/3.23.2

If you wish, you may add information and/or packages for other systems.

The changes were mostly for Windows anyway, those onther-systems packages would provide functionality similar to the previous 3.23.1 version.

scrubbbbs:
> If you wish, you may add information and/or packages for other systems.

I don't see any commits for other platforms but the build recipe changed so I'll add them.

scrubbbbs:
@novomesk  I'm looking to add a package to the windows build. This will be for single-instance function. Do you have any thoughts on how we go about this? Should I commit the bat file first so CI can pass until the feature using it can be merged sometime later?

novomesk:
Tell me more info. What package you would like to add?
I am still tuning Windows build, so it is a work in progress.

scrubbbbs:
[...]

PR #1616, opened 2026-07-03
New build scripts for Windows

novomesk:
https://github.com/nomacs/nomacs/issues/1613


novomesk:
Hi @scrubbbbs 

I made some several radical changes:

I remove submodules (LibRaw, exiv2, expat, imageformats, opencv, quazip). These dependencies should be built externally.

All of them (except opencv) was patched full of various workarounds, and it was difficult to update and to work together with full functionality.

As a consequence of the submodules removal, I removed also building of quazip from source (USE_SYSTEM_QUAZIP=OFF), to reduce complexity of cmake scripts and to reduce number of possible build configurations. Package maintainers don't like bundled dependencies anyway.

Another temporary consequence of the submodules removals is that macOS workflow doesn't build opencv from from source now, opencv build can be re-added in the near future inside the workflow.

In the future, I'd like to see that cmake scripts for various platforms are unified (not very different to each other) and easier to update/maintain.

AppVeyor now build only nomacs, depencencies are downloaded in binary form, it is faster. I'll add the scripts how I built the dependencies on Windows. The dependencies were build on my computer, as AppVeyor is slow to built everything in our hour.

Working on this helped me to understand how the things worked and I am convinced that further reforms are needed.

scrubbbbs:
I guess I'm ok with pulling the submodules we don't really need, long-term this is probably better. Will you be checking in the recipe for the windows dependencies (python script or at least something in manuals)

> Another temporary consequence of the submodules removals is that macOS workflow doesn't build opencv from from source now, opencv build can be re-added in the near future inside the workflow.

Most of the stuff deleted from macos.yml needs to be added back. Can you commit a more minimal changeset that comments out only the breaking lines (when config.opencv=false and submodule is missing)? I'll figure out how to add support back.

> In the future, I'd like to see that cmake scripts for various platforms are unified (not very different to each other) and easier to update/maintain.

It would be nice but I have doubts hare far we can get this way. We can always extract common functions, but then we find each target needs a slightly different version so in the end not much complexity wrt maintenance was removed. Maybe we need another build system that is more friendly towards this.

.github/workflows/macos.yml
```diff
@@ -74,54 +62,20 @@ jobs:
...
-          key: ${{ runner.os }}-opencv-${{ steps.get_info.outputs.OPENCV_KEY }}
-
-      - name: Build opencv
-        if: matrix.config.opencv && steps.cache_opencv.outputs.cache-hit != 'true'
-        run: |
-          cd ${GITHUB_WORKSPACE}
-          # put opencv in /opt to prevent conflicts with homebrew and/or apple silicon
-          # note kimageformats is installed into qt plugins directory so it won't be a problem,
-          ./scripts/build-opencv.sh -D CMAKE_INSTALL_PREFIX=/opt/opencv
-
```
scrubbbbs, line 119: This chunk (Line 104-119) can be left alone

.github/workflows/macos.yml
```diff
@@ -74,54 +62,20 @@ jobs:
...
           echo "Compiler: $COMPILER_ID"
-          echo "OpenCV: $OPENCV_ID"
-          echo "OpenCV Cache Key: $OPENCV_KEY"
-          echo "OPENCV_KEY=${OPENCV_KEY}" >> $GITHUB_OUTPUT
-
-      - name: Setup Cache
-        run: |
-          # we must install things to /opt for packaging script to work, but cache@v4 won't write there
-          sudo mkdir /opt/opencv
-          sudo chown $USER:$(id -gn) /opt/opencv
```
scrubbbbs, line 95: The removed lines can be moved into "Setup Cache" step with an "if config.opencv" check

.github/workflows/macos.yml
```diff
@@ -44,22 +44,10 @@ jobs:
...
-          - type: quazip
-            packages: libraw
-            flags: "-DENABLE_QUAZIP=ON -DUSE_SYSTEM_QUAZIP=OFF"
-            kimageformats: true # compile kimageformat-plugins
-            opencv: true # compile a minimal opencv
-            portable: true # use the new portable bundle script
-            bundle: false # build portable bundle using macdeployqt
-            artifacts: true # upload artifacts
-            release: false # TODO: publish release
-
```
scrubbbbs, line 57: Comment out this block instead

novomesk:
I was able to re-add opencv on macOS again.

scrubbbbs:
> I was able to re-add opencv on macOS again.

Nice, I think that means quazip is only thing missing now?

novomesk:
Yes, only quazip is missing on macOS.

Issue #1079, opened 2024-05-27
The scroll wheel will scroll multiple pictures at a time


flowerseacar:

scrubbbbs:
Thanks for reporting. I can't reproduce the bug on 3.17/ubuntu. What is your version?

I did notice the slide animation restarts when the next wheel event comes so it hiccups a bit but there it seems to slide past all pictures.

I assume you are using Xwindows, please use `xev` to see that events are reporting properly from the mouse/x server.

scrubbbbs:
Another idea, the "Skip Images" setting could affect this. Also, you can use the File Explorer panel to more easily see if something is skipped.

scrubbbbs:
Ensure imwheel daemon is not  running `ps  -aux | grep imwheel` and use `xev` to check if  the mouse or X server is sending bad events.

scrubbbbs:
You say pictures are skipped, did you check the "Skip Images" setting in preferences? This is exactly what it does if you scroll too fast.

scrubbbbs:
One click of the scrollwheel equals one image in nomacs, if you are getting something else it is a bug with the mouse hardware, mouse driver or other software like imwheel. I don't think this is going to change as it is generally expected behavior.

With mice that have non-clicky scroll wheels it is up to the system mouse settings/driver to determine the scroll rate (scroll clicks per second).

Closing as this is an unusual device/configuration issue and not something we need to handle in nomacs

Issue #630, opened 2021-04-06
Flipping through images in thumbnail view with mouse wheel shows some intermediate image for a split second


solarjoe:
**Describe the bug**
When I flip through the images with the mouse wheel hovering the thumbnails 
there is some intermediate image for a split second that also shows the zoom level
in percent in the lower left corner.

This looks sort of weird, almost if you skipped a image, but it is
actually the next one maybe shifted or resized.

When skipping through the images with the arrow keys it looks fine.

**Expected behavior**
Don't show a different view for a split second. :)

 - OS: Linux Mint
 - Version 3.8

solarjoe:
When I skip through the images with the mouse wheel the zoom level displayed at the lower left seems to keep ascending or descending with the next image, but resets at some point. The images get smaller or larger depending on the mouse wheel direction until it resets.

leejuyuu:
Hello @solarjoe, I do not understand nor am I able to reproduce the problem. Could you try a newer version? If the issue persists, a screen recording would greatly help us identify the problem.

solarjoe:
Sorry, this is three years old and I am no longer using Nomacs. Closing.

PR #1617, opened 2026-07-04
fix: skip-free image navigation with high-resolution mouse wheels

kn9ff:
## Problem

High-resolution mouse wheels (Logitech MX / G-series, many modern mice, some touchpads) report several small-delta wheel events per physical notch instead of a single event with `angleDelta() == ±120`.

`DkViewPort::wheelEvent()` navigates one image per event regardless of the delta magnitude, so a single notch skips several images. This makes wheel navigation unusable with these devices — even in the mouse's ratchet mode.

Reported before in #1079 (closed without a fix); related: #630.

## Fix

Accumulate `angleDelta()` in a per-viewport member and only navigate once a full wheel step (120 = 15°) is reached, as recommended by the Qt docs for handling hi-res scroll events. The accumulator resets on direction change so quick back-and-forth scrolling stays responsive.

Classic wheels are unaffected: they deliver ±120 per notch, which triggers navigation immediately, exactly as before.

## Testing

- Fedora (Wayland), Qt 6, Logitech G502 X Lightspeed (hi-res wheel):
  - before: 3–5 images per notch
  - after: exactly 1 image per notch, both directions, ratchet and free-spin
- Classic ±120 wheel: unchanged behavior (1 image per notch)
- Zoom path (`Ctrl+wheel` / "zoom on wheel") untouched — only the navigation branch changed
- Builds warning-free with clang-format applied

### DIFF
--- a/ImageLounge/src/DkGui/DkViewPort.cpp
+++ b/ImageLounge/src/DkGui/DkViewPort.cpp
@@ -1317,10 +1317,25 @@ void DkViewPort::wheelEvent(QWheelEvent *event)
         } else {
             delta = event->angleDelta().y();
         }
-        if (delta < 0)
+
+        // accumulate deltas and only navigate on a full wheel step (120):
+        // high-resolution wheels (e.g. Logitech MX/G-series, some touchpads) report
+        // several small-delta events per notch, which skipped multiple images (#1079)
+        constexpr int wheelStep = 120; // Qt reports wheel deltas in 1/8 degree; one notch = 15 degrees
+
+        // reset the accumulator when the scroll direction changes
+        if ((delta < 0 && mWheelAccumulator > 0) || (delta > 0 && mWheelAccumulator < 0))
+            mWheelAccumulator = 0;
+
+        mWheelAccumulator += delta;
+
+        if (mWheelAccumulator <= -wheelStep) {
             loadNextFileFast();
-        if (delta > 0)
+            mWheelAccumulator = 0;
+        } else if (mWheelAccumulator >= wheelStep) {
             loadPrevFileFast();
+            mWheelAccumulator = 0;
+        }
     } else
         DkBaseViewPort::wheelEvent(event);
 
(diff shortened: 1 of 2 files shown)


scrubbbbs:
If there is a mouse wheel with notches, and the angle delta reported < 15 degrees will this work? Or will the wheel have to click multiple times?

kn9ff:
Good question. In practice such a device shouldn't exist: notched wheels report 120 per detent by platform contract on both Windows (`WHEEL_DELTA`) and Linux (the kernel's `REL_WHEEL_HI_RES` protocol defines 120 units per detent, and hi-res sub-events are calibrated to sum to 120 per physical notch). Sub-120 events only come from hi-res wheels, where accumulating to 120 fires exactly once per notch.

That said, you made me notice the code dropped the remainder after navigating (reset to 0 instead of subtracting 120), which could stall with a driver whose sub-events round to slightly under 120 per notch (e.g. 3×39=117). Tested on my Logitech G502X and my older logitech G604. Pushed a follow-up commit that carries the remainder as the Qt docs recommend, so even those accumulate correctly across notches.

ImageLounge/src/DkGui/DkViewPort.cpp
```diff
@@ -1317,10 +1317,25 @@ void DkViewPort::wheelEvent(QWheelEvent *event)
...
-        if (delta < 0)
+
+        // accumulate deltas and only navigate on a full wheel step (120):
+        // high-resolution wheels (e.g. Logitech MX/G-series, some touchpads) report
+        // several small-delta events per notch, which skipped multiple images (#1079)
+        constexpr int wheelStep = 120; // Qt reports wheel deltas in 1/8 degree; one notch = 15 degrees
+
+        // reset the accumulator when the scroll direction changes
+        if ((delta < 0 && mWheelAccumulator > 0) || (delta > 0 && mWheelAccumulator < 0))
+            mWheelAccumulator = 0;
```
scrubbbbs, line 1328: style: braces around single-line if

scrubbbbs:
Thanks for this contribution. I tested on windows, macos, and linux/x11, for traditional mouse, seems to work fine. I am unable to test on hires mouse or touchpad at the moment.

I have some requests.

- "fix" commit subject(s) should say what bug was fixed, not how it was fixed (body)
- maybe more concise comments or commit message

scrubbbbs:
@kn9ff I'd like to merge this, can you make the requested changes?

PR #1620, opened 2026-07-06
Fixes for linux, mxe, and macos build, and macos workflow

scrubbbbs:
Add back quazip support on macos and fix a few minor issues I found elsewhere

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -44,12 +44,12 @@ jobs:
             cmakefiles: "qt6/lib/cmake" # root dir is runner.homebrew-opt
 
         config:
-          # this is extremely slow (~15 minutes) due to opencv
-          - type: opencv
+          - type: full
             packages: libraw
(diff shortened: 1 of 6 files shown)


novomesk:
I think we can completely delete `ImageLounge/3rdparty/win-binaries` folder.
`qjp2.dll` is an old plugin for Qt5 which is no longed used.

And I think `ImageLounge/3rdparty/libqpsd` is also good candidate for removal, isn't it enough to use **kimg_psd** from kimageformats?

scrubbbbs:
> I think we can completely delete `ImageLounge/3rdparty/win-binaries` folder. `qjp2.dll` is an old plugin for Qt5 which is no longed used.

Sounds reasonable.

> And I think `ImageLounge/3rdparty/libqpsd` is also good candidate for removal, isn't it enough to use **kimg_psd** from kimageformats?

I checked formats_testset and the only difference was libqpsd has a fallback for when there is no composite:

image

I don't think that is a big deal since we get an error logged for this from the plugin as well.

PR #1622, opened 2026-07-06
Fix build against opencv 5.0

### LINKED ISSUE #1619: Not compatible (build failed) with opencv 5.0.0
**Describe the bug**
Building the latest git on archlinux using opencv 5.0.0 failed.

**To Reproduce**
Using the build-from-source steps for linux, cmake configuration succeeds but build failed.

```
.../nomacs/ImageLounge/src/DkGui/DkDialog.cpp:95:10: fatal error: opencv2/imgproc/imgproc_c.h: No such file or directory
   95 | #include "opencv2/imgproc/imgproc_c.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Opencv version is updated to 5.0.0-1 on archlinux.

**Additional context**
Currently I have to use `cmake -DENABLE_RAW=OFF -DENABLE_TIFF=OFF -DENABLE_OPENCV=OFF -DENABLE_PLUGINS=OFF` to disable opencv features to pass the build.

This is OK for me since I don't use the features supported by opencv. But, I hope it will get fixed later.

jetm:
opencv 5.0 removed the legacy C API (`opencv2/imgproc/imgproc_c.h`) and the `CV_*` / `IPL_*` constant aliases it defined, and moved the shape/contour analysis functions and the `DistanceTypes` enum into the new `geometry` module. nomacs still uses those removed names, so it no longer builds against opencv 5 (#1619).

Following @scrubbbbs's suggestion in #1619 to move off the `_c` header — which on its own isn't enough, since the removed `CV_RGB2Lab` / `CV_INTER_AREA` constants still fail to compile (as @FabioLolix hit in the issue) — this does the full port:

- Rename the removed color-conversion / interpolation / border constants to `cv::COLOR_*`, `cv::INTER_*`, `cv::BORDER_*`. These have the same numeric values as the old `CV_*` aliases, so there is no behaviour change.
- Drop the `imgproc_c.h` include (replacing it with `imgproc.hpp` only in the one file where it was the sole imgproc header).
- Include `opencv2/geometry.hpp` behind `#if CV_MAJOR_VERSION >= 5` where the relocated shape functions and `cv::DIST_C` are used, so opencv 4 keeps building against `imgproc` unchanged.

Left as-is: matrix-type constants (`CV_8U`, `CV_32F`, ...), `CV_PI`, `CV_MAJOR_VERSION`, `cvRound` (all retained by opencv 5), and the `CV_MAJOR_VERSION <= 3` fallback branches.

Built cleanly against opencv 5.0.0 with Qt 6.11 (core plus all bundled plugins); opencv 4.x is unaffected since the `geometry.hpp` includes are version-guarded and the renamed enums exist there too.

Closes #1619

### DIFF
--- a/ImageLounge/plugins/AffineTransformations/src/DkSkewEstimator.cpp
+++ b/ImageLounge/plugins/AffineTransformations/src/DkSkewEstimator.cpp
@@ -32,7 +32,7 @@
 #include <QRandomGenerator>
 
 #ifdef WITH_OPENCV
-#include "opencv2/imgproc/imgproc_c.h"
+#include "opencv2/imgproc/imgproc.hpp"
 #endif
 
 namespace nmp
@@ -91,7 +91,7 @@ double DkSkewEstimator::getSkewAngle()
         cv::Mat matGray;
 
         if (mMatImg.channels() > 1)
-            cv::cvtColor(mMatImg, matGray, CV_BGR2GRAY);
+            cv::cvtColor(mMatImg, matGray, cv::COLOR_BGR2GRAY);
         else
             matGray = mMatImg;
 
@@ -224,7 +224,7 @@ cv::Mat DkSkewEstimator::computeSeparability(cv::Mat integral, cv::Mat integralS
 
     // for displaying:
     // cv::normalize(separability, separability, 0, 255, NORM_MINMAX, CV_8UC1);
-    // cvtColor(separability, separability, CV_GRAY2RGB);
+    // cvtColor(separability, separability, cv::COLOR_GRAY2RGB);
 
     return separability;
 }
--- a/ImageLounge/plugins/CompositePlugin/src/SbChannelWidget.cpp
+++ b/ImageLounge/plugins/CompositePlugin/src/SbChannelWidget.cpp
@@ -14,7 +14,6 @@
 #include <QSettings>
 
 #include "opencv2/imgproc/imgproc.hpp"
-#include "opencv2/imgproc/imgproc_c.h"
 
 namespace nmc
 {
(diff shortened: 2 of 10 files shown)


FabioLolix:
Is this AI written?

scrubbbbs:
Thanks, it seems to work on opencv 4. 

This is more of a refactor than a fix based on content (renaming C symbols to their c++ equivalents), so adjust the commit message appropriately, for example:

```
refactor(opencv): drop opencv C headers and support opencv 5

Replace C api constants with their C++ equivalents and add header file changes required for opencv 5
```

scrubbbbs:
Please also fix the clang-format errors. On Linux, you can check your commit is good with "scripts/format-cpp.sh"

jetm:
@scrubbbbs Reworded the commit to the refactor form you suggested and sorted the clang-format errors. My Arch box ships clang-format 22, so that's what I reached for first, but its output diverges from CI's v18 and would have kept the check red. I pinned clang-format-18 to match CI instead. It flagged one trailing-comment alignment in `DkBasicLoader.cpp` that the constant rename had knocked off-column; that's folded into the same commit. `git clang-format --diff` is clean now, and the branch still builds against opencv 5.0.0, which Arch already ships (rolling release, moves fast). Bumping CI to clang-format 22 might be worth it eventually, but that's a separate change. Force-pushed.

jetm:
@FabioLolix I hit the opencv 5.0 build break on my own machine (the same `imgproc_c.h` / `CV_*` failure from #1619) and did the port by hand, checking each renamed constant against the opencv 5 headers and rebuilding against 5.0.0 to confirm nothing shifted. If a specific hunk looks wrong, point me at it.

scrubbbbs:
Thanks! The commit message is still a bit long for my preference, it could be summarized sufficiently with one or two sentences (see my suggestion)

jetm:
@scrubbbbs Done. Trimmed the body to the one-line summary you suggested and force-pushed.

PR #1626, opened 2026-07-13
Fix restart nomacs issue, improve messagebox, kimageformats dialog

scrubbbbs:
Fixes the problem where nomacs restarts and leaves another instance running. This is not possible anymore since we don't restart until all windows are closed.

Refactor `DkMessageBox` to remove `WA_DeleteOnClose` odd design and follow code conventions.

New feature to choose a duration when saving dialog responses:
image

Open issues to discuss:
- ~Should the session remain valid for all instances, including mode changes? I think yes.~
- Some options should probably be disabled depending on dialog:
    - "This Session" doesn't make sense for a dialog that will quit nomacs
    - Longer durations could be disabled for "dangerous" features like file deletion or overwriting.


scrubbbbs:
New dialog when file load fails and kimageformats is missing (linux/unix only)

image

ImageLounge/src/DkCore/DkImageContainer.cpp
```diff
@@ -693,6 +694,42 @@ void DkImageContainerT::loadingFinished()
...
+#ifndef Q_OS_WIN
+        static bool kimageformatsCheck = false;
+        if (!kimageformatsCheck) {
+            kimageformatsCheck = true;
+
+            const QStringList pluginPaths = QCoreApplication::libraryPaths();
+
+            bool exists = false;
+            for (auto &path : pluginPaths) {
+                const QString fullPath = path + "/imageformats/kimg_qoi.so";
```
novomesk: I suggest that instead of searching for kimg_qoi.so, just check if `QImageReader::supportedImageFormats()` contains `qoi` 

Is it possible to limit the dialog only for specific set of file extensions we know kimageformats potentially supports?
So that the dialog is not displayed when there is absolutely no chance that installing kimaformats would make any difference.

scrubbbbs: > I suggest that instead of searching for kimg_qoi.so, just check if QImageReader::supportedImageFormats() contains qoi.

Thanks, that is much better

> Is it possible to limit the dialog only for specific set of file extensions we know kimageformats potentially supports?
> So that the dialog is not displayed when there is absolutely no chance that installing kimaformats would make any difference.

This would be a maintenance issue to keep that list up to date, but it could be done.

scrubbbbs:
> Some options should probably be disabled depending on dialog:
> 
>     * "This Session" doesn't make sense for a dialog that will quit nomacs

I think this has to stay since quitting nomacs could mean restarting for a settings change or switching to frameless mode where it could be useful.

>     * Longer durations could be disabled for "dangerous" features like file deletion or overwriting.

I'm now leaning towards leaving this open. The "Forever" option to remember dialogs has always been a feature, though we know it has well-known downsides I hesitate to take it away for some things and not others.

scrubbbbs:
Final layout unless there are objections. Unless there is any problem I plan to merge this soon. I indented and trimmed the combo box:

image

@novomestk, @leejuyuu. What do we think about reworking the dialogs in nomacs in general? I feel they are unpolished in many ways. I have a bunch of ideas (not for this PR)
- standardize language/tone/grammar
- standardize formatting for file names (elide or wrap long names, don't show full path, show in different color)
- on buttons instead of Yes/No/Cancel use action words like "Save/Discard/Cancel", "Overwrite,Save Copy, Cancel"
- maybe use rich text formatting (bolded important words)

ImageLounge/src/DkCore/DkImageContainer.cpp
scrubbbbs: Added this, file extensions were extracted from kimageformats plugin json files.

novomesk:
I am not against the text changes/re-wording. Just avoid frequent changes because it would have impact on translations.

scrubbbbs:
> I am not against the text changes/re-wording. Just avoid frequent changes because it would have impact on translations.

We could possibly freeze translation updates for a while to prevent this from happening. So translators are not seeing anything to update until we accumulate a large batch of changes.

Issue #257, opened 2018-09-13
Support HEIF and HEIC File formats


asjones987:
I may be too early, but i would be great if nomacs supported HEIF and HEIC file formats (some reading shows Apple's HEIC is a bit different than just HEIF for images.

I am sure you have already looked into this and I am confused on the licensing but some things indicate that for free/open source software it may not be an issue vs commercial software.

Really I would like to read and write HEIF/HEIC.  If the standard really took off now that Apple is pushing it (I am not an Apple person) then I would want to batch convert old pictures just to save space  and make them easier to move around.

https://nokiatech.github.io/heif/

https://github.com/nokiatech/heif

https://en.wikipedia.org/wiki/High_Efficiency_Image_File_Format

http://jpgtoheif.com/

.... and it also supports animated images aka like what animated GIF does

asjones987:
Side note I saw that ImageMagick supports the HEEIC format  https://github.com/ImageMagick/ImageMagick

diemmarkus:
we'll try this [Qt plugin](https://github.com/jakar/qt-heif-image-plugin) (since it seems to take longer for Qt to support it [officially](https://bugreports.qt.io/browse/QTBUG-69212)).

geroldmeisinger:
> the next nomacs will be shipped with (rudimentary) heic support.

...but it was not...

https://github.com/nomacs/nomacs/issues/469#issuecomment-2168589297

scrubbbbs:
There was an effort to get it into Qt (at the time) but it stalled. kimageformat-plugins is widespread, or else continue to use qt-heif-image-plugin. We should update the readme and install page [update this page](https://nomacs.org/docs/getting-started/installation/) to reflect that.

InfinityAX:
I checked the install page, and it still needs to be updated. Or, at the very least, instructions could have been more straightforward. For Ubuntu and Linux Mint, run `$sudo apt install kimageformat-plugins qt5-avif-image-plugin`, or you can use the package manager shown below:
![Screenshot from 2024-09-21 10-17-27](https://github.com/user-attachments/assets/adf36e45-3b1d-4737-8190-180347d55f6f)

kintaro1981:
https://github.com/strukturag/libheif
libheif is an HEIF and AVIF file format decoder and encoder.

scrubbbbs:
Since readme is updated this is resolved. kimageformats or another suitable heif plugin for qt is the path forward.

step135:
[...]

PR #1630, opened 2026-07-17
Various trivial fixes

scrubbbbs:


novomesk:
I am curious. Do you have some examples of the "Sony HEIF" images with .HIF extension?

scrubbbbs:
> I am curious. Do you have some examples of the "Sony HEIF" images with .HIF extension?

I was unable to find any and renamed some heic files for testing. The OP said renaming the file [allowed it to open](https://github.com/nomacs/nomacs/issues/257#issuecomment-4177127497) so I assume it is only the extension that is at issue.

novomesk:
It would be useful to ask the OP to contribute some of those images for testing purposes.

scrubbbbs:
I did some digging and found some, added to formats_testset

ImageLounge/src/DkCore/DkSettings.cpp
```diff
@@ -220,6 +219,8 @@ void DkSettings::initFileFilters()
         app_p.openFilters.append("AVIF image sequence (*.avifs)");
     if (qtFormats.contains("apng"))
         app_p.openFilters.append("Animated Portable Network Graphics (*.apng)");
+    if (qtFormats.contains("heif"))
```
novomesk: I think it will be better to add small code to `kimg_heif` to advertise it supports "HIF" format too.

I think .HIF is a general extension and it is not specific to Sony only.

novomesk: kimg_heif.dll on Windows already indicate **hif** reading capability.

scrubbbbs: Thanks, I assume you mean since your commit 7ec1170?

PR #1632, opened 2026-07-20
Copy file paths as text from the thumbnail grid

ludekvodicka:
## What

`DkThumbScene::copySelected()` (thumbnail grid → *Copy* / `Ctrl+C`) put the selected files on the clipboard only as URLs. This also adds the file paths as newline-separated plain text, so a grid selection can be pasted into a terminal, an editor, or any text target.

## Why

The viewport *Copy* already copies the file path as text (since "change edit/copy to always copy the file path"), but the thumbnail grid did not — pasting a grid selection into a plain-text target produced nothing. This aligns the grid with the viewport.

## Change

One line in `ImageLounge/src/DkGui/DkThumbsWidgets.cpp`, in `copySelected()`:

```cpp
mimeData->setUrls(urls);
mimeData->setText(fileList.join("\n"));   // added
```

The URLs are unchanged, so pasting the selection as files (Explorer, mail, and nomacs' own paste handler, which reads `hasUrls()`) keeps working; the added text is only used by targets that want text.

## Testing

Built and run on Windows (MSVC 2022, Qt 6.11.1). Selecting images in the thumbnail view and pressing `Ctrl+C` now places the newline-separated file paths on the clipboard as text (verified by reading the clipboard back), in addition to the file URLs. As noted in review, this is worth checking on other file managers / OSes and against nomacs' own paste handlers.

### DIFF
--- a/ImageLounge/src/DkGui/DkThumbsWidgets.cpp
+++ b/ImageLounge/src/DkGui/DkThumbsWidgets.cpp
@@ -1751,6 +1751,7 @@ void DkThumbScene::copySelected() const
 
     auto *mimeData = new QMimeData();
     mimeData->setUrls(urls);
+    mimeData->setText(fileList.join("\n"));
 
     QApplication::clipboard()->setMimeData(mimeData);
 }


scrubbbbs:
I don't understand why this is needed. Do you have any concrete example? I tested with several terminal programs and editors and they all pasted the plain text, not the URL.

`DkViewPort::copyImagePath()` sets both URL and plain text MIME data, the receiving program chooses which one to use. Some programs have a "Paste as Plain Text" (Ctrl+Shift+V) specifically for this purpose.

The greater issue seems to be that copy path on the thumb grid vs viewport behaves differently. The former does not include plain text so it always pastes as a URL. This one line change fixes the problem:

```diff
--- ImageLounge/src/DkGui/DkThumbsWidgets.cpp
+++ ImageLounge/src/DkGui/DkThumbsWidgets.cpp
@@ -1753,2 +1753,3 @@ void DkThumbScene::copySelected() const
     mimeData->setUrls(urls);
+    mimeData->setText(fileList.join("\n"));
 ```
That needs to be tested on different file managers, OS etc and nomacs own paste handlers to verify it is a valid fix, but I believe it will be.

ludekvodicka:
Sorry for the noise on this one. I was on an older build where *Copy* did not put the path on the clipboard, and I implemented this right after checking out, without testing against current master — where *Copy* already copies the file path. You're right that the viewport action was redundant.

I've repurposed this PR to the actual gap you pointed out: `DkThumbScene::copySelected()` only set the file URLs, so a grid selection never pasted as a path. It now also sets the newline-separated paths as text, matching the viewport. The URLs are untouched, so file-paste (and nomacs' own `hasUrls()`-based paste handler) still works. Verified on Windows; as you said, it's worth checking across file managers / OSes too.

scrubbbbs:
I am just curious is AI writing everything or just the PR text? Because that is a lot more detail that we are used to for something simple.

ludekvodicka:
> I am just curious is AI writing everything or just the PR text? Because that is a lot more detail that we are used to for something simple.

It’s my custom Claude Code agent that helps me with my work. I give it a draft or concept, and it helps me fill in all the necessary information, improve the English, and so on. :-)

Issue #260, opened 2018-10-01
Settings -> Option to Choose Open in New Instance or Open in New Tab


asjones987:
This may actually already be there and I just can't find it (search website and settings).
Currently when I double click on an Image it opens a new instance of Nomacs.  So I have a large number of windows open when I might have preferred tabs. Is there or it would be nice to have a setting to default to open in tabs and not new windows/instances.

Seeing this on Windows 7 Nomacs 3.11

thanks

Alan

diemmarkus:
Hi,

do you mean, when you double click in e.g. windows explorer? if so, it is related to #95 and not there yet - sorry!

best,
markus

asjones987:
Yes, sorry I did not see bug #95.  Double clicking on any application. I had a coworker also send me screen shots in Skype and each one opened in a new Window.  In Some ways I see this bug related to #261 as if the option is set to open in an new tab I will want to "break" that to new windows an if I set it the other way I will want to combine windows to tabs :)

I am find with making this a dupe of #95.

Issue #95, opened 2017-03-06
Single Instance Option


diemmarkus:
if user open an image by double clicking it from a file browser, a tab should be added to an existing instance rather than opening a new instance.

artofit:
In 3.17 no change
Next milestones?

Parodocs:
I'm guessing that there isn't a solution. Odd. All of the other icons in the taskbar stack.
No worries - new software.
Bye.

Zector:
got really excited this would be a feature when i realized settings opens in a new tab. what's up with that?

leejuyuu:
Closing because of duplicate.

PR #1633, opened 2026-07-20
Add single instance and fix tab issues 

### LINKED ISSUE #67: Single Instance
Is this possible? I can't seem to find a setting for it in nomacs or settings.nfo

Win7 x64
3.2 portable

### LINKED ISSUE #1336: Downloading uri-list only opens the last image
When pasting a list of URLs (newline separated), a tab is opened for each url and the urls are download, but only the last image is displayed. The other tabs show recent files, but will display the image if recent files is toggled off.

Seems to have something to do with changes to background images/tabs not being handled.

### LINKED ISSUE #1610: Currently opened image is always replaced on macOS
I think there is an open event on macOS that does this. Maybe the event handler should create a new tab or open in a new instance.

_Originally posted by @gcdukelow in https://github.com/nomacs/nomacs/discussions/1609_

### LINKED ISSUE #1649: Restoring a tab in thumbnail grid mode doesn't show anything
The thumbnail grid panel shows but it is empty. Toggling it off and back on shows the last image but doesn't show any thumbnails.

scrubbbbs:
Closes: #67
Fixes: #1336
Fixes: #1610
Fixes: #1649

This currently works on macOS, Linux, Windows but yet to test flatpak. Tried a few different Linux desktops.

Major remaining work:
- [x] Add [KDSingleApplication](https://github.com/KDAB/KDSingleApplication) to workflows and msvc dependencies
- [x] Add setting for enabling single instance (disabled by default)
- [x] Add setting for open in new tab or replace current tab
- [x] Fix "File/Start New Instance" to bypass the single instance feature
- [x] Determine if we will default to D-Bus on Unix/Linux (YES) and use KDSingleApplication only for macOS and Windows
- [x] Fix race condition when restarting nomacs for settings or mode switch

Notable new features for tabs:
- Don't load anything for inactive restored tabs (faster startup)
- Don't reload anything when switching between previously active tabs (fast switching)
- The duplicate tab check now applies to most things except explorer, recent files
- Some tab types are used for open-file-in-new-tab if they don't contain any image or dir
- Show recent files only if there is nothing to display (no saved tabs or image loading)

I confirmed this works with flatpak since the dbus service name matches the flatpak name (org.nomacs.ImageLounge) this is allowed by default - no changes needed to flatpak manifest.

### DIFF
--- a/ImageLounge/src/DkCore/DkImageLoader.cpp
+++ b/ImageLounge/src/DkCore/DkImageLoader.cpp
@@ -117,25 +117,6 @@ DkImageLoader::~DkImageLoader()
 }
 
 /**
- * Clears the path.
- * Calling this method makes the loader forget
- * about the current directory. It also destroys
- * the currently loaded image.
- **/
-void DkImageLoader::clearPath()
-{
-    // lastFileLoaded must exist
-    if (mCurrentImage && mCurrentImage->exists()) {
-        this->receiveUpdates(false);
-        mLastImageLoaded = mCurrentImage;
-        mImages.clear();
-
-        // only clear the current image if it exists
-        mCurrentImage.clear();
-    }
-}
-
-/**
  * Loads a given directory.
  * @param newDir the directory to be loaded.
  **/
@@ -667,7 +648,15 @@ void DkImageLoader::activate(bool isActive /* = true */)
     blockSignals(!isActive);
 
     if (!isActive) {
-        clearPath();
+        // lastFileLoaded must exist
+        if (mCurrentImage && mCurrentImage->exists()) {
+            this->receiveUpdates(false);
+            mLastImageLoaded = mCurrentImage;
+            mImages.clear();
+
+            // only clear the current image if it exists
+            mCurrentImage.clear();
+        }
     } else if (!mCurrentImage) {
         setCurrentImage(mLastImageLoaded);
     } else {
--- a/ImageLounge/src/DkCore/DkImageLoader.cpp
+++ b/ImageLounge/src/DkCore/DkImageLoader.cpp
@@ -664,16 +664,15 @@ bool DkImageLoader::promptSaveBeforeUnload()
  **/
 void DkImageLoader::activate(bool isActive /* = true */)
 {
+    blockSignals(!isActive);
+
     if (!isActive) {
-        // go to sleep - schlofand wöhlar ihr camölar
-        blockSignals(true);
         clearPath();
     } else if (!mCurrentImage) {
-        // wake up again
-        blockSignals(false);
         setCurrentImage(mLastImageLoaded);
-    } else
+    } else {
         emit updateDirSignal(mImages);
+    }
 }
 
 void DkImageLoader::setCurrentImage(QSharedPointer<DkImageContainerT> newImg)
--- a/ImageLounge/src/DkCore/DkImageLoader.cpp
+++ b/ImageLounge/src/DkCore/DkImageLoader.cpp
@@ -636,31 +636,18 @@ bool DkImageLoader::promptSaveBeforeUnload()
     return false;
 }
 
-/**
- * Activates or deactivates the loader.
- * If activated, the directory is indexed & the current image is loaded.
- * If deactivated, the image list & the current image are deleted which
- * should save some memory. In addition, all signals are mBlocked.
- * @param isActive if true, the loader is activated
- **/
 void DkImageLoader::activate(bool isActive /* = true */)
 {
+    // stop or start using the loader, for example when switching tabs or
+    // a mode switch like thumb scene hides the image
     blockSignals(!isActive);
+    receiveUpdates(isActive);
 
-    if (!isActive) {
-        // lastFileLoaded must exist
-        if (mCurrentImage && mCurrentImage->exists()) {
-            this->receiveUpdates(false);
-            mLastImageLoaded = mCurrentImage;
-            mImages.clear();
-
-            // only clear the current image if it exists
-            mCurrentImage.clear();
+    if (mCurrentImage) {
+        if (!isActive) {
+            // we don't need the image immediately, it can be discarded after loading
+            mCurrentImage->cancel();
         }
-    } else if (!mCurrentImage) {
-        setCurrentImage(mLastImageLoaded);
-    } else {
-        emit updateDirSignal(mImages);
     }
 }
 
--- a/ImageLounge/src/DkCore/DkLocalIPC.cpp
+++ b/ImageLounge/src/DkCore/DkLocalIPC.cpp
@@ -1,0 +1,254 @@
+#include "DkLocalIPC.h"
+
+#include "DkCentralWidget.h"
+
+#include <QWindow>
+
+#ifdef WITH_KDSINGLEAPPLICATION
+#include "kdsingleapplication.h"
+#endif
+
+#ifdef WITH_DBUS
+#include <QtDBus/QtDBus>
+#endif
+
+namespace nmc
+{
+
+#ifdef WITH_KDSINGLEAPPLICATION
+
+class DkLocalSocketIPC : public QObject, public DkLocalIPC
+{
+    Q_OBJECT
+    Q_DISABLE_COPY_MOVE(DkLocalSocketIPC)
+
+public:
+    DkLocalSocketIPC()
+    {
+        if (mSocket.isPrimaryInstance()) {
+            connect(&mSocket, &KDSingleApplication::messageReceived, this, &DkLocalSocketIPC::messageReceived);
+        }
+    }
+    ~DkLocalSocketIPC() override = default;
+
+private:
+    bool isFirstInstance() const override
+    {
+        return mSocket.isPrimaryInstance();
+    }
+
+    void messageReceived(const QByteArray &msgData)
+    {
+        Q_ASSERT(isFirstInstance());
+
+        // TODO: validate the message, invalid data will segfault
+        // in QDataStream::operator>>
+        QByteArray encoded = msgData;
+        QDataStream data(&encoded, QIODevice::ReadOnly);
+        QStringList msg;
+        data >> msg;
+        qInfo() << "[local-socket]: msg received:" << msg;
+        if (msg.empty()) {
+            qWarning() << "[local-socket]: ignoring empty message";
+            return;
+        }
+        if (!mCentralWidget) {
+            qWarning() << "[local-socket]: no registered central widget";
+            return;
+        }
+        if (msg[0] == "activate" && msg.length() == 2) {
+            const QString &token = msg[1];
+            qInfo() << "[local-socket] activate:" << token;
+
+            // TODO: we need to know if we are setting wayland or X11 token
+            qputenv("XDG_ACTIVATION_TOKEN", token.toLatin1());
+
+            QWidget *top = mCentralWidget->topLevelWidget();
+            top->show();
+
+            QWindow *window = top->windowHandle();
+            if (window) {
+                window->requestActivate();
+            }
+
+        } else if (msg[0] == "load" && msg.length() == 2) {
+            mCentralWidget->load(msg[1]);
+        } else if (msg[0] == "loadToTab" && msg.length() == 2) {
+            mCentralWidget->loadToTab(msg[1]);
+        } else {
+            qWarning() << "[local-socket]: invalid/unknown message" << msg;
+        }
+    }
+
+    void setCentralWidget(DkCentralWidget *centralWidget) override
+    {
+        Q_ASSERT(isFirstInstance());
+        mCentralWidget = centralWidget;
+    }
+
+    void sendMessage(const QStringList &msg)
+    {
+        Q_ASSERT(!isFirstInstance());
+        QByteArray encoded;
+        QDataStream ds(&encoded, QIODevice::WriteOnly);
+        ds << msg;
+        mSocket.sendMessage(encoded);
+    }
+
+    void activate() override
+    {
+        QStringList msg;
+        // TODO: on X11 send DESKTOP_STARTUP_ID
+        msg << "activate" << qEnvironmentVariable("XDG_APPLICATION_TOKEN");
+        sendMessage(msg);
+    }
+
+    void load(const QString &path) override
+    {
+        QStringList msg;
+        msg << "load" << path;
+        sendMessage(msg);
+    }
+
+    void loadToTab(const QString &path) override
+    {
+        QStringList msg;
+        msg << "loadToTab" << path;
+        sendMessage(msg);
+    }
+
+    KDSingleApplication mSocket{};
+    DkCentralWidget *mCentralWidget{};
+};
+
+#endif // WITH_KDSINGLEAPPLICATION
+
+#ifdef WITH_DBUS
+
+class DkDBusAdapter : QDBusAbstractAdaptor
+{
+    Q_OBJECT
+    Q_CLASSINFO("D-Bus Interface", "org.nomacs.ImageLounge.main")
+public:
+    DkDBusAdapter(DkCentralWidget *cw, QObject *parent)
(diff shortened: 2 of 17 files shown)


novomesk:
Why 2 methods? (dbus vs kdsingleapplication)

I'd prefer just one to prevent complexity of multiple configurations.

scrubbbbs:
While kdsingleapplication also works on Linux, it is non-standard. D-Bus is ubiquitous on Linux/Unix and has a few added benefits:
- No special permissions needed for flatpak
- No added dependency to maintain; included by default with qtcore
- Future features will want to use dbus, such as setting the desktop wallpaper

Abendstern51:
Hmm... how could I use it already? The way default nomacs opens everything in different windows makes me want to rip the skin off my knees.

scrubbbbs:
> Hmm... how could I use it already? The way default nomacs opens everything in different windows makes me want to rip the skin off my knees.

We have continuous builds you could try. Look for latest CI workflow runs on the `single-instance` branch (you may have to scroll for a bit), and download the build artifact for your system:
- Linux/macOS https://github.com/nomacs/nomacs/actions
- Windows https://ci.appveyor.com/project/novomesk/nomacs/history

There is a checkbox to enable it on general settings.

scrubbbbs:
@novomesk  @leejuyuu  pretty much done on this one too. There will be a separate PR to work though remaining tab issues.

leejuyuu:
Is this the intended behavior? Opening new images replace the current tab instead of open a new tab.

https://github.com/user-attachments/assets/283b2b7a-8033-4bcd-b434-1c3e38555aaa

leejuyuu:
Single instance allows nomacs starting with `--private` to end up in a non-private instance.

scrubbbbs:
> Is this the intended behavior? Opening new images replace the current tab instead of open a new tab.
> 2026-09-06.18-47-56.mp4

It's an option on general prefs. The default is to replace, to provide a single-image/single-tab experience by default.

leejuyuu:
> > Is this the intended behavior? Opening new images replace the current tab instead of open a new tab.
> > 2026-09-06.18-47-56.mp4
> 
> It's an option on general prefs. The default is to replace, to provide a single-image/single-tab experience by default.

This is not what I would expect when I am enabling the single-instance option. Most things with tabs default to opening a new tab.

leejuyuu:
Relative paths are not resolved before sending to the main instance. For example, if I do `./nomacs ../ImageLounge/screenshots/nomacs_3_21-raw_metadata.png &` in the build directory, `cd ..` and then `./build/nomacs ./ImageLounge/screenshots/nomacs_3_22-avif_settings.png &`. Should work but fails.

ImageLounge/src/DkCore/DkImageLoader.cpp
```diff
@@ -750,6 +750,12 @@ void DkImageLoader::load(QSharedPointer<DkImageContainerT> image /* = QSharedPoi
     if (!image)
         return;
 
+    if (image->getLoadState() == DkImageContainerT::loaded) {
+        mCurrentImage = image;
+        imageLoaded(true); // pretend we loaded image so all signals are delivered
+        return;
+    }
+
```
leejuyuu, line 758: This should be moved below the block of `mCurrentImage && mCurrentImage->getLoadState() == DkImageContainerT::loading` and do not modify mCurrentImage. It needs to go through setCurrentImage for the signals to wire properly (by receiveUpdates).

ImageLounge/src/DkGui/DkCentralWidget.cpp
```diff
@@ -485,9 +485,6 @@ void DkCentralWidget::updateLoader(QSharedPointer<DkImageLoader> loader) const
     if (!loader)
         return;
 
-    if (hasViewPort())
-        getViewPort()->setImageLoader(loader);
-
```
leejuyuu, line 490: This should not be removed. The viewport also stores this as a QSharedPointer, so it will continue to hold a stale one if a tab is closed.

ImageLounge/src/DkCore/DkLocalIPC.cpp
```diff
@@ -0,0 +1,564 @@
...
+#if defined(WITH_KDSINGLEAPPLICATION)
+    static DkLocalSocketIPC ipc;
+#elif defined(WITH_DBUS)
+    static DkDBusIPC ipc;
+#else
+#error No IPC implementation configured
+#endif
+    return ipc;
+}
+}
```
leejuyuu, line 562: Is this check necessary? This will force the KDSingleApplication dependency outside Linux distro that supports DBus, even when the single instance feature is not needed.

ImageLounge/src/DkCore/DkLocalIPC.cpp
```diff
@@ -0,0 +1,564 @@
...
+        qInfo() << "[dbus] waiting for first instance to exit";
+        QDeadlineTimer timer(5000);
+        while (!mFirstInstance && !timer.hasExpired()) {
+            QThread::msleep(100);
+            mFirstInstance = mDbus.registerService("org.nomacs.ImageLounge");
+        }
+        if (!mFirstInstance) {
+            qWarning() << "[dbus] timed out waiting for first instance";
+        }
+    }
```
leejuyuu, line 479: If the connection fails, e.g., for a system that is using neither common DE nor systemd (no DBus), this should allow fallback to no single instance behavior.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
```diff
@@ -900,7 +900,8 @@ void DkFilePreview::updateThumbs(QVector<QSharedPointer<DkImageContainerT>> imag
         }
     }
 
-    update();
+    // do not update here; we'll wait for file index to avoid repaints
+    // update();
```
leejuyuu, line 904: This makes it possible for directory update not trigger repaint. It can be reproduced by having two files in a directory. Do a `sleep 5s && trash file2` and focus on nomacs before the trash happens, the thumbnails ribbon will not update when the trash completed.

ImageLounge/src/DkCore/DkImageLoader.cpp
```diff
@@ -636,31 +636,18 @@ bool DkImageLoader::promptSaveBeforeUnload()
...
- * Activates or deactivates the loader.
- * If activated, the directory is indexed & the current image is loaded.
- * If deactivated, the image list & the current image are deleted which
- * should save some memory. In addition, all signals are mBlocked.
- * @param isActive if true, the loader is activated
- **/
 void DkImageLoader::activate(bool isActive /* = true */)
 {
+    // stop or start using the loader, for example when switching tabs or
+    // a mode switch like thumb scene hides the image
```
leejuyuu, line 642: Nit: it is weird to see a refactor with a `!` that indicates a breaking change. If it changes visible behavior, then it is probably part of a feature or fix.

leejuyuu (commented):
a11141a commit message is still WIP

leejuyuu:
Another potential issue is that the IPC starts listening to events before the central widget is created. This creates a (potentially short) window during which a second instance can connect, but the event gets dropped.

ImageLounge/src/DkCore/DkImageLoader.cpp
scrubbbbs: I tried to stay out of setCurrentImage() to limit the blast radius of any changes. A refactor can be done at later time to clean things up. I suspect loadImage/setCurrentImage could be combined since there is never a case where you want to set the image and not load anything.

It does work properly for now since when the loader is activated receiveUpdates() is called, which always happens when loading a new image to a tab or switching tabs.

leejuyuu: No, it is not working properly now. Try it in a directory with 2 images, navigate through both files so both are loaded, then delete the first image. Then, the second should be visible, but actions like `R` have no effect.

ImageLounge/src/DkCore/DkLocalIPC.cpp
scrubbbbs: KDSingleApplication isn't forced. We only require it on Linux/Unix if `-DWITH_DBUS=OFF`

ImageLounge/src/DkCore/DkLocalIPC.cpp
scrubbbbs: It won't crash the app, just delay startup by 5s. In the extremely rare case the system is unsupported they can disable the feature in settings.

scrubbbbs:
> Another potential issue is that the IPC starts listening to events before the central widget is created. This creates a (potentially short) window during which a second instance can connect, but the event gets dropped.

Seems to be an issue on DBus but not KDSA. Looking for a solution.

Update: solved in yesterdays push, I was not sequencing the DBus setup correctly.

ImageLounge/src/DkCore/DkLocalIPC.cpp
leejuyuu: It not only delays startup, but also makes mFirstInstance remain false, and then the app just exits.

ImageLounge/src/DkCore/DkLocalIPC.cpp
leejuyuu: I mean, it is forced on Windows or Mac.

ImageLounge/src/DkCore/DkImageLoader.cpp
scrubbbbs: Thanks, yes it seems loadImage() is used in some places when it should be setCurrentImage() behavior. I added a fix to call receiveUpdates() if needed.

ImageLounge/src/DkCore/DkLocalIPC.cpp
scrubbbbs: There is no other option on Windows/macOS. Single-instance is not an optional feature.

ImageLounge/src/DkGui/DkCentralWidget.cpp
scrubbbbs: I don't think that matters much for now, and I have a fix for this already in the followup PR to this one

ImageLounge/src/DkCore/DkLocalIPC.cpp
scrubbbbs: Yes I did not notice that one. Fixed in today's push, added isOk() to check if app should exit or not.

ImageLounge/src/DkGui/DkThumbsWidgets.cpp
scrubbbbs: I've changed loader to always send the index changed single when dir is resorted or read which fixes the issue.

scrubbbbs:
> > > Is this the intended behavior? Opening new images replace the current tab instead of open a new tab.
> > > 2026-09-06.18-47-56.mp4
> > 
> > It's an option on general prefs. The default is to replace, to provide a single-image/single-tab experience by default.
> 
> This is not what I would expect when I am enabling the single-instance option. Most things with tabs default to opening a new tab.
 
The open in new tab option has been around for a while and defaults to false. So we would have one setting overriding or changing another. Besides that conundrum I can make the case for why defaults to false may be better.

- nomacs is first an image viewer, second an editor/utility. So for most users, most of the time, they aren't wanting to keep the previous image in a tab.
- having to close tabs all of the time somewhat defeats the benefit of single-instance which is not having to close the window after opening each image. This is made worse when tabs are set to be saved, they will have a tendency to accumulate and require more attention than not using tabs at all.

scrubbbbs:
> The open in new tab option has been around for a while and defaults to false. So we would have one setting overriding or changing another. 

Reading this back I had forgotten it wasn't an option that was around but the *behavior* was, which was we open a new tab only if at least two tabs are already open. So the default is not to create a new tab in effect, until the user added a new tab (or opened settings/batch.

ImageLounge/src/DkCore/DkLocalIPC.cpp
leejuyuu: Then KDSingleApplication should be documented as a required dependency on these platforms. Also, it seems mingw build defaults to DBUS?

ImageLounge/src/DkGui/DkCentralWidget.cpp
leejuyuu: Some actions are still connected. Try opening two images in separate tabs. Rotate clockwise in the first, switch to the other, and rotate it as well, then undo. The undo will apply to the first tab as well.

ImageLounge/src/DkCore/DkLocalIPC.cpp
scrubbbbs: After this merge we can update the readme, a lot is out of date now. @novomesk wants to keep submodules out of nomacs, and I agree it is a good idea for maintenance. So we'll have to add build instructions for KDSA.

I'll remove ENABLE_DBUS on targets that don't support it.

ImageLounge/src/DkGui/DkCentralWidget.cpp
scrubbbbs: fix pushed

scrubbbbs: I backed out this change, and accomplish the same thing, when a tab is deactivated it gets an empty image loader.

ImageLounge/src/DkCore/DkImageLoader.cpp
scrubbbbs: I decided to back out this change, it doesn't seem to help with reloading the image after all, and has the caveats you noted.

scrubbbbs:
@leejuyuu  I think your disagreement with the default for open in new tab is the only remaining unresolved issue you brought up. Did you see my response?

leejuyuu:
I think opening a new tab is what people expect. See

- https://github.com/nomacs/nomacs/issues/95
- https://github.com/nomacs/nomacs/issues/260

In the same instance, opening in the current tab is reasonable because we are in the context. However, when opening another image in the single-instance setting, the user is likely not looking at the nomacs window; this makes changing the current tab less natural because it is not immediately clear which image we navigated away from. For example, browsers behave like this as well. Always appending a new tab would likely make the logic simpler to handle.

leejuyuu:
Another bug: instances started by "New instance" always wait for 5 s before actions require restarting, e.g. frameless, setting apply.

ImageLounge/src/DkCore/DkLocalIPC.cpp
```diff
@@ -358,6 +358,9 @@ class DkLocalSocketIPC : public QObject, public DkLocalIPC
     {
         Q_ASSERT(isFirstInstance());
         mCentralWidget = centralWidget;
+        QObject::connect(mCentralWidget, &QObject::destroyed, this, [this]() {
+            mCentralWidget = nullptr;
+        });
```
leejuyuu: Maybe this is a good place to use QPointer? These should not be needed then.

scrubbbbs:
> I think opening a new tab is what people expect. See

Changed in latest push, if the file open is external we make a new tab

leejuyuu:
I am not sure whether we still need the setting. Its existence would likely confuse because whether it takes effect is conditional, and the user likely cannot understand the distinction at first glance. I know there's a tooltip that describes this, but people do not always read it, and, as the commit message shows, the whole condition cannot be summarized well in one line. Can this follow the original behavior before this patch?

scrubbbbs:
> I am not sure whether we still need the setting

As unchecked, it doesn't follow the old behavior which was:
- one tab, always open in current tab
- more than one tab, always open in a new tab

At the time I thought that was even weirder than adding the checkbox, so at least you could have it one way or the other and not the mixed condition.

Another option is we drop the old behavior and setting and you just get something simpler and less confusing (hopefully). If you want to open something in a new tab more efficiently, we'd have to specifically add menus, buttons or hotkeys for that.

PR #1666, opened 2026-08-19
ci(linux): add arm64/aarch64 AppImage

scrubbbbs:


scrubbbbs:
@novomesk I noticed HEIC loading on neon (latest stable ISO from today) doesn't work with only kf6-kimageformats-plugins installed. kimg_heif.so is installed and Qt loads it, and libheif scans its plugins, but QImageReader doesn't list heif as supported. I was able to get it working by also installing libheif-plugins-libde265. By default the only heif plugins are the aom ones.

Does this sound like a bug to you, or was I doing it wrong all along?

Edit: do you think this change was made due to h.265 licensing concerns, e.g. ubuntu policy is not to install such software by default.

Edit: [answered my own question](https://daily.dev/posts/fix-heic-images-not-loading-in-ubuntu-26-04-lts-ovreedjjz)

novomesk:
libheif can have different configurations with or without HEVC decoders/encoder.

kimg_heif.so uses run-time detection whether the decoder is present and signals heif support accordingly.

PR #1669, opened 2026-08-21
Misc fixes: thumbscene double-click, broken TIFF files prevention, animation freeze, shortcuts, dir watcher, save-as drops metadata, tiff color profile,  painting to indexed

scrubbbbs:


scrubbbbs:
@novomesk @leejuyuu , I think this is at a good place to merge. I'll stop updating until you are able to sign off.

ImageLounge/src/DkGui/DkViewPort.cpp
```diff
@@ -1000,29 +1001,44 @@ void DkViewPort::loadMovie()
...
+    // Quick check if we have animation without reading the whole thing,
+    // otherwise we read the file again in the image loader.
+    {
+        std::unique_ptr<QIODevice> io = fileInfo.getIODevice();
+        if (!io) {
+            return;
+        }
+
+        QImageReader reader(io.get(), format);
+        if (!reader.canRead() || !reader.supportsAnimation() || reader.imageCount() <= 1) {
```
novomesk: when `imageCount()` returns **0**, it means that total number of images is not yet known.

"If the image format does not support animation, or if it is unable to determine the number of images, 0 is returned.
The default implementation returns 1 if canRead() returns true; otherwise 0 is returned."
https://doc.qt.io/qt-6/qimageiohandler.html#imageCount

I know that https://doc.qt.io/qt-6/qimagereader.html#imageCount documentation doesn't mention that, but QImageReader forwards what it receives from QImageIOHandler.

Actually, the MNG handler has such behavior that it knows the total number only after reading all frames.

We have a testfile here: https://github.com/nomacs/formats_testset/blob/main/animated/numbers.mng

novomesk:
Are you able to build [kimageformats](https://invent.kde.org/frameworks/kimageformats) master on your computer.

I made a few changes recently in AVIF and JXL plugins that may have impact on animation support. It is worthy to test with this PR.

ImageLounge/src/DkGui/DkViewPort.cpp
scrubbbbs: Interesting, nice catch. I wasn't aware qt had MNG plugin, it was always missing from builds and my development Qt build by coincidence. I have that fixed now.

Do you know of any APNG plugin that works? I tried https://github.com/fanxinglanyu/Qt6APNG, no luck. The plugin registers but `reader.supportsAnimation()` is false. Although at first glance the source code seems correct in this regard.

scrubbbbs:
> Are you able to build [kimageformats](https://invent.kde.org/frameworks/kimageformats) master on your computer.
> 
> I made a few changes recently in AVIF and JXL plugins that may have impact on animation support. It is worthy to test with this PR.

Could be a pain, I am still working from Ubuntu 24.04 and deps are always breaking. What is the philosophy of having versions locked into such narrow ranges? It's like there is no leeway at all for anything.

It will be easier to compile on VM from neon-iso, I might try that.

ImageLounge/src/DkGui/DkViewPort.cpp
novomesk: I think I tried DLL from https://github.com/Skycoder42/QtApng in the past and it worked but it was when nomacs used Qt5.

novomesk:
> Could be a pain

Manually edit [CMakeLists.txt](https://invent.kde.org/frameworks/kimageformats/-/blob/master/CMakeLists.txt), try to change these two lines (set older versions):

`set(KF_DEP_VERSION "6.29.0") # handled by release scripts`

`find_package(ECM 6.29.0  NO_MODULE)`

scrubbbbs:
> Manually edit [CMakeLists.txt](https://invent.kde.org/frameworks/kimageformats/-/blob/master/CMakeLists.txt), 

Thanks, the downgrade worked. Did not see any issues with latest plugins and animations. Also Windows animated cursor files (ANI) work now.

PR #1688, opened 2026-09-04
Update dependencies on Windows

novomesk:


scrubbbbs:
Can we add gtest? I think it will be possible to run tests on windows soon.

novomesk:
> Can we add gtest?

It's possible but I don't have experiences with that. I don't know if you want shared or static build, etc...

Issue #1011, opened 2024-01-06
Long freeze when containing folder has a disconnected network shortcut


DaPa:
Example scenario on Windows: if Desktop folder has a network drive shortcut  which is disconnected then NoMacs will freeze for tens of seconds when trying to open an image from Desktop, this is not the happening with other image viewers (e.g. IrfanView).
A solution I found was to limit folder scanning to certain known extensions only, see https://github.com/DaPa/nomacs/commit/e6cdafbfd02bb7e44db59c7f639212593f9f3fc2

 - OS: Windows 10 x64 121H2
 - Version: latest 3.17.2289 Beta (https://github.com/nomacs/nomacs/releases/download/3.17.2287/nomacs-portable-win.zip)

scrubbbbs:
> A solution I found was to limit folder scanning to certain known extensions only, see https://github.com/DaPa/nomacs/commit/e6cdafbfd02bb7e44db59c7f639212593f9f3fc2

The directory scanner resolves shortcuts and checks the file header when there is no file extension. Both of which will touch the remote filesystem. We would need an option to disable that.

PR #1689, opened 2026-09-07
Fix folder-freeze and startup-persistence issues on slow/network drives

fabrice35860-web:
## Summary

Three related fixes for issues encountered when browsing folders on
slow network mounts (NAS, GVFS):

1. **Clear recent history button** — adds a "Clear Now" button next
   to "Remember Recent Files History" in preferences, so users can
   purge accumulated `recentFiles`/`recentFolders` entries without
   resetting all settings.

2. **Confirm before loading large files** — adds an optional
   confirmation dialog (with configurable size threshold) before
   loading files over a certain size, since a single click on a
   large file over a slow network mount could previously freeze the
   app with no feedback and no way to back out.

3. **Startup mode setting + stop loading button** — adds a
   preference to control what nomacs shows on launch (restore last
   session / start empty / start in home folder), fixing three
   separate persistence mechanisms that were bypassing this setting
   (DkTabInfo's saved path, DkBrowseExplorer's saved root path, and
   Qt's own QFileDialog recent-places history). Also adds a "Stop
   Loading" toolbar button (with configurable shortcut) to cancel
   pending thumbnail requests in the preview panel — related to
   issue #1011.

## Testing

Tested on Linux (Debian, Qt 6.8.2) against a slow NFS/GVFS-mounted
network share. Not yet tested on Windows or macOS.

Includes updated French translations for all new strings.


fabrice35860-web:
For anyone who wants to quickly test these changes without compiling: a Linux x86_64 AppImage build is available here: https://github.com/fabrice35860-web/nomacs/releases/tag/v3.23.3-loading-progress-test

scrubbbbs:
There are a lot of good ideas here, but trying to do too much in one pull request for a new contributor.

Also there is no open issue related to this PR, to discuss workarounds or different solutions. Additionally:
- Did not follow Contributing guidelines WRT commit messages and size/scope of commits
- Translations should not be included, they should be separate PR for only the affected language, or preferably authored on CrowdIn.
- Don't post executable binaries from unofficial/untrusted sources. If non-developer wants to test they can download the workflow artifact when we review changes and approve the workflow.

To proceed I suggest you open an issue for one of the things you want to do and we will go from there. State your case as to why you feel it is important and how you want to go about fixing it.  I suggest choosing something simple since we are nearing feature freeze for 3.24 release. 

I hope that is not too discouraging but baby-steps is the best way to be a new contributor. Many have given up after similar comments, I hope you will keep trying.

fabrice35860-web:
Updated test build available: a newer AppImage including the Emboss plugin and extended Paint plugin (shadow, textures, custom shapes, text rotation, layer ordering) is now here: https://github.com/fabrice35860-web/nomacs/releases/tag/v3.23.3-loading-progress-test-2

fabrice35860-web:
Thank you for the detailed feedback, I appreciate you taking the time to explain the concerns clearly.

You're right that this PR grew too large in scope, and I understand the points about commit conventions, translations, and not posting unofficial binaries. That makes sense from a maintainer's perspective.

I'll take the baby-steps approach going forward and open a focused issue for a single change first. Thanks for the encouragement, not discouraged at all.

PR #1692, opened 2026-09-20
Fix frameless click-through bug on Xorg and Wayland

scrubbbbs:
This fixes the old bug where we can't click through the transparent region of nomacs window to other windows.

FWICT this originates from the fact that the input protocols of Xorg/Wayland don't have visibility of window backbuffer to do any hit-testing on it as on Windows/macOS. But they do support input masking by regions so I've implemented that. This is possible with QWindow::setMask() without using any xcb/wayland specific APIs.

It seems to be working reasonably well with two limitations:
- Animations: when animating image the input mask doesn't refresh - only at the end of painting. I had to do this to work around suspected Qt/Wayland bug and Xorg limitation.
- Transparent images: not trying to mask the transparency of transparent images yet. It could probably be done if we want to. So this will work on macOS/Windows but not Xorg/Wayland.


ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -2295,7 +2365,7 @@ DkNoMacsContrast::DkNoMacsContrast(QWidget *parent, Qt::WindowFlags flags)
     DkToolBarManager::inst().createTransferToolBar();
 
     setAcceptDrops(true);
-    setMouseTracking(true); // receive mouse event everytime
+    // setMouseTracking(true); // receive mouse event everytime
```
scrubbbbs, line 2368: Meant to remove this

ImageLounge/src/DkGui/DkNoMacs.cpp
```diff
@@ -2154,9 +2156,37 @@ DkNoMacsFrameless::DkNoMacsFrameless(QWidget *parent, Qt::WindowFlags flags)
...
     setWindowFlags(Qt::FramelessWindowHint);
-    setAttribute(Qt::WA_TranslucentBackground, true);
+    setAttribute(Qt::WA_TranslucentBackground, true); // request backing store with an alpha channel
+
+    //
+    // On Wayland and Xorg, click-through on frameless background is only
+    // possible if we use setMask() since these window managers do not hit-test
+    // the backing store like Windows and macOS do.
+    //
+    // On Xorg, setMask() happens immedately and masks painting and mouse
```
scrubbbbs, line 2166: immediately

